PR #7, opened 2024-07-03
Add support for internalization

lebaudantoine:
## Purpose

Fix various issues related to i18n.

## Proposition

- [x] Add an i18n dependency (i18next?)
- [x] Fix `i18n-crowdin` CI steps
- [x] Fix `frontend-i18n-*` makefile commands
- [ ] Validate Crowdin credentials works
- [ ] Validate backend translation works


lebaudantoine:
Huge challenge to support quickly the French, English and German translations.

manuhabitela:
So I think this is mergeable as is.

The idea is:

- use i18next in the frontend app with a global namespace and one namespace for each "feature" dir
- have a script that checks for all `t` calls in the app and extracts the missing keys in the `frontend/locales/xx/*.json` translation files
- in the CI, run this script, and send the updated "source" location files to Crowdin, which are the french translations. Came to the conclusion Crowdin needed a source language and we couldn't really edit translation strings of this source language in the Crowdin UI. Might be wrong but it'll do as is for now I guess :) So we have to remember french locale strings can only be changed through the code, not Crowdin.
- note that the crowdin "push translations" CI job is currently set to run on every PR, that might create some conflicts. I say let's handle that when that poses any problem
- When this is merged I'll be able to better test the new GH workflow I added: it listens to a webhook configured on Crowdin, to create pull requests when new things get translated in Crowdin.

lebaudantoine:
Should we open an issue to keep in mind?

> note that the crowdin "push translations" CI job is currently set to run on every PR, that might create some conflicts. I say let's handle that when that poses any problem

With a low priority tag

src/frontend/package-lock.json
```diff
@@ -1,21 +1,27 @@
...
   "name": "meet",
-  "version": "0.0.0",
+  "version": "0.1.0",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "meet",
-      "version": "0.0.0",
+      "version": "0.1.0",
```
lebaudantoine, line 9: my bad 😬

src/frontend/package.json
```diff
@@ -7,17 +7,24 @@
     "dev": "panda codegen && vite",
     "build": "panda codegen && tsc -b && vite build",
     "lint": "eslint . --ext ts,tsx --report-unused-disable-directives --max-warnings 0",
-    "preview": "vite preview"
+    "preview": "vite preview",
+    "i18n:extract": "i18next -c i18next-parser.config.json"
```
lebaudantoine: I am a bit lost and don't know which make commands run ... should we write a bit of documentation in `/docs` it would be beneficial for all projects (Impress, Regie, etc..).

I would be more than happy to take this task

src/frontend/src/locales/de/global.json
```diff
@@ -0,0 +1,28 @@
...
+    "heading": ""
+  },
+  "homepage": {
+    "heading": "",
+    "intro": "",
+    "createMeeting": "",
+    "login": "",
+    "or": "",
+    "copyMeetingUrl": ""
+  }
```
lebaudantoine: I am not sure what the best practices are for organizing translations. Any resource or tips to share?

src/frontend/src/locales/fr/global.json
```diff
@@ -0,0 +1,28 @@
...
+    "heading": "Page introuvable"
+  },
+  "homepage": {
+    "heading": "Bienvenue dans Meet",
+    "intro": "Que voulez vous faire ? Vous pouvez :",
+    "createMeeting": "Créer une conférence",
+    "login": "Vous connecter pour créer une conférence",
+    "or": "Ou",
+    "copyMeetingUrl": "copier une URL de conférence dans votre barre d'adresse pour rejoindre une conférence existante"
+  }
```
lebaudantoine: Good as it is. In the later run, should we have a dedicated Json for homepage as for the room?

src/frontend/src/locales/fr/global.json
```diff
@@ -0,0 +1,28 @@
...
+  "forbidden": {
+    "heading": "Accès interdit"
+  },
+  "app": "Meet",
+  "login": "Se connecter",
+  "logout": "Se déconnecter",
+  "languageSelector": {
+    "popoverLabel": "Choix de la langue",
+    "buttonLabel": "Changer de langue (actuellement: {{currentLanguage}})"
+  },
```
lebaudantoine: Running the make command will remove these. I am afraid the CI job will not push these, thus creating diff for any dev bootstrapping the project.

lebaudantoine:
Reading your PR I opened this issue https://github.com/numerique-gouv/meet/issues/49

src/frontend/src/locales/fr/global.json
manuhabitela: oh yeah woops, those are used by uncommited stuff, i can remove them for now

src/frontend/src/locales/fr/global.json
manuhabitela: for now each json is tied to a "feature" directory but yeah we could have a homepage feature for sure

src/frontend/package.json
manuhabitela: you can use the `i18n:extract` command (or `make frontend-i18n-extract`) if you want to generate missing translation keys in the json files (it checks all `t` calls in the react app). Handy while developing to make sure you didn't forget any new string to translate. But not actually necessary to use, you can also add your keys by hand while developing normally.

In the CI this is called to send translation keys to Crowdin.

src/frontend/src/locales/de/global.json
manuhabitela: I guess there are multiple ways to see it:

- either have keys that are the actual text to translate. like `Error: page not found`,
- either have keys that represent what we say. Like `pageNotFound`.
- either have keys that represent in what context the string is used. Like `notFound.heading`
- either have keys that are strictly tied to their use in the source code. Like `layouts.NotFoundScreen.heading`

First option goes a bit against how i18next works as it's more meant to be used with actual keys. So just changing a typo in an existing "key" makes us having to update the key in all languages.

I feel second option could quickly become a bit hard to manage, if lots of translation strings at the root level, maybe.

Fourth option makes translations extremely tied to our component structure. If we happen to use a different component for the not found page, we also have to change the key translation everywhere.

That's why I went with the `notFound.heading` style. It's a bit messier compared to fourth option I guess, but the fact our translations are not 1:1 match with components means the translations are more future proof.

And for now, all the "global" translation strings that are not really tied to a specific UI component are placed at the root of the global namespace.

We can figure this out as we go really!

PR #18, opened 2024-07-11
♻️(frontend) reorganize starting frontend code

manuhabitela:
- we now have "features" to try to organize code by intent instead of code type. everything at the root of frontend, not in feature/, is global
- customized the panda config a bunch to try to begin to have an actual design system. The idea is to prevent using arbitrary values here and there in the code, but rather semantic tokens
- changed the userAuth code logic to handle the fact that a 401 on the users/me call is not really an error per say, but rather an indication the user is not logged in


src/frontend/src/features/auth/api/useUser.tsx
```diff
@@ -0,0 +1,17 @@
...
+  const query = useQuery({
+    queryKey: [keys.user],
+    queryFn: fetchUser,
+  })
+
+  return {
+    ...query,
+    // if fetchUser returns false, it means the user is not logged in: expose that
+    user: query.data === false ? undefined : query.data,
+    isLoggedIn: query.data !== false && !!query.data?.email,
```
lebaudantoine: As AC reconciles users via email, the email attribute should always be defined. However, I would argue that the email attribute should not be responsible for determining if a user is logged in or not.

src/frontend/src/features/rooms/api/fetchRoom.ts
```diff
@@ -0,0 +1,20 @@
...
+  username = '',
+}: {
+  roomId: string
+  username?: string
+}) => {
+  return fetchApi<ApiRoom>(
+    `/rooms/${roomId}?username=${encodeURIComponent(username)}`
+  ).then((room) => {
+    if (!room.livekit?.token) {
+      throw new ApiError(500, 'Livekit token not found')
```
lebaudantoine: nitpicking: `LiveKit`hihi 🫣😂 just kiding

src/frontend/src/features/rooms/api/fetchRoom.ts
```diff
@@ -0,0 +1,20 @@
...
+  roomId,
+  username = '',
+}: {
+  roomId: string
+  username?: string
+}) => {
+  return fetchApi<ApiRoom>(
+    `/rooms/${roomId}?username=${encodeURIComponent(username)}`
+  ).then((room) => {
+    if (!room.livekit?.token) {
```
lebaudantoine: should it also check for liveKit url?

src/frontend/src/features/rooms/components/Conference.tsx
```diff
@@ -0,0 +1,48 @@
...
+        roomId: roomId as string,
+        username: userConfig.username,
+      }),
+  })
+
+  return (
+    <QueryAware status={status}>
+      <LiveKitRoom
+        serverUrl={data?.livekit?.url}
+        token={data?.livekit?.token}
```
lebaudantoine, line 32: As you check in `fetchRoom`, might be unnecessary?
```suggestion
        serverUrl={data.livekit.url}
        token={data.livekit.token}
```

src/frontend/src/features/rooms/api/fetchRoom.ts
lebaudantoine:
```suggestion
    if (!room?.livekit?.token) {
```

src/frontend/src/features/rooms/components/Join.tsx
```diff
@@ -0,0 +1,19 @@
...
+}: {
+  onSubmit: (choices: LocalUserChoices) => void
+}) => {
+  return (
+    <Box title="Verify your settings before joining" withBackButton>
+      <PreJoin
+        persistUserChoices
+        onSubmit={(choices) => {
+          onSubmit(choices)
+        }}
```
lebaudantoine:
```suggestion
        onSubmit={onSubmit}
```

src/frontend/src/primitives/Badge.tsx
```diff
@@ -0,0 +1,32 @@
...
+    size: 'normal',
+  },
+})
+
+export type BadgeProps = React.HTMLAttributes<HTMLSpanElement> &
+  RecipeVariantProps<typeof badge>
+
+/**
+ * anchor component styled with underline
+ */
```
lebaudantoine: mismatch

src/frontend/src/primitives/Button.tsx
```diff
@@ -1,8 +1,58 @@
...
+type ButtonProps = RecipeVariantProps<typeof button> &
+  (RACButtonsProps | LinkProps)
+
+export const Button = (props: ButtonProps) => {
+  const [variantProps, componentProps] = button.splitVariantProps(props)
+  if ((props as LinkProps).href !== undefined) {
+    return <Link className={button(variantProps)} {...componentProps} />
+  }
+  return (
+    <RACButton
```
lebaudantoine, line 53: nit-picking so sorryyy:

In panda.config.ts, I noticed that we're using `ra` as the abbreviation for the React Aria Component, but in this instance, we've used `RAC`. Could we standardize the abbreviation to avoid confusion?

src/frontend/src/primitives/H.tsx
```diff
@@ -1,24 +1,14 @@
...
+import { Text } from './Text'
 
-export const H1 = ({
+export const H = ({
   children,
-  className,
+  lvl,
   ...props
-}: HTMLAttributes<HTMLHeadingElement>) => {
+}: React.HTMLAttributes<HTMLHeadingElement> & { lvl: 1 | 2 | 3 }) => {
```
lebaudantoine, line 7: out of curiosity, why having only 1,2,3?

src/frontend/src/primitives/Text.tsx
```diff
@@ -0,0 +1,78 @@
...
+})
+
+type TextHTMLProps = HTMLAttributes<HTMLElement>
+type TextElement =
+  | 'h1'
+  | 'h2'
+  | 'h3'
+  | 'h4'
+  | 'h5'
+  | 'h6'
```
lebaudantoine, line 63: what does happen for those?

lebaudantoine (approved):
GGWP

src/frontend/src/features/rooms/components/Conference.tsx
manuhabitela: typescript has trouble understanding in that case data is always set, that's why

src/frontend/src/primitives/H.tsx
manuhabitela: I didn't setup any actual styles for level 4 headings and more, waiting to see if we have any

src/frontend/src/primitives/Text.tsx
manuhabitela: We make a distincting between heading styles (where 1 2 3 are configured) and the Text HTML tag we might want to use. The `H` component is there to use headings styled as headings. The `Text` component lets you create any "text html tag" (the list you mention) with any "text style".

src/frontend/src/primitives/Button.tsx
manuhabitela: for sure

src/frontend/src/features/rooms/api/fetchRoom.ts
manuhabitela: je vais vous fumer chef

src/frontend/src/features/rooms/api/fetchRoom.ts
manuhabitela: I don't understand why this should be added as when we are here it means the api responded correctly?

manuhabitela: I'll merge as is and change if needed :)

PR #25, opened 2024-07-11
Fix: Error 404 in staging

### LINKED ISSUE #24: Error 404 on staging
## Bug Report

**Problematic behavior**
When pasting a Meet URL in a browser, frontend displays a `404 Not Found` Nginx error.

Capture d’écran 2024-07-11 à 19 29 11

**Additional context/Screenshots**

To reproduce the bug locally using `make start-tilt`), update the tag by `frontend-production` [here](https://github.com/numerique-gouv/meet/blob/d9166e489d47893557f7e9d5ebcef062261e1eec/bin/Tiltfile#L25) . It will reproduce `production` conditions.  (Locally tilt target the `meet-dev` image, which run using `npm run dev`. In production, build's outputs are served as statics.)

Also, the Nginx is wrongly configured, as the `404.html` file described in the [conf](https://github.com/numerique-gouv/meet/blob/d9166e489d47893557f7e9d5ebcef062261e1eec/src/frontend/default.conf#L11), doesn't exist. mybad.

```
2024/07/11 17:26:31 [error] 8#8: *20 open() "/usr/share/nginx/html/404.html" failed (2: No such file or directory), client: 10.42.0.113, server: localhost, request: "GET /aaab-aaabn-aabn HTTP/1.1", host: "meet-staging.beta.numerique.gouv.fr"
```

lebaudantoine:
## Purpose

It closes #24 

I have added some caching on static assets, lmk @manuhabitela wdyt about it.

### DIFF
--- a/src/frontend/default.conf
+++ b/src/frontend/default.conf
@@ -4,12 +4,16 @@ server {
 
   root /usr/share/nginx/html;
 
+  # Serve static files
   location / {
-      try_files $uri index.html $uri/ =404;
+      try_files $uri $uri/ /index.html;
+      # Add no-cache headers
+      add_header Cache-Control "no-cache, no-store, must-revalidate";
+      add_header Pragma "no-cache";  # HTTP 1.0 header for backward compatibility
+      add_header Expires 0;
   }
 
-  error_page 404 /404.html;
-  location = /404.html {
-      internal;
-  }
+  # Optionally, handle 404 errors by redirecting to index.html
+  error_page 404 =200 /index.html;
+  
 }
--- a/src/frontend/default.conf
+++ b/src/frontend/default.conf
@@ -4,6 +4,12 @@ server {
 
   root /usr/share/nginx/html;
 
+  # Serve static files with caching
+  location ~* ^/assets/.*\.(css|js|json|png|jpg|jpeg|gif|ico|svg|woff|woff2|ttf|eot)$ {
+      expires 30d;
+      add_header Cache-Control "public, max-age=2592000";
+  }
+
   # Serve static files
   location / {
       try_files $uri $uri/ /index.html;
@@ -15,5 +21,4 @@ server {
 
   # Optionally, handle 404 errors by redirecting to index.html
   error_page 404 =200 /index.html;
-  
 }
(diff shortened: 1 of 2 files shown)


lebaudantoine:
My fix works locally, however, I cannot deploy it to staging to test it. We need to publish the new frontend image to the registry. 

One amazing enhancement would be to be able having feature branches deployed to a staging environment for review. @rouja

src/frontend/default.conf
```diff
@@ -4,12 +4,16 @@ server {
 
   root /usr/share/nginx/html;
 
+  # Serve static files
   location / {
-      try_files $uri index.html $uri/ =404;
+      try_files $uri $uri/ /index.html;
```
manuhabitela, line 15: Cool!

src/frontend/default.conf
```diff
@@ -4,12 +4,16 @@ server {
...
 
-  error_page 404 /404.html;
-  location = /404.html {
-      internal;
+  # Optionally, handle 404 errors by redirecting to index.html
+  error_page 404 =200 /index.html;
+  
+  location ~* \.(css|js|json|png|jpg|jpeg|gif|ico|svg|woff|woff2|ttf|eot)$ {
+      expires 30d;
+      add_header Cache-Control "public, max-age=2592000";
```
manuhabitela: This makes me thinking:

- we could explicitly say `index.html` wants no cache
- everything in `assets/*` can be cached basically forever instead of "only" 30 days, those assets have cache-busting names
- other assets (the ones from the `public/` dir) we have to watch out because their name always stay the same. It's easy to update a file and forget that it's cached for a long time.

lebaudantoine: I have edited my commits to follow your recommendation. I have set no cache for the `public/`. I'll watch Nginx [webinar](https://www.f5.com/company/events/webinars/content-caching-nginx-plus?_ga=2.38457827.1335572977.1721049773-99741700.1721049773) on caching statics to level up my skills.

lebaudantoine: Inspecting the network tab, cache headers on response seems to be properly set. However, I was wondering why the request has a no-cache header.
Capture d’écran 2024-07-15 à 16 04 38

PR #32, opened 2024-07-16
Poc SSO

### LINKED ISSUE #26: Support SSO
## Feature Request: Enable Seamless Single Sign-On Experience

Currently, when transitioning from another service within La Suite to Meet, users are required to log in again, disrupting workflow and user experience. I propose implementing the Single Sign-On (SSO) to mitigate this issue.

It would be a huge improvement compared to WebConf which untimely asks for login.

Introduce a mechanism to attempt silent authentication using the authorize endpoint with `prompt=none` when the user is not already logged in. If successful, seamlessly authenticate the user in Meet. If unsuccessful, treat the user as anonymous and provide a visible login button within the Meet interface.

No alternatives have been considered at this time.

lebaudantoine:
## Purpose

Offer SSO to our users.

## Proposal

It addresses #26.
Please refer to the issue or commits for more information.

This PR is not intended to be merged. It's only a POC to open a discussion with @manuhabitela.

### DIFF
--- a/src/backend/core/authentication/views.py
+++ b/src/backend/core/authentication/views.py
@@ -1,15 +1,23 @@
 """Authentication Views for the People core app."""
 
+import copy
 from urllib.parse import urlencode
 
 from django.contrib import auth
 from django.core.exceptions import SuspiciousOperation
 from django.http import HttpResponseRedirect
+from django.shortcuts import resolve_url
 from django.urls import reverse
 from django.utils import crypto
 
 from mozilla_django_oidc.utils import (
     absolutify,
+)
+from mozilla_django_oidc.views import (
+    OIDCAuthenticationCallbackView as Wip,
+)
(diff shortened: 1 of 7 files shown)


lebaudantoine:
I believe we should brainstorm the best approach for implementing silent login. Here are some questions to consider:

- Should we perform a silent login once per hour?
- Should we do it once per page visit?
- Should we do it once per user session?

My initial thought is to perform the silent login once per user session. By "user session," I mean the period from when a user visits the site to join or create a meeting until they close the tab after the meeting ends. However, we need to figure out how to define and manage this "user session" technically.

Any thoughts on how we could achieve this?

lebaudantoine:
> Both Failure and Success URL properties are implemented uniformly in callback
  views, redirecting users back to the originating view even on authentication
  failure.
  
Re-thinking of it, I should verify it doesn't introduce any bug or regression.

PR #36, opened 2024-07-17
Handle username in the API

### LINKED ISSUE #29: Get username when requesting an access token
## Bug Report

**Problematic behavior**
Anonymous user won't have any name and logged-in ones, will have their db string representation as displayed name in LiveKit.

**Expected behavior/code**
Allow the frontend to pass Username to the backend, so the LiveKit access token is generated with the right name and identity. LiveKit server then will display the right information.

lebaudantoine:
## Purpose

Display the username typed by user.

## Proposal

The frontend was ready. I've added a new query parameter `username` to the `/room` route in the API. We moved away from email/UUID-v4 usernames, which were totally ugly.

I haven't thoroughly tested all corner cases when passing the string yet. Also, the `generate_token` function lacks unit tests; I will address this asap.

If we're not planning to use named/persistent rooms soon, I propose extracting the LiveKit token generation into a dedicated endpoint. We could take inspiration from the LiveKit demo app, like `https://meet.livekit.io/api/token?identity=Antoine&name=Antoine&roomName=guj4-212l`.

I'm concerned about updating code that might not be used. Before adding complexity to the existing viewset and serializer, I'd prefer to ensure we're heading in the right direction. It might be great to wait for @sampaccoud to initiate architecture discussions.

It closes #29 

Capture d’écran 2024-07-17 à 17 48 02

### DIFF
--- a/src/backend/core/api/serializers.py
+++ b/src/backend/core/api/serializers.py
@@ -134,6 +134,4 @@ class RoomSerializer(serializers.ModelSerializer):
 
         output["is_administrable"] = is_admin
 
-        # todo - pass properly livekit configuration
-
         return output
--- a/src/backend/core/api/serializers.py
+++ b/src/backend/core/api/serializers.py
@@ -122,10 +122,14 @@ class RoomSerializer(serializers.ModelSerializer):
         if role is not None or instance.is_public:
             slug = f"{instance.id!s}".replace("-", "")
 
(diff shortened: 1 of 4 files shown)


lebaudantoine:
it closes #29

manuhabitela:
Looks greaaaat :tada:

PR #39, opened 2024-07-17
📝(tilt) document running the application with Tilt

lebaudantoine:
Based on @rouja instructions, try to document the Tilt stack, to enhance the DX of any newcomers, discovering Meet and trying to run it on K8s.

Having a shared/common onboarding documentation on Tilt with Impress, Regie, and Meet would be amazing.

Especially, to document how to install Tilt and its dependencies.

Important: The frontend is not deployed locally using the production target, and I feel important to document it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -94,6 +94,56 @@ You first need to create a superuser account:
 $ make superuser
 ```
 
+### Run application on local Kubernetes
+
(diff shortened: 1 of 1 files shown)


manuhabitela:
Great :)

PR #42, opened 2024-07-17
🚸(frontend) generate shorter room IDs making URLs easier to share

### LINKED ISSUE #41: Simplify room's name
## Feature Request

The current use of UUID-v4 as room IDs is suboptimal for user experience (UX). These IDs are long and complex, making them cumbersome for users to share in room URLs. ex: 

`https://meet.numerique.gouv.fr/bc54fd45-bffe-4ffa-8797-6d0499b43f56`

**Examples from other platforms:**
- **Google Meet** uses a shorter, user-friendly format for room IDs, such as `raq-rsqx-mjh`, which follows a `xxx-xxxx-xxx` pattern where `x` represents random characters, including letters.
- **LiveKit** uses a similar approach with room IDs like `yyyy-yyyy`, where `y` represents random characters, including letters and numbers.
- **Jitsi** [documentation](https://jitsi.org/security/) suggests avoiding long and uninviting UUIDs in favor of custom, yet sufficiently long, names to prevent collisions.

> If you don’t like the funky way the auto-generated names sound and you don’t want to use a long and uninviting [UUID](https://www.uuidgenerator.net/) (which you totally could), then go ahead ...

UUID-v4 room IDs are long and uninviting. Shorter, custom room IDs can enhance UX by making URLs easier to share and remember.

**Collision Considerations:**
- Google Meet's format results in 26^10 = 1.411671e+14 possible combinations.
- The collision odds for any two UUID-v4s are 1 in 2.71e+18.
  
While UUID-v4s are typically used in database systems for their low collision probability, for ephemeral room IDs, the collision risk of e+14 combinations is acceptable. Additionally, once a meeting ends, its ID can be reused, further reducing collision concerns.

___ 
Implement a custom function in the frontend to generate room IDs similar to Google Meet's naming convention.

___ 
Keeping the current UUID-v4 format.

___ 
LiveKit Example:
```javascript
export function generateRoomId(): string {
  return `${randomString(4)}-${randomString(4)}`;
}

export function randomString(length: number): string {
  let result = '';
  const characters = 'abcdefghijklmnopqrstuvwxyz0123456789';
  const charactersLength = characters.length;
  for (let i = 0; length > i; i++) {
    result += characters.charAt(Math.floor(Math.random() * charactersLength));
  }
  return result;
}
```
[source](https://github.com/livekit-examples/meet/blob/main/lib/client-utils.ts#L31)

lebaudantoine:
## Proposition

Simplify room's name (ID).

## Proposal

Please refer to the issue #41 which explains why simplifying room IDs could be beneficial in terms of user experience (UX).
It closes #41.

### DIFF
--- a/src/frontend/src/features/rooms/utils/generateRoomId.ts
+++ b/src/frontend/src/features/rooms/utils/generateRoomId.ts
@@ -25,3 +25,6 @@ export const generateRoomId = () => {
   ];
   return parts.join('-');
 }
+
+export const roomIdRegex = /^[/](?<roomId>[a-z]{3}-[a-z]{4}-[a-z]{3})$/;
+
--- a/src/frontend/src/features/rooms/utils/generateRoomId.ts
+++ b/src/frontend/src/features/rooms/utils/generateRoomId.ts
@@ -1,5 +1,27 @@
-import { slugify } from '@/utils/slugify'
+
+const getRandomChar = () => {
+  // Google Meet uses only letters in a room identifier
+  const characters = 'abcdefghijklmnopqrstuvwxyz';
+  const charactersLength = characters.length;
+  return characters.charAt(Math.floor(Math.random() * charactersLength))
+}
+
+const generateSegment = (length: number): string => {
+  let segment = '';
+  for (let i = 0; i < length; i++) {
+    segment += getRandomChar();
+  }
+  return segment;
+};
 
 export const generateRoomId = () => {
-  return slugify(crypto.randomUUID())
+  // Generates a unique room identifier following the Google Meet format
+  const shortLength = 3;
+  const longLength = 4;
+  const parts = [
+    generateSegment(shortLength),
+    generateSegment(longLength),
+    generateSegment(shortLength)
+  ];
+  return parts.join('-');
 }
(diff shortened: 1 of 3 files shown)


lebaudantoine:
I'm currently debating whether room ID validation should be handled on the frontend or backend side. While my last commit meets our short-term needs, introducing this 'lock' regex might pose issues in the long run.

Addition of 'named' rooms would contradict my previous commit, as we'd only need to validate roomId formats for dynamically generated rooms. Again, we should discuss this further with Samuel.

During my development, I noticed that passing an `undefined` roomId to the Conference component doesn't trigger an error; instead, the `undefined` value is converted to the string `'undefined'`. Should we improve this part? @manuhabitela

src/frontend/src/features/rooms/utils/generateRoomId.ts
```diff
@@ -1,5 +1,30 @@
-import { slugify } from '@/utils/slugify'
+
+const getRandomChar = () => {
+  // Google Meet uses only letters in a room identifier
```
manuhabitela: so we officially are google-meet driven development huh? :smirk:

manuhabitela:
> During my development, I noticed that passing an undefined roomId to the Conference component doesn't trigger an error; instead, the undefined value is converted to the string 'undefined'. Should we improve this part? @manuhabitela

Good point, I'll change that

manuhabitela:
Thanks! Allowed myself to rewrite a tiny bit the content of the small id generation functions to have a little less boilerplate.

src/frontend/src/features/rooms/utils/generateRoomId.ts
lebaudantoine: We totally remove this mention if you feel it's weird

PR #54, opened 2024-07-21
Language selector (adding new Popover primitives along the way)

manuhabitela:
Add a language selector in the header. Select the language you want through a new popover component.

I feel this will be better in a settings page later, but this will do for now!


manuhabitela:
Allowing myself to merge as this contains a few interesting stuff (and I realized I already merged a thing in the i18n PR that should be in this :grimacing:). Please comment here afterwards if you have any feedback, I'll handle it.

PR #64, opened 2024-07-24
Support SSO (Single Sign-On)

### LINKED ISSUE #26: Support SSO

lebaudantoine:
## Purpose

Offer SSO to our users.

## Proposal

It closes #26. Take a look at the ticket 👀
Cleaned the POC discussed in #32, and added some tests on the backend.

Feel free to drop my last commit and build the frontend as you want. I encountered a "blinking" issue and suggest that we intercept all routes with an authentication middleware on the frontend. This middleware could display a loading spinner until we can confirm the user's login status.

Edit: [your commit ](https://github.com/numerique-gouv/meet/pull/58/commits/b75caab685676180357b3c728952a187f3e01bd8) would probably fix the blinking issue.

Edit 2: It did not


manuhabitela:
Thanks Antoine for this! Should be better now visually. Can you try again @lebaudantoine?

PR #67, opened 2024-07-25
🔥(tsclient) remove tsclient sources

lebaudantoine:
It won't be any useful in the short term, and was broken. If needed, we would add it back to the stack.
(Opinionated choice, feel free to discuss it)


manuhabitela:
Agree!

Something like it will certainly be useful later when we have more api endpoints and all, directly inside the frontend/ dir.

PR #68, opened 2024-07-25
🩹(frontend) fix ignored pre-join options

### LINKED ISSUE #50: Pre-join choices are not respected when joining a room
## Bug Report

**Problematic behavior**
I disabled my mic and my camera, I join a room, my mic and my camera are enabled.

**Expected behavior/code**
Default room configuration, should not override the user choices.
Let's investigate if it's a bug in the backend while generating the access_token or in the frontend when passing the user configuration.

lebaudantoine:
It closes #50 
Adopted LiveKit demo app approach for passing configuration to `LiveKitRoom`. Introduced `roomOptions` for future customizations (e.g., quality, codec).

DeviceId is persisted in local storage despite the boolean flag.
We ended up passing the DeviceId without ever setting the boolean flag to false when any parameter was disabled.

I might have missed smth, please feel free to add any commits.

### DIFF
--- a/src/frontend/src/features/rooms/components/Conference.tsx
+++ b/src/frontend/src/features/rooms/components/Conference.tsx
@@ -1,9 +1,11 @@
+import { useMemo } from "react";
 import { useQuery } from '@tanstack/react-query'
 import {
   LiveKitRoom,
   VideoConference,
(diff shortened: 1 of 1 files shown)


manuhabitela:
thanks!

PR #78, opened 2024-07-25
Rooms navigation: skip join on create and invite people

manuhabitela:
Trying out the "skip prejoin screen when creating" idea you mentioned. I feel it's good for testing even if username is missing (it's the user email).

I also refactored the way we handle navigation because it was pretty cumbersome.


manuhabitela:
Added an invite dialog that shows up when creating a meeting. Style could be less intrusive but it'll do for user testing!


https://github.com/user-attachments/assets/d5fdfb5c-bf98-45bf-9b8a-7fd8a8a50c30

PR #79, opened 2024-07-25
Protect room creation

lebaudantoine:
## Purpose

PoC to prevent anonymous users creating room.

## Proposal

Please, I've explained everything in [this 5min loom](https://www.loom.com/share/303fd0e500c64b158ce081cad3a735bb).

**What's needed to make my code mergeable:**

1. **Code Quality and Organization**
   - Ensure code quality, organization, and typing are matching your standards/rules 🤩🥹

2. **Optimize Token Fetching**
   - Avoid fetching a LiveKit token twice if the user has just created a room through a POST request. It could help determine whether showing a pre-join screen, if the token is already present while navigating a room, the user should not be prompted a pre-join.

3. **Improve Error Messages**
   - Display more specific error messages for anonymous users, such as suggesting they log in (Optionnal).

4. **Handle Room Creation for Authenticated Users**
   - With my changes, if an authenticated user tries to create a room by typing a URL, the room creation won't work anymore. 
   - We need to decide whether to:
     - Make a POST request to create a room if the backend returns a 404 to the user.
     - Modify the backend to allow a `get_or_create` method for rooms.
   
   What do you think about these options?

Let's refine this PoC and merge it !! 🚀

___

**Important considerations ⚠️:** by disallow unregistered rooms, third parties (e.g. webmail, calendar, etc) won't be able anymore to generate a valid room URL only with client-executed code; They will have to make a request to our backend (not a big deal) ... This needs to be discussed, as it's not a current issue, we might ignore this issue as of today #YAGNI


lebaudantoine:
[Room configurations](https://github.com/numerique-gouv/magnify/blob/main/src/frontend/packages/core/src/components/rooms/RoomConfig/RoomConfig.tsx) I mentioned during my Loom from Magnify

lebaudantoine:
Please note that waiting room in Magnify seem to be a default feature from Jitsi. I'll investigate it deeper. Such feature would be so nice when the meeting would be private.

manuhabitela:
super great :)

manuhabitela:
noticed the header duplication on error but didn't fix it yet

lebaudantoine:
it closes #85

lebaudantoine:
I've updated my code, with all the suggestions I proposed.
Quite urgent to merge, to avoid any 'misusage' from anonymous users. I've the feeling of creating a room by navigating to a new URL is really debatable and adds complexity to the code for our MVP. Never mind!

Edit: Releasing this feature might create regression for our current user, as existing room URLs would now require authentication. Let's merge and release it asap.

src/frontend/src/features/rooms/api/createRoom.ts
```diff
@@ -0,0 +1,25 @@
+import { useMutation, UseMutationOptions } from "@tanstack/react-query";
+import { fetchApi } from '@/api/fetchApi';
+import { ApiError } from "@/api/ApiError";
+import { ApiRoom } from "./ApiRoom";
+
+export interface CreateRoomParams {
+  slug: string;
+}
+
+const _createRoom = async ({slug}: CreateRoomParams): Promise<ApiRoom> => {
```
manuhabitela: why the `_`? :)

manuhabitela: is the `async/await` boilerplate necessary there? Can't we just `return fetchApi` ?

src/frontend/src/features/rooms/components/Conference.tsx
```diff
@@ -47,6 +65,12 @@ export const Conference = ({
 
   const [showInviteDialog, setShowInviteDialog] = useState(mode === 'create')
 
+  useEffect(() => {
+    if (isFetchError) {
```
manuhabitela: do we want to create the room for any error or should check for more error details here?

src/frontend/src/features/rooms/components/Conference.tsx
```diff
@@ -77,7 +107,9 @@ export const Conference = ({
           audio={userConfig.audioEnabled}
           video={userConfig.videoEnabled}
         >
-          <VideoConference />
+          <VideoConference
+            chatMessageFormatter={formatChatMessageLinks}
```
manuhabitela, line 110: :tada:

src/frontend/src/layout/CenteredContent.tsx
```diff
@@ -1,13 +1,15 @@
...
 import { Center } from '@/styled-system/jsx'
 
 export const CenteredContent = ({
   title,
+  body,
   children,
   withBackButton,
 }: {
   title?: string
+  body?: string
```
manuhabitela: Two things here:

1) I admit it feels a bit weird to have the possibility for both `children` and `body` here. I'd rather ditch the `children` entirely if we want to say that "any centered content should be one title + one optional body string", and update the couple components that currently pass `children`.
2) I'd rename `body` to `description` here and in other components where it was added, as I feel it better exposes the fact we want a string of text (because it's wrapped in a `p` internally), rather than any "body text".

src/frontend/src/features/home/routes/Home.tsx
```diff
@@ -32,10 +42,10 @@ export const Home = () => {
...
               onPress={
                 isLoggedIn
-                  ? () =>
-                      navigateTo('room', generateRoomId(), {
-                        state: { create: true },
-                      })
+                  ? async () => {
+                    const slug = generateRoomId()
+                    await createRoom({slug})
+                  }
```
manuhabitela, line 48: can we write:

```js
isLoggedIn
  ? () => createRoom({slug: generateRoomId()})
  : undefined
```

?

src/frontend/src/features/rooms/api/ApiRoom.ts
```diff
@@ -8,4 +8,7 @@ export type ApiRoom = {
     room: string
     token: string
   }
+  configuration?: {
+    [key: string]: string | number | boolean;
```
manuhabitela, line 12: didn't quite get what the configuration could contain just reading the code. Is this an object with specific keys that we should list here or…?

manuhabitela (commented):
Cooooool :tada: :tada:

src/frontend/src/features/home/routes/Home.tsx
manuhabitela: I'd even argue, since the user doesn't choose its room name, that the homepage shouldn't have to `generateRoomId`, `createRoom` could internally generate the id.

src/frontend/src/features/rooms/api/ApiRoom.ts
lebaudantoine: Json field in the backend, no format specified yet.

src/frontend/src/features/rooms/components/Conference.tsx
lebaudantoine: Already answer in commits or comments, we should improve this part to only triggers when the error is a 404.

src/frontend/src/layout/CenteredContent.tsx
lebaudantoine: not sure to get the body vs description point, but will follow blindly your reco ! 
As we are early in the project, I would keep the children prop

src/frontend/src/features/home/routes/Home.tsx
lebaudantoine: users will soon have the option to either generate a random ID or type a room name. I believe the homepage should handle this responsibility, allowing the `createRoom` function to be extended accordingly.

src/frontend/src/layout/CenteredContent.tsx
lebaudantoine: Actually, I've removed the body props, and kept the original one.

src/frontend/src/features/rooms/components/Conference.tsx
lebaudantoine: fixup

PR #95, opened 2024-08-03
Cleaning and fixing security issues

lebaudantoine:
## Purpose

General cleanup and fixing a few issues.

## Proposal

These changes are intended to improve the codebase and resolve several issues. The commit [04f1198](https://github.com/numerique-gouv/meet/pull/95/commits/04f1198d8bf52795a112e9e78b80a6e83ed82a69) is still not working. I need to pair program with @sampaccoud to address these issues.


src/helm/env.d/preprod/values.meet.yaml.gotmpl
```diff
@@ -10,7 +10,7 @@ backend:
   envVars:
     DJANGO_CSRF_TRUSTED_ORIGINS: http://meet-preprod.beta.numerique.gouv.fr,https://meet-preprod.beta.numerique.gouv.fr
     DJANGO_CONFIGURATION: Production
-    DJANGO_ALLOWED_HOSTS: "*"
+    DJANGO_ALLOWED_HOSTS: meet-preprod.beta.numerique.gouv.fr
```
sampaccoud, line 13: I don't think this should be defined here in the helm chart. This should be set as environment variable in your deployment configuration?
Idem for other envs.

lebaudantoine: According to @rouja this HelmFile is well configured. I'll merge as it is and let you discuss it

PR #106, opened 2024-08-06
Homemade join

manuhabitela:
## Purpose

Have a custom join screen, that doesn't use the livekit prefab, so that we can precisely render what we want and improve UX overall.

## Proposal

This adds a `HomemadeJoin` component, that the `Room` component now uses instead of the (not deleted) `Join` component.

Notable things:

- a new valtio store has been added to store user choices. It replaces the one integrated in livekit because now we can store more things if needed. Like the livekit user choices, it is persisted in localstorage,
- one commit adds audio output selection, one thing that possible in the livekit prefab, and that can't be stored in the livekit persisted user choices mentioned just above. I'll let you decide if you want to include it now or not. Since the conference screen doesn't have the button yet, maybe it's misleading to have it only in the join screen.
- this adds a few reusable components: toggle buttons, dropdown menus.

## State of this pull request

@lebaudantoine :

- the current join screen was not deleted because maybe we'll want to wait before showing the custom one, I'll let you decide and rename the new component accordingly when we actually switch,
- I didn't test that much cross-browsers because I can't from where I am right now, so you might want to check at least on Safari if everything renders properly,
- dropdowns don't have any default media devices selected when opening them for the first time, I didn't quite get how livekit was doing it in their demo by checking the code… we could improve that,
- the mic/cam toggle buttons could be put in their own components rather easily when we need to, for now everything was made inline in the join component to prevent too much premature optimization

https://github.com/user-attachments/assets/5dcee074-3c3f-4d1d-bf8e-e2597ca77261


lebaudantoine:
I've merged the first commits of your work in a dedicated PR #113

lebaudantoine:
Related to this PR (donno if we address it here), a user suggested, for the first connection, to populate the username field with the Agent Connect First and Last name (if available), to avoid typing your name.

PR #141, opened 2024-08-30
Setting dialog audio v2

lebaudantoine:
## Purpose

Second iteration on the Audio tab under settings

## Proposal

Added the active speaker indicator to help users easily determine if their mic is active. Refactored the layout to match Gmeet's design and improved the responsiveness of several elements.

Capture d’écran 2024-08-30 à 15 58 09


lebaudantoine:
Improvements I suggest for the V3 https://github.com/numerique-gouv/meet/issues/142

PR #191, opened 2024-10-14
Integrate Chat in the sidepanel

lebaudantoine:
Integrate LiveKit chat in the custom Side Panel


lebaudantoine:
Few comments, after a quick auto review. In the next PR I should fix :  
- Wrap the current text area into a component that states it's an auto-resize TextArea
- Extract CSS magic value in global CSS variables for consistent layout.
- Support newline in message formatter.
- Set up chat notifications.
- Enhance Side bar interactivity, always showing it when the div is in scroll mode

PR #193, opened 2024-10-24
Recording feature

lebaudantoine:
## Purpose

Support recording. 

## Proposal

_Will be completed by tomorrow morning_

Mini video presenting the PoC until I write a proper PR body https://www.loom.com/share/5ec61112a18044e09d04e639d90397c0


src/backend/core/admin.py
```diff
@@ -77,3 +77,24 @@ class RoomAdmin(admin.ModelAdmin):
     """Room admin interface declaration."""
 
     inlines = (ResourceAccessInline,)
+
+
+@admin.register(models.Recording)
+class RecordingAdmin(admin.ModelAdmin):
+    """Recording admin interface declaration."""
+
+    list_display = ("id", "status", "room", "creator", "created_at", "worker_id")
```
qbey: I'm not sure, but displaying the "creator" may make N+1 queries here, (to check) you may need to add a select_related to prevent this (or not display the creator here).

src/backend/core/api/viewsets.py
```diff
@@ -165,6 +184,12 @@ class RoomViewSet(
     permission_classes = [permissions.RoomPermissions]
     queryset = models.Room.objects.all()
     serializer_class = serializers.RoomSerializer
+    session_manager = RecordingSessionManager(
+        worker_class=LivekitEgressWorker,
+        output_folder="recordings",
+        server_configurations=settings.LIVEKIT_CONFIGURATION,
+        enable_output_logging=settings.LOG_RECORDING_OUTPUT,
```
qbey: I would move this inside the `LivekitEgressWorker` :

``` python
class LivekitEgressWorker(AbstractRecordingWorker):

    def __init__(self):
        self._output_folder = "recordings"  # might be a setting too?
        self._server_configurations = settings.LIVEKIT_CONFIGURATION
        self._enable_output_logging = settings.LOG_RECORDING_OUTPUT
```

because it highly relies on the "worker kind".

src/backend/core/api/viewsets.py
```diff
@@ -165,6 +184,12 @@ class RoomViewSet(
     permission_classes = [permissions.RoomPermissions]
     queryset = models.Room.objects.all()
     serializer_class = serializers.RoomSerializer
+    session_manager = RecordingSessionManager(
+        worker_class=LivekitEgressWorker,
```
qbey: You are loosing the generic side of the `RecordingSessionManager` here.

Two solutions:
- use a setting along with `django.utils.import_string`
- use a setting and allow `RecordingSessionManager` to have a list of factories to instantiate the class itself.

I would use the first one, possibly with:

``` python
class RecordingSessionManager:
    def __init__(self):
        # settings.RECORDING_WORKER_CLASS = "core.recording.worker.LivekitEgressWorker"
        self._worker: type[AbstractRecordingWorker] = import_string(settings.RECORDING_WORKER_CLASS)()
```

src/backend/core/api/viewsets.py
```diff
@@ -209,10 +227,98 @@ def retrieve(self, request, *args, **kwargs):
...
         return drf_response.Response(data)
 
+    @decorators.action(detail=True, methods=["post"], url_path="start-recording")
+    def start_room_recording(self, request, pk=None):  # pylint: disable=unused-argument
+        """Start room recording."""
+        if not settings.ENABLE_RECORDING:
+            return drf_response.Response(
+                {"error": "Recording is disabled."},
+                status=drf_status.HTTP_403_FORBIDDEN,
+            )
```
qbey: NIT: You may directly do 
```suggestion
            raise PermissionDenied({"error": "Recording is disabled."})
```

src/backend/core/api/viewsets.py
```diff
@@ -209,10 +227,98 @@ def retrieve(self, request, *args, **kwargs):
...
+                creator=request.user, room=room, mode=mode
+            )
+        except IntegrityError as e:
+            # todo - integrity error not specific enough
+            logger.error(
+                "An active recording already exists for room %s: %s", room.slug, e
+            )
+            return drf_response.Response(
+                {"error": f"An active recording already exists for room {room.slug}."},
+                status=drf_status.HTTP_200_OK,
```
qbey: NIT: A 200 Ok with an error? Maybe a 409 Conflict?

qbey (commented):
First part review

src/backend/core/models.py
```diff
@@ -325,3 +363,73 @@ def clean_fields(self, exclude=None):
...
+        choices=RecordingStatusChoices.choices,
+        default=RecordingStatusChoices.ACTIVE,
+    )
+    mode = models.CharField(
+        max_length=20,
+        choices=RecordingModeChoices.choices,
+        default=RecordingModeChoices.SCREEN_RECORDING,
+        verbose_name=_("Recording mode"),
+        help_text=_("Defines the type of recording being performed."),
+    )
```
qbey: Note from our IRL discussion:

I would replace the mode, with a worker_kind, like `screen-recording-livekit-egress-worker` where the kind is an HRID declared by the worker:

``` python
class LivekitEgressWorker(AbstractRecordingWorker):
    # ...


class ScreenRecordingLivekitEgressWorker(LivekitEgressWorker):
    hrid = "screen-recording-livekit-egress-worker"
    # ...
```

So from a recording you can get the worker class directly

src/backend/core/api/viewsets.py
```diff
@@ -293,3 +399,44 @@ class ResourceAccessViewSet(
...
+
+
+class RecordingViewSet(viewsets.GenericViewSet, mixins.ListModelMixin):
+    """
+    API endpoints to access and perform actions on recording.
+    """
+
+    storage = StorageHandler(
+        bucket_name=settings.AWS_STORAGE_BUCKET_NAME, parser_class=MinioParser
+    )
```
qbey: Might be silly from me: The `StorageHandler` seems highly related to the worker, so maybe it should be deduced from the worker class? Or we expect all workers to use the same storage? For a first implementation, we can keep it as it.

lebaudantoine: Actually it's not related to the worker!

lebaudantoine: as of today, I expect all workers to use the same storage

src/backend/core/api/viewsets.py
lebaudantoine: I have enhanced this part and refactored it.

src/backend/core/models.py
lebaudantoine: Mode field is retained, but worker selection is handled by the backend. This keeps client logic simple - clients specify the mode they want, and backend matches appropriate workers.

PR #196, opened 2024-11-04
Refactor authentication

lebaudantoine:
## Purpose

Inspired by @sampaccoud and @qbey works.
Catching up with People and Impress recent's updates.

## Proposal

- [X] Handle empty string sub.
- [X] Fallback on email when no sub matches.
- [X] Handle inactive user.
- [ ] Add full name / short name (will be done in a dedicated PR, it implies migrations)

### DIFF
--- a/src/backend/core/authentication/backends.py
+++ b/src/backend/core/authentication/backends.py
@@ -1,5 +1,6 @@
 """Authentication Backends for the Meet core app."""
 
+from django.conf import settings
 from django.core.exceptions import SuspiciousOperation
 from django.utils.translation import gettext_lazy as _
 
@@ -71,22 +72,31 @@ class OIDCAuthenticationBackend(MozillaOIDCAuthenticationBackend):
                 _("User info contained no recognizable user identification")
             )
 
-        try:
-            user = User.objects.get(sub=sub)
-        except User.DoesNotExist:
-            if self.get_settings("OIDC_CREATE_USER", True):
-                user = User.objects.create(
-                    sub=sub,
-                    email=user_info.get("email"),
-                    password="!",  # noqa: S106
-                )
-            else:
-                user = None
+        email = user_info.get("email")
+        user = self.get_existing_user(sub, email)
 
-        if not user:
+        if not user and self.get_settings("OIDC_CREATE_USER", True):
+            user = User.objects.create(
+                sub=sub,
+                email=email,
+                password="!",  # noqa: S106
+            )
+        elif not user:
             return None
 
         if not user.is_active:
             raise SuspiciousOperation(_("User account is disabled"))
 
         return user
+
+    def get_existing_user(self, sub, email):
+        """Fetch existing user by sub or email."""
+        try:
+            return User.objects.get(sub=sub)
+        except User.DoesNotExist:
+            if email and settings.OIDC_FALLBACK_TO_EMAIL_FOR_IDENTIFICATION:
+                try:
+                    return User.objects.get(email=email)
+                except User.DoesNotExist:
+                    pass
+        return None
--- a/src/backend/core/authentication/backends.py
+++ b/src/backend/core/authentication/backends.py
@@ -66,7 +66,7 @@ class OIDCAuthenticationBackend(MozillaOIDCAuthenticationBackend):
         user_info = self.get_userinfo(access_token, id_token, payload)
         sub = user_info.get("sub")
 
-        if sub is None:
+        if not sub:
             raise SuspiciousOperation(
                 _("User info contained no recognizable user identification")
             )
--- a/src/backend/core/authentication/backends.py
+++ b/src/backend/core/authentication/backends.py
@@ -96,7 +96,11 @@ class OIDCAuthenticationBackend(MozillaOIDCAuthenticationBackend):
         except User.DoesNotExist:
             if email and settings.OIDC_FALLBACK_TO_EMAIL_FOR_IDENTIFICATION:
                 try:
-                    return User.objects.get(email=email)
+                    return User.objects.get(email__iexact=email)
                 except User.DoesNotExist:
                     pass
+                except User.MultipleObjectsReturned as e:
+                    raise SuspiciousOperation(
+                        _("Multiple user accounts share a common email.")
+                    ) from e
         return None
--- a/src/backend/core/authentication/backends.py
+++ b/src/backend/core/authentication/backends.py
@@ -75,26 +75,12 @@ class OIDCAuthenticationBackend(MozillaOIDCAuthenticationBackend):
             user = User.objects.get(sub=sub)
         except User.DoesNotExist:
             if self.get_settings("OIDC_CREATE_USER", True):
-                user = self.create_user(user_info)
+                user = User.objects.create(
+                    sub=sub,
+                    email=user_info.get("email"),
+                    password="!",  # noqa: S106
+                )
             else:
                 user = None
-
-        return user
-
-    def create_user(self, claims):
-        """Return a newly created User instance."""
-
-        sub = claims.get("sub")
-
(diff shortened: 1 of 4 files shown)


src/backend/core/authentication/backends.py
```diff
@@ -68,36 +68,43 @@ def get_or_create_user(self, access_token, id_token, payload):
...
+                "OIDC_FALLBACK_TO_EMAIL_FOR_IDENTIFICATION", True
+            ):
+                try:
+                    return User.objects.get(email__iexact=email.strip())
+                except User.DoesNotExist:
+                    pass
+                except User.MultipleObjectsReturned as e:
+                    raise SuspiciousOperation(
+                        _("Multiple user accounts share a common email.")
+                    ) from e
```
qbey, line 105: This will raise after deployment, don't you want to add a cleanup data migration? Or do you have an already written script to run after deployment to prevent too long downtime for user who authenticated with an Identity Provider which generates random sub?

src/backend/core/authentication/backends.py
```diff
@@ -68,36 +68,43 @@ def get_or_create_user(self, access_token, id_token, payload):
...
         return user
+
+    def get_existing_user(self, sub, email):
+        """Fetch existing user by sub or email."""
+        try:
+            return User.objects.get(sub=sub)
+        except User.DoesNotExist:
+            if email and self.get_settings(
+                "OIDC_FALLBACK_TO_EMAIL_FOR_IDENTIFICATION", True
+            ):
```
qbey: Warning, the default in your `settings.py` for `OIDC_FALLBACK_TO_EMAIL_FOR_IDENTIFICATION` is `False`

src/backend/core/tests/authentication/test_backends.py
```diff
@@ -102,3 +106,139 @@ def get_userinfo_mocked(*args):
...
+    db_user = UserFactory(email="foo@mail.com")
+
+    with django_assert_num_queries(1):
+        user = klass.get_existing_user(db_user.sub, db_user.email)
+
+    assert user == db_user
+
+
+@override_settings(OIDC_FALLBACK_TO_EMAIL_FOR_IDENTIFICATION=False)
+def test_finds_user_when_email_fallback_disabled(django_assert_num_queries):
```
qbey: A more "pytestic" way to do it:
```suggestion
def test_finds_user_when_email_fallback_disabled(django_assert_num_queries, settings):
    settings.OIDC_FALLBACK_TO_EMAIL_FOR_IDENTIFICATION = False
```

src/backend/core/tests/authentication/test_backends.py
```diff
@@ -102,3 +106,139 @@ def get_userinfo_mocked(*args):
...
+    """Wip."""
```
qbey: Remaining "WIP"

src/backend/core/authentication/backends.py
lebaudantoine: removing default in my code. Misleading, having two default values, with one that superseded the other.

src/backend/core/tests/authentication/test_backends.py
lebaudantoine: Aligned all tests.

src/backend/core/authentication/backends.py
lebaudantoine: I am not sure to fully understand this comment, let's discuss it IRL!

lebaudantoine: Discussed IRL, this check is necessary for future users. Current users, would be blocked because of this check. I need to sanitize user's accounts, either manually or with a script, before applying new code.

src/backend/core/api/viewsets.py
```diff
@@ -210,6 +204,12 @@ def retrieve(self, request, *args, **kwargs):
             }
         else:
             data = self.get_serializer(instance).data
+        finally:
```
sampaccoud: :medal_sports:

src/backend/core/authentication/backends.py
```diff
@@ -68,36 +69,41 @@ def get_or_create_user(self, access_token, id_token, payload):
...
         return user
+
+    def get_existing_user(self, sub, email):
+        """Fetch existing user by sub or email."""
+        try:
+            return User.objects.get(sub=sub)
+        except User.DoesNotExist:
+            if email and settings.OIDC_FALLBACK_TO_EMAIL_FOR_IDENTIFICATION:
+                try:
+                    return User.objects.get(email__iexact=email.strip())
```
sampaccoud: any other improvements than this one that I should backport to impress?

src/backend/core/authentication/backends.py
sampaccoud: In impress this migration script solved the issues: https://github.com/numerique-gouv/impress/blob/main/src/backend/core/migrations/0007_fix_users_duplicate.py
A better practice would be to write proper ORM code but in impress we were rushing a hotfix...

src/backend/core/authentication/backends.py
sampaccoud: @lebaudantoine I tried to port this to impress but in fact it raises a lot of questions! Let me contact you irl.

src/backend/core/migrations/0006_merge_duplicate_users.py
```diff
@@ -0,0 +1,64 @@
...
+        User.objects.values('email')
+        .annotate(count=Count('id'))
+        .filter(count__gt=1)
+        .values_list('email', flat=True)
+    )
+
+    for email in emails_with_duplicates:
+        # Keep the oldest user
+        primary_user = User.objects.filter(email=email).order_by('created_at').first()
+        duplicate_user_accounts = User.objects.filter(email=email).exclude(id=primary_user.id)
```
qbey, line 34: nit: I would group these two DB calls into one:

``` python
all_duplicate_user_accounts = list(User.objects.filter(email=email).order_by('created_at')).values_list('id', flat=True)
primary_user_id = all_duplicate_user_accounts[0]
duplicate_user_account_ids = duplicate_user_accounts[1:]
```

I guess there are not much user concerned yet, so you can keep your code.

src/backend/core/migrations/0006_merge_duplicate_users.py
```diff
@@ -0,0 +1,64 @@
...
+                user_id=primary_user.id,
+                resource_id=resource_access.resource.id
+            )
+
+            if existing_primary_access.exists():
+                continue
+
+            # Transfer access to primary user
+            resource_access.user_id = primary_user.id
+            resource_access.save()
```
qbey: Two points:
 - I would also check to give the "max" access if it already exists (ie if the `existing_primary_access` is lower than the `resource_access` it should be updated, don't you think?
 - I guess all the existing_primary_access could be fetched before the loop.

The second point is not mandatory, as stated before, I guess there are not many updates do the migration will remain fast enough.

qbey (commented):
Review of [0006_merge_duplicate_users.py](https://github.com/numerique-gouv/meet/pull/196/commits/e3f54bfc201ed59c9453ede9b6bcc895b92cefd1#diff-ea69145138f2baff1f6411043f9b6f1de7bd9fa9bbfe88535f95714c84a800f7)

Very nice, and much appreciated comments! 👌

src/backend/core/migrations/0006_merge_duplicate_users.py
lebaudantoine: yes, I was not optimizing for performance, but you might be true

src/backend/core/migrations/0006_merge_duplicate_users.py
lebaudantoine: I fixed your first point !

PR #201, opened 2024-11-06
Recording models

lebaudantoine:
No proper description … but commits are well-written. Lmk If you have any question. 
In the next PR, will be added:

- Route to start a recording
- Route to stop a recording
- Worker to interact with LiveKit Egress
- Hook to be notified by Minio the recording was properly saved.

Code is already ready. I just preferred to split it into several "smaller" PR to ship faster, and facilitate the review.


src/backend/core/models.py
```diff
@@ -325,3 +388,225 @@ def clean_fields(self, exclude=None):
...
+        return str(self.id)
+
+    def get_abilities(self, user):
+        """Compute and return abilities for a given user on the recording."""
+
+        roles = set(get_resource_roles(self, user))
+
+        is_owner_or_admin = bool(
+            roles.intersection({RoleChoices.OWNER, RoleChoices.ADMIN})
+        )
```
lebaudantoine, line 514: should be encapsulated in `RoleChoices` method, `has_privileges`  ?

src/backend/core/models.py
```diff
@@ -171,10 +205,39 @@ def email_anonymized(self):
         return f"***@{self.email.split('@')[1]}"
 
 
+def get_resource_roles(resource: models.Model, user: User) -> List[str]:
```
lebaudantoine, line 208: should it return a set?

src/backend/core/models.py
```diff
@@ -325,3 +388,225 @@ def clean_fields(self, exclude=None):
...
+        roles = get_resource_roles(resource, user)
+
+        # if no roles on the resource early return with no abilities
+        if not roles:
+            return abilities
+
+        abilities["retrieve"] = True
+
+        is_admin = RoleChoices.ADMIN in roles
+        is_owner = RoleChoices.OWNER in roles
```
lebaudantoine: should be methods on `RoleChoices`?

src/backend/core/tests/recording/test_api_recordings_delete.py
```diff
@@ -0,0 +1,115 @@
...
+
+
+@pytest.mark.parametrize(
+    "role",
+    ["owner", "administrator"],
+)
+def test_api_recordings_delete_active(role):
+    """
+    Authenticated users should not be allowed to delete an active recording
+    from which they are an admin or owner.
```
lebaudantoine: "from which they have privileges" ?

src/backend/core/tests/recording/test_api_recordings_list.py
```diff
@@ -0,0 +1,186 @@
...
+    mock_user_get_teams.return_value = ["team1", "team2", "unknown"]
+
+    recordings_team1 = [
+        access.recording
+        for access in factories.TeamRecordingAccessFactory.create_batch(2, team="team1")
+    ]
+    recordings_team2 = [
+        access.recording
+        for access in factories.TeamRecordingAccessFactory.create_batch(3, team="team2")
+    ]
```
lebaudantoine, line 91: I don't explicitly test different role for the access, should I parametrize the test?

src/backend/core/tests/test_models_recording.py
```diff
@@ -0,0 +1,212 @@
...
+        "destroy": False,  # Not final status
+        "partial_update": False,
+        "retrieve": True,  # Privileged users can always retrieve
+        "stop": True,  # Not final status, so can stop
+        "update": False,
+    }
+
+
+def test_models_recording_get_abilities_member_active():
+    """Test abilities for a user who isn't the creator."""
```
lebaudantoine: *who is unprivileged

src/backend/core/tests/test_models_recording.py
```diff
@@ -0,0 +1,212 @@
...
+        "destroy": True,
+        "partial_update": False,
+        "retrieve": True,  # Privileged users can always retrieve
+        "stop": False,  # In final status, so can not stop
+        "update": False,
+    }
+
+
+def test_models_recording_get_abilities_member_final():
+    """Test abilities for a user who isn't the creator."""
```
lebaudantoine: *who is unprivileged

src/backend/core/tests/test_models_recording_accesses.py
```diff
@@ -0,0 +1,312 @@
...
+import pytest
+
+from core import factories
+
+pytestmark = pytest.mark.django_db
+
+
+def test_models_recording_accesses_str():
+    """
+    The str representation should include user email, recording title and role.
```
lebaudantoine: recording id*

src/backend/core/tests/test_models_recording_accesses.py
```diff
@@ -0,0 +1,312 @@
...
+    )
+
+
+def test_models_recording_accesses_unique_user():
+    """Recording accesses should be unique for a given couple of user and recording."""
+    access = factories.UserRecordingAccessFactory()
+
+    with pytest.raises(
+        ValidationError,
+        match="This user is already in this recording.",
```
lebaudantoine, line 36: poorly worded. How can you be in the recording.

src/backend/core/models.py
```diff
@@ -325,3 +388,225 @@ def clean_fields(self, exclude=None):
...
+        Room,
+        on_delete=models.CASCADE,
+        related_name="recordings",
+        verbose_name=_("Room"),
+    )
+    status = models.CharField(
+        max_length=20,
+        choices=RecordingStatusChoices.choices,
+        default=RecordingStatusChoices.INITIATED,
+    )
```
lebaudantoine, line 474: Should I add db constraint to avoid transitioning from a state to another (e.g. `Active -> Initiated`). 

Normal flow would look like :
`Initiated -> Active -> Stopped -> Saved`

lebaudantoine (commented):
auto review

src/backend/core/api/permissions.py
```diff
@@ -83,3 +83,14 @@ def has_object_permission(self, request, view, obj):
             return obj.user == user
 
         return obj.resource.is_administrator(user)
+
+
+class AccessPermission(permissions.BasePermission):
+    """Permission class for access objects."""
+
+    def has_permission(self, request, view):
+        return request.user.is_authenticated
```
qbey, line 88: You can inherit from `IsAuthenticated` so you don't have to redefine this method.

src/backend/core/api/permissions.py
```diff
@@ -83,3 +83,14 @@ def has_object_permission(self, request, view, obj):
...
+
+    def has_object_permission(self, request, view, obj):
+        """Check permission for a given object."""
+        return obj.get_abilities(request.user).get(view.action, False)
```
qbey, line 92: Smart :)

Nit: Maybe the class name could be `HasAbilityPermission` to reflect this? Because `AccessPermission` looks like `PermissionPermission` to me 😆

src/backend/core/api/viewsets.py
```diff
@@ -291,3 +291,37 @@ class ResourceAccessViewSet(
...
+    pagination_class = Pagination
+    permission_classes = [permissions.AccessPermission]
+    queryset = models.Recording.objects.all()
+    serializer_class = serializers.RecordingSerializer
+
+    def list(self, request, *args, **kwargs):
+        """Restrict recording returned by the list endpoint"""
+        queryset = self.filter_queryset(self.get_queryset())
+        user = self.request.user
+        if user.is_authenticated:
```
qbey: Actually the user is authenticated I think (thanks to the `permissions.AccessPermission`).

Anyway, if you want to keep it to be sure, you may make an early return instead.

src/backend/core/api/viewsets.py
```diff
@@ -291,3 +291,37 @@ class ResourceAccessViewSet(
...
+            queryset = queryset.filter(
+                Q(accesses__user=user) | Q(accesses__team__in=user.get_teams())
+            )
```
qbey: Don't you want to put this in the `get_queryset`method?

``` python
def get_queryset(self):
        """Restrict recordings to the user's ones."""
        return super().get_queryset().filter(
                Q(accesses__user=user) | Q(accesses__team__in=user.get_teams())
       )
```
doing this would prevent to override the `list` method I guess.

src/backend/core/factories.py
```diff
@@ -65,3 +65,48 @@ class Meta:
...
+
+class UserRecordingAccessFactory(factory.django.DjangoModelFactory):
+    """Create fake recording user accesses for testing."""
+
+    class Meta:
+        model = models.RecordingAccess
+
+    recording = factory.SubFactory(RecordingFactory)
+    user = factory.SubFactory(UserFactory)
+    role = factory.fuzzy.FuzzyChoice([r[0] for r in models.RoleChoices.choices])
```
qbey:
```suggestion
    role = factory.fuzzy.FuzzyChoice(models.RoleChoices.values)
```

src/backend/core/factories.py
```diff
@@ -65,3 +65,48 @@ class Meta:
...
+
+class TeamRecordingAccessFactory(factory.django.DjangoModelFactory):
+    """Create fake recording team accesses for testing."""
+
+    class Meta:
+        model = models.RecordingAccess
+
+    recording = factory.SubFactory(RecordingFactory)
+    team = factory.Sequence(lambda n: f"team{n}")
+    role = factory.fuzzy.FuzzyChoice([r[0] for r in models.RoleChoices.choices])
```
qbey:
```suggestion
    role = factory.fuzzy.FuzzyChoice(models.RoleChoices.values)
```

src/backend/core/models.py
```diff
@@ -171,10 +205,38 @@ def email_anonymized(self):
...
+
+    # Use pre-annotated roles if available from viewset optimization
+    if hasattr(resource, "user_roles"):
+        return resource.user_roles or []
+
+    try:
+        user_teams = user.get_teams()
+        role_query = resource.accesses.filter(
+            models.Q(user=user) | models.Q(team__in=user_teams)
+        )
```
qbey: Seing it a second time, I wonder whether the access models manager should provide a `filter_for_user` like

``` python
def filter_for_user(self, user, *args, **kwargs):
    return self.filter(
        models.Q(user=user) | models.Q(team__in=user.get_teams()),
        *args, 
        **kwargs,
    )
```

src/backend/core/models.py
```diff
@@ -325,3 +387,225 @@ def clean_fields(self, exclude=None):
...
+        assignable_roles = {RoleChoices.ADMIN, RoleChoices.MEMBER}
+
+        # Add owner role if user is owner
+        if is_owner:
+            assignable_roles.add(RoleChoices.OWNER)
+
+        # Remove current role as it can't be assigned
+        assignable_roles.discard(self.role)
+
+        return sorted(str(role) for role in assignable_roles)
```
qbey: I never remember what is the return of the "str" of an enum, I propose to make things explicit:

```suggestion
        return sorted(role.value for role in assignable_roles)
```

src/backend/core/models.py
qbey: Not sure: what would it looks like? `RoleChoices.is_admin(roles)` is not a real improvement to me :)

src/backend/core/models.py
qbey: Probably, but it's ok like that (since you used a distinct)

src/backend/core/models.py
qbey: Not sure a DB constraint is needed, but I would definitely add this expected workflow in the model docstring ;)

src/backend/core/models.py
```diff
@@ -325,3 +387,225 @@ def clean_fields(self, exclude=None):
...
+    class Meta:
+        db_table = "meet_recording"
+        ordering = ("-created_at",)
+        verbose_name = _("Recording")
+        verbose_name_plural = _("Recordings")
+        constraints = [
+            models.UniqueConstraint(
+                fields=["room"],
+                condition=models.Q(status=RecordingStatusChoices.ACTIVE)
+                | models.Q(status=RecordingStatusChoices.INITIATED),
```
qbey: nit
```suggestion
                condition=models.Q(status__in=[RecordingStatusChoices.ACTIVE, RecordingStatusChoices.INITIATED]),
```

src/backend/core/models.py
```diff
@@ -325,3 +387,225 @@ def clean_fields(self, exclude=None):
...
+                name="unique_initiated_or_active_recording_per_room",
+            )
+        ]
+
+    def __str__(self):
+        return str(self.id)
```
qbey: Be careful, because you might end up with an UUID printed/logged without knowing what object it is. I would rather do
```suggestion
        return f"Recording {self.id} (self.status)"  # status is not really needed
```

src/backend/core/models.py
lebaudantoine: agree

src/backend/core/models.py
lebaudantoine: true !

src/backend/core/models.py
lebaudantoine: I was using it for the filename, but you're true.
I will create a proper getter method with an explicit naming.

src/backend/core/api/permissions.py
lebaudantoine: true

src/backend/core/api/permissions.py
lebaudantoine: True, I just reused the same naming as Impress, but your suggestion is more explicit

src/backend/core/models.py
lebaudantoine: Done

src/backend/core/models.py
lebaudantoine: true

src/backend/core/api/viewsets.py
lebaudantoine: done, I have updated tests, as delete return a 404 instead of a 403

src/backend/core/models.py
lebaudantoine: done

PR #204, opened 2024-11-08
Minio webhook

lebaudantoine:
## Purpose

Enable automatic status updates when recordings are saved to object storage.

## Proposal

Implement a secure webhook endpoint to process recording completion notifications from MinIO.

### DIFF
--- a/src/backend/core/api/viewsets.py
+++ b/src/backend/core/api/viewsets.py
@@ -26,6 +26,13 @@ from rest_framework import (
 )
 
 from core import models, utils
+from core.recording.event.authentication import StorageEventAuthentication
+from core.recording.event.exceptions import (
+    InvalidBucketError,
+    InvalidFileTypeError,
+    ParsingEventDataError,
+)
+from core.recording.event.parsers import get_parser
 from core.recording.worker.exceptions import (
     RecordingStartError,
     RecordingStopError,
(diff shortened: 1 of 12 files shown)


src/backend/core/api/viewsets.py
```diff
@@ -301,3 +311,47 @@ def get_queryset(self):
...
+
+        try:
+            recording = models.Recording.objects.get(id=recording_id)
+        except models.Recording.DoesNotExist as e:
+            raise drf_exceptions.NotFound("No recording found for this event.") from e
+
+        if not recording.is_savable():
+            raise drf_exceptions.PermissionDenied(
+                f"Recording with ID {recording_id} cannot be saved because it is either,"
+                f" in an error state or has already been saved."
```
qbey:
```suggestion
                " in an error state or has already been saved."
```

src/backend/core/recording/event/authentication.py
```diff
@@ -0,0 +1,87 @@
...
+
+    AUTH_HEADER = "Authorization"
+    TOKEN_TYPE = "Bearer"  # noqa S105
+
+    def authenticate(self, request):
+        """Validate the Bearer token from the Authorization header."""
+
+        required_token = settings.RECORDING_STORAGE_EVENT_TOKEN
+        if not required_token:
+            return MachineUser(), None
```
qbey: I'm not sure I like the default behavior to be "publicly accessible", don't you think that by default authentication should not be allowed?

lebaudantoine: I have followed your advice, IMO it adds complexity to the code, but following the Single Responsibility Principle, you are true.

PR #206, opened 2024-11-13
Introduce Full and Short Names

lebaudantoine:
## Purpose

Inspired by Impress.

## Proposal

Please refer to commits' body.


src/backend/core/authentication/backends.py
```diff
@@ -104,3 +111,32 @@ def get_existing_user(self, sub, email):
...
+        return full_name or None
+
+    @staticmethod
+    def update_user_if_needed(user, claims):
+        """Update user claims if they have changed."""
+
+        updated_claims = {
+            key: value
+            for key, value in claims.items()
+            if value is not None and value != getattr(user, key)
```
lebaudantoine: There's an open question about filtering None values. Should we consider removing
an existing field as a valid update (e.g., when an identity provider explicitly
removes undesired data)? Current code could lead to data desynchronization.

Since we match users by email, different identity providers might return
conflicting data - one providing names while another returns blank values.
@sampaccoud @qbey  what's your take on this?

qbey: I've seen this pattern in impress (I guess, and maybe in people) and I was wondering the same. 

I ended up with : we don't want to "delete" data, which as you say can lead to desynchronization but I guess we will see if it raises an issue in the future. 

Note: filtering out "None" does not mean filtering out Falsy (ie empty strings), should we? ^^

qbey (commented):
Thanks a lot for writing test using the pytest fixtures ❤️

src/backend/core/authentication/backends.py
lebaudantoine: I will update the code with `if not value`

lebaudantoine: Agree, not an issue today!

Issue #211, opened 2024-11-14
Default browser language detection fails


lebaudantoine:
## Bug Report

**Problematic Behavior**
- Multiple users report app defaulting to English interface (one of them was on Brave).
- Critical for French public sector adoption - government users expect French-first experience

**Expected Behavior**
- App should automatically detect and apply browser's language settings
- French interface should be default for users with French browser settings

**Investigation & Solutions**
1. Immediate Actions:
   - Investigate specific cases (ask me for user details)
   - Verify browser language detection functionality
   
2. UI Improvements:
   - Enhance language selector visibility per Robin's mockups
   - Ensure language preferences persist across sessions

PR #220, opened 2024-11-17
🩹(frontend) use browser language if no localStorage for detection

lebaudantoine:
Previously, language detection was failing to read browser settings correctly. Added explicit detection order to ensure  localStorage preferences are checked before falling back to browser language.


lebaudantoine:
Related to #211

PR #249, opened 2024-11-26
Enable recording

lebaudantoine:
## Description 

The recording feature needs to be enabled in the dev and staging environments. 

Currently, the MinIO webhook is broken. When using the nip.io address, while the MinIO post the webhook, it results in certificate errors, preventing MinIO from posting to the webhook ofc. 

I've tried to read MinIO config documentation. Other type of notification seems to offer a `skip_ssl_verification` which is not supported for the webhook type.

My goal is to add a bucket event notification on `Put` on our bucket. I would like to configure everything in the helmfile, so any dev starting the project has no configuration to do. 

### TODO  

- [x] Fix MinIO webhook certificate errors in the dev environment.  
- [ ] Add a secret for the webhook password and sync secrets.
- [x] Commit frontend code to enable recording (planned for tomorrow morning) to allow Jacques test it.

### DIFF
--- a/src/backend/core/recording/event/authentication.py
+++ b/src/backend/core/recording/event/authentication.py
@@ -47,8 +47,6 @@ class StorageEventAuthentication(BaseAuthentication):
 
     def authenticate(self, request):
         """Validate the Bearer token from the Authorization header."""
-        if not settings.RECORDING_ENABLE_STORAGE_EVENT_AUTH:
-            return MachineUser(), None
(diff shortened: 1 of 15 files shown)


lebaudantoine:
In dev env, the webhook notification needs to query `https://meet.127.0.0.1.nip.io/api/v1.0/recordings/storage-hook/` with authorization, `password`

PR #255, opened 2024-11-29
📱(front) improve control bar responsive

NathanVss:
We want the bottom bar to stay at the center of the screen + I also added a bit of vertical spacing around the control bar @lebaudantoine .

# Before
Capture d’écran 2024-11-29 à 11 35 51

# After
Capture d’écran 2024-11-29 à 11 35 47


lebaudantoine:
That would be amazing, to be able to close the help menu from the toggle, is it doable?

https://github.com/user-attachments/assets/da7a004f-2f62-498e-8672-0afe58444157

PR #272, opened 2024-12-06
Enable toggle for Sound Notifications

### LINKED ISSUE #251: Enable/disable sound notification with a preference
## Feature Request

Users cannot enable or disable sound notifications, which can be disruptive in quiet environments or meetings.

Add a toggle in the notification settings to enable/disable sound notifications, with the preference saved across sessions.

lebaudantoine:
## Purpose

Address user complaints about sound notifications in large meetings by allowing toggle to disable

It closes #251 

## Proposal

https://github.com/user-attachments/assets/f3f9e6aa-8d66-4449-8f4c-9d65bfa8886b


lebaudantoine:
To avoid triggering notifications for reconnecting participants, you could store participant identities in the app state and check against this list when a new participant joins.

### Pros:
- Prevents unnecessary notifications for reconnects.
- Improves user experience by reducing distractions.

### Cons:
- **State complexity:** The list should be initialized with participants already in the room, adding complexity to managing and updating the state when participants join, leave, or reconnect.

- **Anonymous users:** They will have a new identity each time they connect, which may lead to incorrect assumptions about whether a participant is reconnecting or joining for the first time.

- **Performance overhead:** Maintaining participant identities for large rooms or frequent reconnections could introduce performance concerns.

- **Edge cases:** Participants changing identities (e.g., due to network issues or device changes) may not be detected properly, affecting the logic.


I've chosen a more naive option, user who are annoyed by notifications can disable them.
Networking issues are supposed to be solved any time soon.

src/frontend/src/stores/notifications.ts
```diff
@@ -0,0 +1,44 @@
...
+      '[NotificationsStore] Failed to parse stored settings:',
+      error
+    )
+    return DEFAULT_STATE
+  }
+}
+
+export const notificationsStore = proxy<State>(getNotificationsState())
+
+subscribe(notificationsStore, () => {
```
NathanVss, line 39: I like this approach with Valtio. In the future, we could maybe think of a much generic approach in order to do that. With a global store, that use only one central function to persist the store, wdyt?

src/frontend/src/locales/en/settings.json
```diff
@@ -19,6 +19,19 @@
...
+  "notifications": {
+    "heading": "Sound notifications",
+    "label": "sound notifications for",
+    "actions": {
+      "disable": "Disable",
+      "enable": "Enable"
+    },
+    "items": {
+      "participantJoined": "Participant joined",
+      "handRaised": "Message received",
```
NathanVss: In english those two translations are inverted.

NathanVss:
Capture d’écran 2024-12-09 à 14 35 03

When being activated the switch circle is not quite sticked to the right edge.

NathanVss (approved):
LGTM ;) !

src/frontend/src/stores/notifications.ts
lebaudantoine: Not sure to get it, let's discuss it IRL

PR #290, opened 2024-12-20
POC a Brevo integration

lebaudantoine:
## Purpose

Create the infrastructure for Marketing automation with our first users.

## Proposal

Please check my commits;

### DIFF
--- a/src/backend/core/services/marketing_service.py
+++ b/src/backend/core/services/marketing_service.py
@@ -1,0 +1,134 @@
+"""Marketing service in charge of pushing data for marketing automation."""
+
+import logging
+from dataclasses import dataclass
+from functools import lru_cache
+from typing import Dict, List, Optional, Protocol
+
+from django.conf import settings
+from django.core.exceptions import ImproperlyConfigured
+from django.utils.module_loading import import_string
+
+import brevo_python
+
+logger = logging.getLogger(__name__)
+
+
+class ContactCreationError(Exception):
+    """Raised when the contact creation fails."""
+
+
+@dataclass
+class ContactData:
+    """Contact data for marketing service integration."""
+
+    email: str
+    attributes: Optional[Dict[str, str]] = None
+    list_ids: Optional[List[int]] = None
+    update_enabled: bool = True
+
+
+class MarketingServiceProtocol(Protocol):
+    """Interface for marketing automation service integrations."""
+
+    def create_contact(
+        self, contact_data: ContactData, timeout: Optional[int] = None
+    ) -> dict:
+        """Create or update a contact.
+
+        Args:
+            contact_data: Contact information and attributes
+            timeout: API request timeout in seconds
+
+        Returns:
+            dict: Service response
+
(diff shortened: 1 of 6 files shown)


src/backend/meet/settings.py
```diff
@@ -457,6 +457,16 @@ class Base(Configuration):
...
 
+    # Marketing and communication settings
+
+    BREVO_API_KEY = values.Value(
+        None, environ_name="BREVO_API_KEY", environ_prefix=None
+    )
+    BREVO_API_CONTACT_LIST_IDS = values.ListValue(
+        [], environ_name="BREVO_API_CONTACT_LIST_IDS", environ_prefix=None
+    )
+    BREVO_API_CONTACT_ATTRIBUTES = values.DictValue({"VISIO_USER": True})
```
sampaccoud: Should be MARKETING_* or MARKETING_SERVICE_* or EMAILING_* instead of BREVO_*?

src/backend/core/services/marketing_service.py
```diff
@@ -0,0 +1,105 @@
...
+class ContactData:
+    """Data structure for contact information."""
+
+    email: str
+    attributes: Optional[Dict[str, str]] = None
+    list_ids: Optional[List[int]] = None
+    update_enabled: bool = True
+
+
+class MarketingService:
```
sampaccoud: MarketingService should probably be a BaseMarketingService base class which you derive as BrevoMarketingService, one for every backend that people will want to support? I shouldn't be more work.

src/backend/core/services/marketing_service.py
```diff
@@ -0,0 +1,105 @@
...
+
+        configuration = brevo_python.Configuration()
+        configuration.api_key["api-key"] = settings.BREVO_API_KEY
+
+        self._api_client = brevo_python.ApiClient(configuration)
+
+    def create_contact(self, contact_data: ContactData, timeout=None) -> dict:
+        """Create or update a contact in Brevo.
+
+        Contacts are automatically added to both a shared list for "La Suite"
```
sampaccoud: don't mention La Suite as Visio is not related to it?

src/backend/core/services/marketing_service.py
```diff
@@ -0,0 +1,105 @@
...
+        list_ids = (contact_data.list_ids or []) + settings.BREVO_API_CONTACT_LIST_IDS
+
+        contact = brevo_python.CreateContact(
+            email=contact_data.email,
+            attributes=attributes,
+            list_ids=list_ids,
+            update_enabled=contact_data.update_enabled,
+        )
+
+        kwargs = {}
```
sampaccoud: I wouldn't call kwargs as it doesn't carry meaning and is expected by convention to come from the function signature?

src/backend/meet/settings.py
lebaudantoine: The settings should match the renamed class structure and add a configuration option to specify which marketing service to use, with a factory method.

PR #296, opened 2025-01-06
Poc e2ee

lebaudantoine:
## Purpose

This PR implements end-to-end encryption (E2EE) capabilities for LiveKit rooms using WebRTC's Frame Encryption API. The implementation ensures that media streams are encrypted on the sender's side and can only be decrypted by authorized participants, maintaining confidentiality even from the LiveKit server.

It's a proof of concept, not intended to be merged.

According to LiveKit documentation : 

> Connections are established using 256-bit TLS encryption and media streams are encrypted with AES-128. All data at rest is encrypted with AES-256

E2EE adds an extra layer of security features, when you cannot trust the LiveKit server.

## Proposal

During room access token generation, the first participant triggers the creation of a room passphrase, which is encrypted using Fernet symmetric encryption and stored in Redis. Subsequent participants retrieve this encrypted passphrase from the cache. When the room ends, LiveKit's webhook notifies the backend to invalidate the cached passphrase.

Using Fernet symmetric encryption is fine for short-term storage. 

It's a naive approach, but yet simple.

All participants share an encryption key that remains constant throughout the meeting session, without rotation.

## Technical Limitations

Current E2EE feature isn't perfect yet, for a few reasons: 

**Codec Compatibility**

Limited to VP8 video and Opus audio
VP9/AV1 incompatible due to lack of encrypted backup codec support
Required for backwards compatibility with non-VP9 clients

→ doesn't support most recent and optimized Media codec.

**Audio Processing**

RED (Redundant Encoding) unsupported
Server cannot strip RED packets when E2EE enabled

→ Cannot enable this optimization.

**Browser Support**

Requires Firefox 117+ for compatibility
Chrome/Edge: 86+
Safari: 15.4+

→ Not available for our FF 115 users

**Performance Considerations**

Increased CPU usage on resource-constrained devices
Web Worker implementation mitigates main thread impact

**AI or recording**

E2EE disables Egress functionality, preventing AI features that require room audio recording and processing.

→ Disable a differentiating feature

**Documentation**

I haven't found any documentation on LiveKit website.

**Algorithm**

Yet supports only AES-GCM 128, no 256.

## Source code

[FrameCryptor](https://github.com/livekit/client-sdk-js/blob/d6d03ee6844ef49edd5d5783e436a94e53a5390c/src/e2ee/worker/FrameCryptor.ts) is responsible for en-/decrypting media frames.

[KeyProvider](https://github.com/livekit/client-sdk-js/blob/d6d03ee6844ef49edd5d5783e436a94e53a5390c/src/e2ee/KeyProvider.ts#L12) is responsible for key lifecycle.

## Future Enhancements

**Dynamic Encryption**

- Investigation needed: Enable/disable encryption mid-session
- Requires careful key distribution and participant synchronization

**Passphrase Requirements**

- Define secure passphrase format
- verify the passphrase generation is totally secure

**Room Configuration**

- Make encryption opt-in rather than default
- Provide admin controls for encryption settings

## My 2cts

E2EE should be disabled by default for several reasons:

- Not all meetings require the additional security layer and computational overhead
- Enables broader codec support (VP9, AV1) for standard meetings
- Maintains compatibility with features like audio RED and egress
- Not supported on our targeted FF version

End-to-end encryption becomes critical in scenarios where:

- The LiveKit infrastructure might be compromised
- Meeting content is highly sensitive (e.g., "Diffusion Restreinte" discussions, medical data)
- Zero-trust security model is mandated

### DIFF
--- a/src/helm/env.d/dev/values.livekit.yaml.gotmpl
+++ b/src/helm/env.d/dev/values.livekit.yaml.gotmpl
@@ -17,6 +17,10 @@ livekit:
     udp_port: 443
     domain: livekit.127.0.0.1.nip.io
     loadBalancerAnnotations: {}
+  webhook:
+    api_key: devkey
+    urls:
+      - https://meet.127.0.0.1.nip.io/api/v1.0/rooms/livekit-webhook/
 
 
 loadBalancer:
--- a/src/backend/core/api/serializers.py
+++ b/src/backend/core/api/serializers.py
@@ -130,16 +130,16 @@ class RoomSerializer(serializers.ModelSerializer):
             del output["configuration"]
 
         if role is not None or instance.is_public:
-            slug = f"{instance.id!s}"
+            room_id = f"{instance.id!s}"
             username = request.query_params.get("username", None)
 
             output["livekit"] = {
                 "url": settings.LIVEKIT_CONFIGURATION["url"],
-                "room": slug,
+                "room": room_id,
                 "token": utils.generate_token(
-                    room=slug, user=request.user, username=username
+                    room=room_id, user=request.user, username=username
                 ),
-                "passphrase": utils.get_cached_passphrase(slug)
+                "passphrase": utils.get_cached_passphrase(room_id)
             }
             
         output["is_administrable"] = is_admin
--- a/src/backend/core/api/serializers.py
+++ b/src/backend/core/api/serializers.py
@@ -143,6 +143,8 @@ class RoomSerializer(serializers.ModelSerializer):
 
         output["is_administrable"] = is_admin
 
+        output['passphrase'] = "thisisapassphrase"
+
         return output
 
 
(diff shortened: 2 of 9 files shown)


src/helm/env.d/dev/values.livekit.yaml.gotmpl
```diff
@@ -17,6 +17,10 @@ livekit:
     udp_port: 443
     domain: livekit.127.0.0.1.nip.io
     loadBalancerAnnotations: {}
+  webhook:
+    api_key: devkey
+    urls:
+      - https://meet.127.0.0.1.nip.io/api/v1.0/rooms/livekit-webhook/
```
lebaudantoine, line 23: I manually added the misssing authority certificate inside livekit alpine pod. I need to find a better solution !

lebaudantoine:
Capture d’écran 2025-01-07 à 11 10 25

Zoom offers two options, `enhanced encryption` and `e2e encryption` by default, `enhanced encryption` is selected.
What is enhanced encryption in Zoom?

AES 256-bit GCM encryption, protecting message and room content. Encryption keys are generated and managed by Zoom's server.

lebaudantoine:
/!\ default expiration of the cache, corner case, when a participant join right before, and right after.

PR #297, opened 2025-01-07
✨(front) add blurring support on Firefox

NathanVss:
## The issue

The library we use to make video filtering on the project is [track-processor-js](https://github.com/livekit/track-processors-js). This vendor is not compatible with browser other than Chrome and Edge. 

## Why ?

- This vendor uses GPU computing to run inference, which is not possible on Firefox.
- This vendor uses MediaStreamTrackProcessor and MediaStreamTrackGenerator that are not available on Firefox.

## The solution

Knowing that I had to make a custom implementation of video filtering compatible with Firefox, we use https://ai.google.dev/edge/mediapipe/solutions/guide to run inference with the https://storage.googleapis.com/mediapipe-models/image_segmenter/selfie_segmenter_landscape/float16/latest/selfie_segmenter_landscape.tflit model. The inferences run on CPU with gives inferior performances. 

In order to apply post processing to video source frame I could not rely on MediaStreamTrackProcessor and MediaStreamTrackGenerator so I setup a different approach using multiple canvas to draw video source frame and apply inferred mask on it. By using `captureStream` it is possible to extract a video track from a canvas.

### Light Blur
Capture d’écran 2025-01-10 à 16 21 55

# Heavy Blur
Capture d’écran 2025-01-10 à 16 21 59

## Performances

Running inferences on CPU comes at a cost, so in order to optimize the inference time, the image is down sized to 256x144, then fed to the model, and than upscaled as a mask on the video source. 

Pay attention to the small black and white thumbail, this is the mask canvas inferred by the segmentation model:

Capture d’écran 2025-01-10 à 16 24 44

### DIFF
--- a/src/frontend/src/features/rooms/livekit/components/Effects.tsx
+++ b/src/frontend/src/features/rooms/livekit/components/Effects.tsx
@@ -5,11 +5,9 @@ import { Text, P, ToggleButton, Div, H } from '@/primitives'
 import { useTranslation } from 'react-i18next'
 import { HStack, styled, VStack } from '@/styled-system/jsx'
 import {
-  BackgroundBlur,
-  BackgroundOptions,
-  ProcessorWrapper,
-  BackgroundTransformer,
-} from '@livekit/track-processors'
+  BackgroundBlurFactory,
+  BackgroundBlurProcessorInterface,
+} from './blur/index'
 
 const Information = styled('div', {
   base: {
@@ -27,6 +25,8 @@ enum BlurRadius {
   NORMAL = 10,
 }
 
+const isSupported = BackgroundBlurFactory.isSupported()
+
 export const Effects = () => {
   const { t } = useTranslation('rooms', { keyPrefix: 'effects' })
   const { isCameraEnabled, cameraTrack, localParticipant } =
@@ -37,15 +37,12 @@ export const Effects = () => {
   const localCameraTrack = cameraTrack?.track as LocalVideoTrack
 
   const getProcessor = () => {
-    return localCameraTrack?.getProcessor() as ProcessorWrapper<BackgroundOptions>
+    return localCameraTrack?.getProcessor() as BackgroundBlurProcessorInterface
   }
 
   const getBlurRadius = (): BlurRadius => {
     const processor = getProcessor()
-    return (
-      (processor?.transformer as BackgroundTransformer)?.options?.blurRadius ||
-      BlurRadius.NONE
-    )
+    return processor?.options.blurRadius || BlurRadius.NONE
   }
 
   const toggleBlur = async (blurRadius: number) => {
@@ -61,9 +58,11 @@ export const Effects = () => {
       if (blurRadius == currentBlurRadius && processor) {
         await localCameraTrack.stopProcessor()
       } else if (!processor) {
-        await localCameraTrack.setProcessor(BackgroundBlur(blurRadius))
+        await localCameraTrack.setProcessor(
+          BackgroundBlurFactory.getProcessor({ blurRadius })!
+        )
       } else {
-        await processor?.updateTransformerOptions({ blurRadius })
+        processor?.update({ blurRadius })
       }
     } catch (error) {
       console.error('Error applying blur:', error)
@@ -147,7 +146,7 @@ export const Effects = () => {
         >
           {t('heading')}
         </H>
-        {ProcessorWrapper.isSupported ? (
+        {isSupported ? (
           <HStack>
             <ToggleButton
               size={'sm'}
--- a/src/frontend/src/features/rooms/livekit/components/blur/BackgroundBlurCustomProcessor.ts
+++ b/src/frontend/src/features/rooms/livekit/components/blur/BackgroundBlurCustomProcessor.ts
@@ -1,0 +1,278 @@
+import { ProcessorOptions, Track } from 'livekit-client'
+import {
+  FilesetResolver,
+  ImageSegmenter,
+  ImageSegmenterResult,
+} from '@mediapipe/tasks-vision'
+import {
+  CLEAR_TIMEOUT,
+  SET_TIMEOUT,
+  TIMEOUT_TICK,
+  timerWorkerScript,
+} from './TimerWorker'
+import { BackgroundBlurProcessorInterface, BackgroundOptions } from '.'
+
+const PROCESSING_WIDTH = 256
+const PROCESSING_HEIGHT = 144
+
+const SEGMENTATION_MASK_CANVAS_ID = 'background-blur-local-segmentation'
+const BLUR_CANVAS_ID = 'background-blur-local'
+
+const DEFAULT_BLUR = '10'
+
+/**
+ * This implementation of video blurring is made to be run on CPU for browser that are
+ * not compatible with track-processor-js.
+ *
+ * It also make possible to run blurring on browser that does not implement MediaStreamTrackGenerator and
+ * MediaStreamTrackProcessor.
+ */
+export class BackgroundBlurCustomProcessor
+  implements BackgroundBlurProcessorInterface
+{
+  options: BackgroundOptions
+  name: string
+  processedTrack?: MediaStreamTrack | undefined
+
+  source?: MediaStreamTrack
+  sourceSettings?: MediaTrackSettings
+  videoElement?: HTMLVideoElement
+  videoElementLoaded?: boolean
+
+  // Canvas containg the video processing result, of which we extract as stream.
+  outputCanvas?: HTMLCanvasElement
+  outputCanvasCtx?: CanvasRenderingContext2D
+
+  imageSegmenter?: ImageSegmenter
+  imageSegmenterResult?: ImageSegmenterResult
+
+  // Canvas used for resizing video source and projecting mask.
+  segmentationMaskCanvas?: HTMLCanvasElement
+  segmentationMaskCanvasCtx?: CanvasRenderingContext2D
+
+  // Mask containg the inference result.
+  segmentationMask?: ImageData
+
+  // The resized image of the video source.
+  sourceImageData?: ImageData
+
+  timerWorker?: Worker
+
+  constructor(opts: BackgroundOptions) {
+    this.name = 'blur'
+    this.options = opts
+  }
+
+  static get isSupported() {
+    return navigator.userAgent.toLowerCase().includes('firefox')
+  }
+
+  async init(opts: ProcessorOptions<Track.Kind>) {
+    if (!opts.element) {
+      throw new Error('Element is required for processing')
+    }
+
+    this.source = opts.track as MediaStreamTrack
+    this.sourceSettings = this.source!.getSettings()
+    this.videoElement = opts.element as HTMLVideoElement
+
+    this._createMainCanvas()
+    this._createMaskCanvas()
+
+    const stream = this.outputCanvas!.captureStream()
+    const tracks = stream.getVideoTracks()
+    if (tracks.length == 0) {
+      throw new Error('No tracks found for processing')
+    }
+    this.processedTrack = tracks[0]
+
+    this.segmentationMask = new ImageData(PROCESSING_WIDTH, PROCESSING_HEIGHT)
+    await this.initSegmenter()
+    this._initWorker()
+  }
+
+  update(opts: BackgroundOptions): void {
+    this.options = opts
+  }
+
+  _initWorker() {
+    this.timerWorker = new Worker(timerWorkerScript, {
+      name: 'Blurring',
+    })
+    this.timerWorker.onmessage = (data) => this.onTimerMessage(data)
+    // When hiding camera then showing it again, the onloadeddata callback is not fired again.
+    if (this.videoElementLoaded) {
+      this.timerWorker!.postMessage({
+        id: SET_TIMEOUT,
+        timeMs: 1000 / 30,
+      })
+    } else {
+      this.videoElement!.onloadeddata = () => {
+        this.videoElementLoaded = true
+        this.timerWorker!.postMessage({
+          id: SET_TIMEOUT,
+          timeMs: 1000 / 30,
+        })
+      }
+    }
+  }
+
+  onTimerMessage(response: { data: { id: number } }) {
+    if (response.data.id === TIMEOUT_TICK) {
+      this.process()
+    }
+  }
+
+  async initSegmenter() {
+    const vision = await FilesetResolver.forVisionTasks(
+      'https://cdn.jsdelivr.net/npm/@mediapipe/tasks-vision/wasm'
+    )
+    this.imageSegmenter = await ImageSegmenter.createFromOptions(vision, {
+      baseOptions: {
+        modelAssetPath:
+          'https://storage.googleapis.com/mediapipe-models/image_segmenter/selfie_segmenter_landscape/float16/latest/selfie_segmenter_landscape.tflite',
+        delegate: 'CPU', // Use CPU for Firefox.
+      },
+      runningMode: 'VIDEO',
+      outputCategoryMask: true,
+      outputConfidenceMasks: false,
+    })
+  }
+
+  /**
+   * Resize the source video to the processing resolution.
+   */
+  async sizeSource() {
+    this.segmentationMaskCanvasCtx?.drawImage(
+      this.videoElement!,
+      0,
+      0,
+      this.videoElement!.videoWidth,
+      this.videoElement!.videoHeight,
+      0,
+      0,
+      PROCESSING_WIDTH,
+      PROCESSING_HEIGHT
+    )
+
+    this.sourceImageData = this.segmentationMaskCanvasCtx?.getImageData(
+      0,
+      0,
+      PROCESSING_WIDTH,
+      PROCESSING_WIDTH
+    )
+  }
+
+  /**
+   * Run the segmentation.
+   */
+  async segment() {
+    const startTimeMs = performance.now()
+    return new Promise<void>((resolve) => {
+      this.imageSegmenter!.segmentForVideo(
+        this.sourceImageData!,
+        startTimeMs,
+        (result: ImageSegmenterResult) => {
+          this.imageSegmenterResult = result
+          resolve()
+        }
+      )
+    })
+  }
+
+  /**
+   * TODO: future improvement with WebGL.
+   */
+  async blur() {
+    const mask = this.imageSegmenterResult!.categoryMask!.getAsUint8Array()
+    for (let i = 0; i < mask.length; ++i) {
+      this.segmentationMask!.data[i * 4 + 3] = 255 - mask[i]
+    }
+
+    this.segmentationMaskCanvasCtx!.putImageData(this.segmentationMask!, 0, 0)
+
+    this.outputCanvasCtx!.globalCompositeOperation = 'copy'
+    this.outputCanvasCtx!.filter = 'blur(8px)'
+
+    this.outputCanvasCtx!.drawImage(
+      this.segmentationMaskCanvas!,
+      0,
+      0,
+      PROCESSING_WIDTH,
+      PROCESSING_HEIGHT,
+      0,
+      0,
+      this.videoElement!.videoWidth,
+      this.videoElement!.videoHeight
+    )
+
+    this.outputCanvasCtx!.globalCompositeOperation = 'source-in'
+    this.outputCanvasCtx!.filter = 'none'
+    this.outputCanvasCtx!.drawImage(this.videoElement!, 0, 0)
+
+    this.outputCanvasCtx!.globalCompositeOperation = 'destination-over'
+    this.outputCanvasCtx!.filter = `blur(${this.options.blurRadius ?? DEFAULT_BLUR}px)`
+    this.outputCanvasCtx!.drawImage(this.videoElement!, 0, 0)
+  }
+
+  async process() {
+    await this.sizeSource()
+    await this.segment()
+    await this.blur()
+    this.timerWorker!.postMessage({
+      id: SET_TIMEOUT,
+      timeMs: 1000 / 30,
+    })
+  }
+
+  _createMainCanvas() {
+    this.outputCanvas = document.querySelector(
+      'canvas#background-blur-local'
+    ) as HTMLCanvasElement
+    if (!this.outputCanvas) {
+      this.outputCanvas = this._createCanvas(
+        BLUR_CANVAS_ID,
+        this.sourceSettings!.width!,
+        this.sourceSettings!.height!
+      )
+    }
+    this.outputCanvasCtx = this.outputCanvas.getContext('2d')!
+  }
+
+  _createMaskCanvas() {
+    this.segmentationMaskCanvas = document.querySelector(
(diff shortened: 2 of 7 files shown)


lebaudantoine:
Test on my machine
(Visio + FF 134)

Around ~75% cpu usage.

https://github.com/user-attachments/assets/3dae41ec-35fb-498b-860b-8bbcccdb48db


Webconf + FF 134
Around ~95% cpu usage

https://github.com/user-attachments/assets/918638d7-69ee-45a0-a1bc-e10d5e4d43e8


Quality is quite similar, however your code CPU usage seems a bit lower.

src/frontend/package.json
```diff
@@ -16,6 +16,7 @@
     "@livekit/components-react": "2.6.9",
     "@livekit/components-styles": "1.1.4",
     "@livekit/track-processors": "0.3.2",
+    "@mediapipe/tasks-genai": "0.10.20",
```
lebaudantoine: is it used anywhere?`@mediapipe/tasks-genai` seems to be purposed for LLM inference.

src/frontend/src/features/rooms/livekit/components/blur/BackgroundBlurCustomProcessor.ts
```diff
@@ -0,0 +1,278 @@
...
+  timerWorker?: Worker
+
+  constructor(opts: BackgroundOptions) {
+    this.name = 'blur'
+    this.options = opts
+  }
+
+  static get isSupported() {
+    return navigator.userAgent.toLowerCase().includes('firefox')
+  }
```
lebaudantoine, line 69: Hmm, I'm perplexed.
Are we using the right criteria here?

If so, the processor would be available for Firefox 115 users but wouldn't work properly.
Additionally, having a clear check in the code would make it explicit which Browser APIs are missing.

What about Safari?

src/frontend/src/features/rooms/livekit/components/blur/BackgroundBlurCustomProcessor.ts
```diff
@@ -0,0 +1,278 @@
...
+import {
+  CLEAR_TIMEOUT,
+  SET_TIMEOUT,
+  TIMEOUT_TICK,
+  timerWorkerScript,
+} from './TimerWorker'
+import { BackgroundBlurProcessorInterface, BackgroundOptions } from '.'
+
+const PROCESSING_WIDTH = 256
+const PROCESSING_HEIGHT = 144
```
lebaudantoine, line 17: Should we compute these in a method instead of using constants?
This way, we could dynamically calculate them based on device resources in the future.

src/frontend/src/features/rooms/livekit/components/blur/BackgroundBlurCustomProcessor.ts
```diff
@@ -0,0 +1,278 @@
...
+      baseOptions: {
+        modelAssetPath:
+          'https://storage.googleapis.com/mediapipe-models/image_segmenter/selfie_segmenter_landscape/float16/latest/selfie_segmenter_landscape.tflite',
+        delegate: 'CPU', // Use CPU for Firefox.
+      },
+      runningMode: 'VIDEO',
+      outputCategoryMask: true,
+      outputConfidenceMasks: false,
+    })
+  }
```
lebaudantoine: could we use the [initial `init` method?](https://github.com/livekit/track-processors-js/blob/bfe85313109ede83962e027e65b70e0d743ed84e/src/transformers/BackgroundTransformer.ts#L38) (I am not sure)
```
  async init({ outputCanvas, inputElement: inputVideo }: VideoTransformerInitOptions) {
    await super.init({ outputCanvas, inputElement: inputVideo });

    const fileSet = await vision.FilesetResolver.forVisionTasks(
      this.options.assetPaths?.tasksVisionFileSet ??
        `https://cdn.jsdelivr.net/npm/@mediapipe/tasks-vision@${dependencies['@mediapipe/tasks-vision']}/wasm`,
    );

    this.imageSegmenter = await vision.ImageSegmenter.createFromOptions(fileSet, {
      baseOptions: {
        modelAssetPath:
          this.options.assetPaths?.modelAssetPath ??
          'https://storage.googleapis.com/mediapipe-models/image_segmenter/selfie_segmenter/float16/latest/selfie_segmenter.tflite',
        delegate: 'GPU',
        ...this.options.segmenterOptions,
      },
      runningMode: 'VIDEO',
      outputCategoryMask: true,
      outputConfidenceMasks: false,
    });

    // Skip loading the image here if update already loaded the image below
    if (this.options?.imagePath && !this.backgroundImage) {
      await this.loadBackground(this.options.imagePath).catch((err) =>
        console.error('Error while loading processor background image: ', err),
      );
    }
  }
  ```
  
  passing the right `segmenterOptions` to override the delegate mode.

File: src/frontend/src/features/rooms/livekit/components/blur/BackgroundBlurCustomProcessor.ts [hunk 5]
```diff
@@ -0,0 +1,278 @@
...
+
+  onTimerMessage(response: { data: { id: number } }) {
+    if (response.data.id === TIMEOUT_TICK) {
+      this.process()
+    }
+  }
+
+  async initSegmenter() {
+    const vision = await FilesetResolver.forVisionTasks(
+      'https://cdn.jsdelivr.net/npm/@mediapipe/tasks-vision@0.10.2/wasm'
```
lebaudantoine: the version is dynamic in the original implementation

src/frontend/src/features/rooms/livekit/components/blur/BackgroundBlurCustomProcessor.ts
```diff
@@ -0,0 +1,278 @@
...
+  _initWorker() {
+    this.timerWorker = new Worker(timerWorkerScript, {
+      name: 'Blurring',
+    })
+    this.timerWorker.onmessage = (data) => this.onTimerMessage(data)
+    // When hiding camera then showing it again, the onloadeddata callback is not fired again.
+    if (this.videoElementLoaded) {
+      this.timerWorker!.postMessage({
+        id: SET_TIMEOUT,
+        timeMs: 1000 / 30,
```
lebaudantoine, line 110: at first read, this ratio isn't explicit. is it a 30 FPS timer (1000/30ms)? (IMHO)

src/frontend/src/features/rooms/livekit/components/blur/BackgroundBlurCustomProcessor.ts
```diff
@@ -0,0 +1,278 @@
...
+  }
+
+  async destroy() {
+    this.timerWorker?.postMessage({
+      id: CLEAR_TIMEOUT,
+    })
+
+    this.timerWorker?.terminate()
+    this.imageSegmenter?.close()
+  }
```
lebaudantoine, line 280: should we cleanup canvas?

src/frontend/src/features/rooms/livekit/components/blur/BackgroundBlurCustomProcessor.ts
```diff
@@ -0,0 +1,278 @@
...
+const DEFAULT_BLUR = '10'
+
+/**
+ * This implementation of video blurring is made to be run on CPU for browser that are
+ * not compatible with track-processor-js.
+ *
+ * It also make possible to run blurring on browser that does not implement MediaStreamTrackGenerator and
+ * MediaStreamTrackProcessor.
+ */
+export class BackgroundBlurCustomProcessor
```
lebaudantoine, line 31: I really think dynamic quality adjustment based on device performance is essential in the longer run to make this code works perfeclty.

src/frontend/src/features/rooms/livekit/components/blur/BackgroundBlurCustomProcessor.ts
```diff
@@ -0,0 +1,278 @@
...
+          resolve()
+        }
+      )
+    })
+  }
+
+  /**
+   * TODO: future improvement with WebGL.
+   */
+  async paint() {
```
lebaudantoine: `paint` is a bit generic? `blur` more appropriate?

src/frontend/src/features/rooms/livekit/components/blur/BackgroundBlurCustomProcessor.ts
```diff
@@ -0,0 +1,278 @@
...
+      PROCESSING_HEIGHT
+    )
+
+    this.sourceImageData = this.segmentationMaskCanvasCtx?.getImageData(
+      0,
+      0,
+      PROCESSING_WIDTH,
+      PROCESSING_WIDTH
+    )
+  }
```
lebaudantoine, line 167: I don't really to what extent apply Claude's suggestion here, but it suggests to use `createImageBitmap` for improved performance.

```
const imageData = context.getImageData(0, 0, width, height);

// Using createImageBitmap - faster
const bitmap = await createImageBitmap(videoElement);
ctx.drawImage(bitmap, 0, 0);
```

src/frontend/src/features/rooms/livekit/components/blur/BackgroundBlurCustomProcessor.ts
```diff
@@ -0,0 +1,278 @@
...
+
+  /**
+   * TODO: future improvement with WebGL.
+   */
+  async paint() {
+    const mask = this.imageSegmenterResult!.categoryMask!.getAsUint8Array()
+    for (let i = 0; i < mask.length; ++i) {
+      this.segmentationMask!.data[i * 4 + 3] = 255 - mask[i]
+    }
```
lebaudantoine, line 193: WebGL texture would have a huge impact on this part.

src/frontend/src/features/rooms/livekit/components/blur/BackgroundBlurCustomProcessor.ts
```diff
@@ -0,0 +1,278 @@
...
+    })
+  }
+
+  /**
+   * TODO: future improvement with WebGL.
+   */
+  async paint() {
+    const mask = this.imageSegmenterResult!.categoryMask!.getAsUint8Array()
+    for (let i = 0; i < mask.length; ++i) {
+      this.segmentationMask!.data[i * 4 + 3] = 255 - mask[i]
```
lebaudantoine, line 192: `mask[i]` is supposed to be 0 and 1 right? I would love more context on this computation.

src/frontend/src/features/rooms/livekit/components/blur/BackgroundBlurCustomProcessor.ts
lebaudantoine: full suggestion:

> 
> Let me explain `createImageBitmap` and why it's particularly useful for performance optimization in image processing scenarios like this background blur implementation.
> 
> `createImageBitmap` is a web API that creates a highly optimized, low-level representation of an image that's well-suited for rendering to canvas. Here's why it's powerful:
> 
> 1. Performance Benefits:
> ```javascript
> // Current approach - slower
> const imageData = context.getImageData(0, 0, width, height);
> 
> // Using createImageBitmap - faster
> const bitmap = await createImageBitmap(videoElement);
> ctx.drawImage(bitmap, 0, 0);
> ```
> 
> 2. Key Advantages:
> - Hardware acceleration: ImageBitmap objects are often GPU-accelerated
> - Asynchronous processing: Doesn't block the main thread
> - Memory efficient: Better memory management than raw ImageData
> - Can be transferred to Web Workers: Enables true parallel processing
> 
> Here's how you could modify the current code to use `createImageBitmap`:
> 
> ```typescript
> export class BackgroundBlurCustomProcessor {
>   // Add bitmap cache
>   private sourceImageBitmap?: ImageBitmap;
>   
>   async sizeSource() {
>     // Instead of drawing directly to canvas and getting ImageData
>     if (this.sourceImageBitmap) {
>       this.sourceImageBitmap.close(); // Clean up previous bitmap
>     }
>     
>     // Create optimized bitmap from video frame
>     this.sourceImageBitmap = await createImageBitmap(
>       this.videoElement!,
>       {
>         resizeWidth: PROCESSING_WIDTH,
>         resizeHeight: PROCESSING_HEIGHT,
>         resizeQuality: 'medium'
>       }
>     );
> 
>     // Draw the bitmap to get required ImageData for segmentation
>     this.segmentationMaskCanvasCtx?.drawImage(
>       this.sourceImageBitmap,
>       0, 0
>     );
> 
>     this.sourceImageData = this.segmentationMaskCanvasCtx?.getImageData(
>       0, 0,
>       PROCESSING_WIDTH,
>       PROCESSING_HEIGHT
>     );
>   }
> 
>   async paint() {
>     // Use bitmap for faster rendering
>     if (this.sourceImageBitmap) {
>       this.outputCanvasCtx!.drawImage(this.sourceImageBitmap, 0, 0);
>     }
>   }
> 
>   async destroy() {
>     // Clean up bitmap
>     if (this.sourceImageBitmap) {
>       this.sourceImageBitmap.close();
>     }
>     // ... rest of cleanup
>   }
> }
> ```
> 
> 3. Advanced Usage with Web Workers:
> ```typescript
> // In main thread
> const bitmap = await createImageBitmap(videoElement);
> worker.postMessage({ bitmap }, [bitmap]); // Transfer bitmap to worker
> 
> // In worker
> self.onmessage = async (e) => {
>   const { bitmap } = e.data;
>   // Process the bitmap in the worker
>   // Send processed result back
> };
> ```
> 
> 4. Practical Benefits:
> - Reduced memory usage: ImageBitmap is more memory-efficient than raw pixel data
> - Better scaling: Built-in scaling operations are optimized
> - Reduced main thread blocking: Asynchronous processing helps UI responsiveness
> - Efficient worker communication: Can be transferred between threads without copying
> 
> The key difference from the current implementation is that `createImageBitmap` provides a more optimized path for image processing, especially when dealing with video frames. Instead of manually managing pixel data through `getImageData`, you get a pre-optimized image object that's ready for efficient rendering.
> 
> However, there are a few considerations:
> 1. It's an asynchronous API, so you need to handle promises
> 2. You should properly manage bitmap lifecycle with `.close()`
> 3. Browser support should be checked (though it's widely supported now)
> 4. For some very specific pixel-level operations, you might still need ImageData
> 
> This optimization would be particularly effective in this background blur case because:
> - You're processing video frames continuously
> - You're doing resize operations
> - You need efficient transfer of image data between contexts
> - Performance is critical for real-time video processing

src/frontend/src/features/rooms/livekit/components/blur/BackgroundBlurCustomProcessor.ts
lebaudantoine: something like `detectDeviceCapabilities`

```
private async detectDeviceCapabilities(): Promise<DeviceCapabilities> {
    return {
      hasGPU: await this.checkGPUAvailability(),
      cpuCores: navigator.hardwareConcurrence || 1,
      isMobile: /Android|iPhone/i.test(navigator.userAgent),
      memoryLimit: (navigator as any).deviceMemory || 4,
      supportedAPIs: {
        webGL2: this.checkWebGL2Support(),
        webGPU: 'gpu' in navigator,
        sharedArrayBuffer: typeof SharedArrayBuffer !== 'undefined'
      }
    };
  }
  ```

File: src/frontend/src/features/rooms/livekit/components/blur/BackgroundBlurCustomProcessor.ts (hunk 4 above)
Comment (lebaudantoine) [reply src/frontend/src/features/rooms/livekit/components/blur/BackgroundBlurCustomProcessor.ts]: we could also tune the `numThreads`

File: src/frontend/src/features/rooms/livekit/components/blur/BackgroundBlurCustomProcessor.ts (hunk 10 above)
Comment (lebaudantoine) [reply src/frontend/src/features/rooms/livekit/components/blur/BackgroundBlurCustomProcessor.ts:167]: Also, claude seems to suggest using pre-allocated buffer, to avoid consuming to much memory when creating a new frame.

**Review by lebaudantoine — COMMENTED** [review]
First review

File: src/frontend/src/features/rooms/livekit/components/Effects.tsx [hunk 13]
```diff
@@ -27,6 +26,8 @@ enum BlurRadius {
   NORMAL = 10,
 }
 
+const isSupported = getBackgroundBlurProcessorClass()
```
lebaudantoine, line 29: it feels weird to call this method to determine if blur is supported.
also, calling both `getBackgroundBlurProcessorClass` and `getBackgroundBlurProcessor` in the same file confused me, idk why.

src/frontend/src/features/rooms/livekit/components/Effects.tsx
```diff
@@ -61,9 +59,11 @@ export const Effects = () => {
       if (blurRadius == currentBlurRadius && processor) {
         await localCameraTrack.stopProcessor()
       } else if (!processor) {
-        await localCameraTrack.setProcessor(BackgroundBlur(blurRadius))
+        await localCameraTrack.setProcessor(
+          getBackgroundBlurProcessor({ blurRadius })!
```
lebaudantoine: my bad, I haven't fully understood the code, why are we adding a `!` operator there? 
As `getBackgroundBlurProcessor` can return a null value, should we make explicit we won't call `localCameraTrack.setProcessor`  with a null?

src/frontend/src/features/rooms/livekit/components/blur/TimerWorker.ts
```diff
@@ -0,0 +1,71 @@
+/**
```
lebaudantoine, line 1: Happy to discuss it IRL

src/frontend/src/features/rooms/livekit/components/Effects.tsx
lebaudantoine: feels weird to rely on `getBackgroundProcessorClass` to determine if the feature is supported, i would say it's not its responsibility, the function looks an internal function. `getBackgroundBlurProcessorClass` and `getBackgroundBlurProcessor ` are really coupled and acts together as a factory pattern ! 

Should we define a proper factory class, with a method `isSupported`, pure brainsto

src/frontend/src/features/rooms/livekit/components/blur/BackgroundBlurCustomProcessor.ts
NathanVss: Can you develop your point about computing these? Based on which criterias?

src/frontend/src/features/rooms/livekit/components/blur/BackgroundBlurCustomProcessor.ts
NathanVss: I'm not sure: do you want to not have a distinct `initSegmenter` method? I did it to split the code that was not really easy to read. If yes, what is the benefit?

Can't find any doc regarding `numThreads` and mediapipe, can you provide link please?

src/frontend/src/features/rooms/livekit/components/blur/BackgroundBlurCustomProcessor.ts
NathanVss: You mean that ? 

```
const vision = await FilesetResolver.forVisionTasks(
    "https://cdn.jsdelivr.net/npm/@mediapipe/tasks-vision/wasm"
);
```

NathanVss: https://www.npmjs.com/package/@mediapipe/tasks-vision

src/frontend/src/features/rooms/livekit/components/blur/BackgroundBlurCustomProcessor.ts
NathanVss: Yes it is, I can provide a comment / constant

src/frontend/src/features/rooms/livekit/components/blur/BackgroundBlurCustomProcessor.ts
NathanVss: As the canvas is not appended to the DOM we should be safe? I can't find any specific method regarding this in the HTMLCanvasElement, anyway I'm gonna add a canvas = null to make sure it is garbage collected even if the Processor is not cleared itself

NathanVss:
> Test on my machine (Visio + FF 134)
> 
> Around ~75% cpu usage.
> 
>  Enregistrement.de.l.ecran.2025-01-14.a.10.54.10.mov 
> Webconf + FF 134 Around ~95% cpu usage
> 
>  Enregistrement.de.l.ecran.2025-01-14.a.11.02.09.mov 
> Quality is quite similar, however your code CPU usage seems a bit lower.

I think we should find a way to display the fps to make proper comparison, we must integrate performance in regard to compute consumption

src/frontend/src/features/rooms/livekit/components/blur/BackgroundBlurCustomProcessor.ts
NathanVss: We need to discuss about this during a call

src/frontend/src/features/rooms/livekit/components/blur/BackgroundBlurCustomProcessor.ts
NathanVss: I anticipated the future when we will paint backgrounds, but it's not now, let's go!

src/frontend/src/features/rooms/livekit/components/blur/BackgroundBlurCustomProcessor.ts
NathanVss: We might give a shot to those suggestion to see if we can make something out of this for sure!

src/frontend/src/features/rooms/livekit/components/blur/BackgroundBlurCustomProcessor.ts
NathanVss: Oh boy :)

src/frontend/src/features/rooms/livekit/components/blur/BackgroundBlurCustomProcessor.ts
NathanVss: Nope, 0 - 255

src/frontend/src/features/rooms/livekit/components/Effects.tsx
NathanVss: TS stuff here, in facts, this method would never be called if the function returns null because we don't display the button in this scenario

src/frontend/src/features/rooms/livekit/components/blur/BackgroundBlurCustomProcessor.ts
NathanVss: Add comment

src/frontend/src/features/rooms/livekit/components/blur/BackgroundBlurCustomProcessor.ts
NathanVss: Duplicated FPS

src/frontend/src/features/rooms/livekit/components/Effects.tsx
NathanVss: Add Factory class

lebaudantoine (approved):

lebaudantoine (approved):

PR #299, opened 2025-01-13
Catch-up with recents change on chart

lebaudantoine:
@rouja did some heavy refactoring during holidays, catch-up with other repos, especially on secrets.


rouja (approved):
Really nice work ! Well done !

PR #305, opened 2025-01-15
✨(front) revamp pre join screen

NathanVss:
## Purpose

The previous prejoin screenshot was misleading for some screen size, for small screen the join button was invisible + we want to align with the most up to date styles we use.

## Proposal

Capture d’écran 2025-01-15 à 17 18 43

## Resizing

https://github.com/user-attachments/assets/0b5aae8c-dd83-414a-94ad-94e3b0e6cc2b

## On small desktop screens

Capture d’écran 2025-01-15 à 17 42 39

# Before

Capture d’écran 2025-01-15 à 17 19 05
Capture d’écran 2025-01-15 à 17 43 35

# Todo

- [x] Delete TextFieldWrapper, voir implications
- [x] Gestion des tracks
- [x] Devices are disabled by default
- [x] Utilisation ToggleDevice en dehors d'une Room
- [x] Camera disabled blanc
- [x] disabled button

### DIFF
--- a/src/frontend/src/features/rooms/components/Join.tsx
+++ b/src/frontend/src/features/rooms/components/Join.tsx
@@ -1,8 +1,27 @@
 import { useTranslation } from 'react-i18next'
-import { PreJoin, type LocalUserChoices } from '@livekit/components-react'
+import {
+  ParticipantPlaceholder,
+  usePersistentUserChoices,
+  usePreviewTracks,
+  type LocalUserChoices,
+} from '@livekit/components-react'
+import { css } from '@/styled-system/css'
+import { log } from '@livekit/components-core'
+import { defaultUserChoices } from '@livekit/components-core'
 import { Screen } from '@/layout/Screen'
-import { CenteredContent } from '@/layout/CenteredContent'
 import { useUser } from '@/features/auth'
+import React from 'react'
+import {
+  facingModeFromLocalTrack,
+  LocalVideoTrack,
+  Track,
+} from 'livekit-client'
+import { H } from '@/primitives/H'
+import { SelectToggleDevice } from '../livekit/components/controls/SelectToggleDevice'
+import { Field } from '@/primitives/Field'
+import { Button } from '@/primitives'
+
+const onError = (e: Error) => console.error('ERROR', e)
 
 export const Join = ({
   onSubmit,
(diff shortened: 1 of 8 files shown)


lebaudantoine:
wow, nice job, already x10 times better.
I would break the row into column slightly before, at this moment.

Capture d’écran 2025-01-15 à 22 26 47

lebaudantoine:
Also, nit-picking comments incoming:

Given and family names are rarely longer than 20–25 characters. Could we reduce the input size (label n°1 in my drawing), particularly when the screen width starts to shrink? This adjustment would help maintain the initial proportions between the left and right columns, which are visually perfect and very satisfying.

(do you like having the input's max size visually displayed?)

Really nit-picking here, but I love the (almost) visual alignment between the title and the CTA (label n°2 in my drawing). I’m not sure if it was intentional, but the button’s smaller width compared to the input makes the input feel like it’s floating on a cloud.

Capture d’écran 2025-01-15 à 22 29 18

NathanVss:
> Also, nit-picking comments incoming:
> 
> Given and family names are rarely longer than 20–25 characters. Could we reduce the input size (label n°1 in my drawing), particularly when the screen width starts to shrink? This adjustment would help maintain the initial proportions between the left and right columns, which are visually perfect and very satisfying.
> 
> (do you like having the input's max size visually displayed?)
> 
> Really nit-picking here, but I love the (almost) visual alignment between the title and the CTA (label n°2 in my drawing). I’m not sure if it was intentional, but the button’s smaller width compared to the input makes the input feel like it’s floating on a cloud.
> 
> Capture d’écran 2025-01-15 à 22 29 18

Here are three propositions that we coud discuss:

# 1 
Capture d’écran 2025-01-16 à 11 24 21

# 2
Capture d’écran 2025-01-16 à 11 24 18

# 3
Capture d’écran 2025-01-16 à 11 24 15


My personal order of preference: 2 3 1

lebaudantoine:
Enhancement: Display an explicit message when camera is disabled instead of an avatar?

Capture d’écran 2025-01-18 à 22 28 38

lebaudantoine:
When I open the camera selector, even if my camera is opened, I have no option selected.

Capture d’écran 2025-01-18 à 22 26 49

lebaudantoine:
Bug: I type my name, then type enter. It submits a form, I navigate to the same URL with `?` appended at the end.

Capture d’écran 2025-01-18 à 22 26 19

lebaudantoine:
error message should be internationalized
Capture d’écran 2025-01-18 à 22 25 24

lebaudantoine:
Enhancement: I would suggest limiting the field to 50/60 chars, I just entered the room with a 200 char name, it feels weird.

Capture d’écran 2025-01-18 à 22 24 38

lebaudantoine:
Question: is that field still pre-filled with username given by ProConnect if no previous one was persisted in local storage?

Capture d’écran 2025-01-18 à 22 22 31

lebaudantoine:
Introducing Marianne caused several text elements in the app to appear excessively small. Let's address this to enhance accessibility.

Capture d’écran 2025-01-18 à 22 19 02

Capture d’écran 2025-01-18 à 22 19 02

lebaudantoine:
Nit-picking: with Marianne, the heading title is bigger, and doesn't fit on two lines properly.

Capture d’écran 2025-01-18 à 22 17 32

lebaudantoine:
Question: Should we align left this element?

Capture d’écran 2025-01-18 à 22 20 22

lebaudantoine:
This transition feels glitchy. Also when I long press the Camera toggle button, only the shadow are shown for ~1s. let's show you IRL


https://github.com/user-attachments/assets/1a415003-3dc5-4d4e-9df1-a8315005886b





https://github.com/user-attachments/assets/68feaba0-0f6c-4a19-be3f-7a02f46474c1

lebaudantoine:
Camera is reverse, and switch back in the right orientation when entering the room

Capture d’écran 2025-01-18 à 22 19 36

lebaudantoine (commented):
Did a product-oriented review. Will review the code

NathanVss:
> Question: is that field still pre-filled with username given by ProConnect if no previous one was persisted in local storage?
> 
> Capture d’écran 2025-01-18 à 22 22 31

It is based on "user?.full_name" as before, so it is supposed to be pre-set

NathanVss:
> Question: Should we align left this element?
> 
> Capture d’écran 2025-01-18 à 22 20 22

We already said that we go for option 2 ( with text centered )

src/frontend/src/primitives/Field.tsx
```diff
@@ -25,11 +25,31 @@ const FieldWrapper = styled('div', {
...
+        width: '100%',
+      },
+    },
+  },
 })
 
 const StyledLabel = styled(Label, {
   base: {
     display: 'block',
+    fontSize: '12px',
```
lebaudantoine: isn't it a bit small in terms of accessibility, also should it be rem units?

src/frontend/src/features/rooms/livekit/components/controls/ToggleDevice.tsx
```diff
@@ -69,3 +74,9 @@ export const ToggleDevice = ({
     </ToggleButton>
   )
 }
+
+const ActiveSpeakerWrapper = () => {
+  const { localParticipant } = useLocalParticipant()
+  const isSpeaking = useIsSpeaking(localParticipant)
+  return <ActiveSpeaker isSpeaking={isSpeaking} pushToTalk />
+}
```
lebaudantoine, line 86: Should we incorporate this logic inside the ActiveSpeaker component directly? 
Should we extend it in a LocalActiveSpeaker?

src/frontend/src/locales/fr/rooms.json
```diff
@@ -18,13 +18,16 @@
       "enable": "Activer le micro",
       "label": "Microphone"
     },
-    "heading": "Vérifiez vos paramètres",
+    "heading": "Rejoignez la réunion",
```
lebaudantoine: nit: `Rejoindre la réunion ?` feels less assertive, and more user-friendly, like, the user has the control

src/frontend/src/features/rooms/livekit/components/controls/SelectToggleDevice.tsx
```diff
@@ -81,7 +94,19 @@ export const SelectToggleDevice = <T extends ToggleSource>({
   const trackProps = useTrackToggle(props)
 
   const { devices, activeDeviceId, setActiveMediaDevice } =
-    useMediaDeviceSelect({ kind: config.kind })
+    useMediaDeviceSelect({ kind: config.kind, track })
+
+  /**
+   * When providing only track outside of a room context, activeDeviceId is undefined.
+   * So we need to initialise it with the initialDeviceId.
```
lebaudantoine: *initialize

src/frontend/src/features/rooms/livekit/components/controls/SelectToggleDevice.tsx
```diff
@@ -81,7 +94,19 @@ export const SelectToggleDevice = <T extends ToggleSource>({
...
+   * nb: I don't understand why useMediaDeviceSelect cannot infer it from track device id.
+   */
+  useEffect(() => {
+    if (initialDeviceId !== undefined) {
+      setActiveMediaDevice(initialDeviceId)
+    }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [setActiveMediaDevice])
```
lebaudantoine, line 108: looks like a hack, have you checked the source code? should we make a PR on their repo?

src/frontend/src/features/rooms/components/Join.tsx
```diff
@@ -11,20 +30,366 @@ export const Join = ({
...
+                fullWidth
+              >
+                {joinLabel}
+              </Button>
+            </form>
+          </div>
+
+          {debug && (
+            <>
+              <strong>User Choices:</strong>
```
lebaudantoine: nit: why not using some`VStack` or `HStack` component whenever needed, to avoid too much css in js?

src/frontend/src/features/rooms/components/Join.tsx
```diff
@@ -11,20 +30,366 @@ export const Join = ({
 }) => {
   const { t } = useTranslation('rooms')
   const { user } = useUser()
+  const defaults: Partial<LocalUserChoices> = { username: user?.full_name }
+  const persistUserChoices = true
+  const joinLabel = t('join.joinLabel')
+  const userLabel = t('join.usernameLabel')
+  const debug = false
```
lebaudantoine: should we keep, and the debug part in the dom? React dev tools are sufficients?

lebaudantoine (approved):
LGTM, I would remove the debug code tbh !

src/frontend/src/features/rooms/livekit/components/controls/ToggleDevice.tsx
NathanVss: It would cause a react error or eslint error saying that hooks cannot be nested inside if statements. useIsSpeaking in throwing outside of room context.

src/frontend/src/primitives/Field.tsx
NathanVss: IMO I like this size, but we could use 14px if you want. Yep I should have used rem

src/frontend/src/locales/fr/rooms.json
NathanVss: Agreed

src/frontend/src/features/rooms/livekit/components/controls/SelectToggleDevice.tsx
NathanVss: It is exactly what they do in the source code 🥲

src/frontend/src/features/rooms/components/Join.tsx
NathanVss: I personnally think using indirect components make the DX worse, everyone knows about flex, but not everyone know avec `MyVariousStack` and stuff. That's why I tend to avoid using those

src/frontend/src/features/rooms/components/Join.tsx
NathanVss: We can remove it yep

PR #308, opened 2025-01-17
Translation: feedback-s

bvhme:
Feedback is already a mass noun.


lebaudantoine:
thanks for your contribution @bvhme. Fixed it in #310

PR #315, opened 2025-01-21
Allow custom room names

lebaudantoine:
## Purpose

Inspired by Jitsi's user-friendly approach, adds support for custom room names (e.g., 'team-standup', 'daily-meeting') in addition to the existing system-generated IDs (e.g., 'abc-defg-hij').

/!\ Breaking change: This introduces a permanent update to room URL patterns. Since both formats (system-generated and custom) will be shareable links, we must maintain backwards compatibility to avoid breaking existing URLs.

## Proposal

Proposed UX. The UI is quite poor, will enhance it in the next iterations!

https://github.com/user-attachments/assets/88f5bc8a-095a-4a31-a4de-d8fda6f106f6

### DIFF
--- a/src/frontend/src/features/home/components/PersonalizeMeetingDialog.tsx
+++ b/src/frontend/src/features/home/components/PersonalizeMeetingDialog.tsx
@@ -1,0 +1,134 @@
+import { FormEvent, useState } from 'react'
+import { useTranslation } from 'react-i18next'
+import { Button, Dialog, type DialogProps, Text } from '@/primitives'
+import { HStack } from '@/styled-system/jsx'
+import { RiSpam2Fill } from '@remixicon/react'
+import { navigateTo } from '@/navigation/navigateTo.ts'
+import { useCreateRoom } from '@/features/rooms'
+import {
+  FieldError,
+  Form as RACForm,
+  Input,
+  Label,
+  TextField,
+} from 'react-aria-components'
+import { css } from '@/styled-system/css'
+
+export const PersonalizeMeetingDialog = ({
+  isOpen,
+  ...dialogProps
+}: Omit<DialogProps, 'title'>) => {
+  const { t } = useTranslation('home', {
+    keyPrefix: 'personalizeMeetingDialog',
+  })
+  const { mutateAsync: createRoom } = useCreateRoom()
+  const [roomSlug, setRoomSlug] = useState('')
+  const [serverErrors, setServerErrors] = useState<Array<string>>([])
+
+  const onSubmit = (e: FormEvent<HTMLFormElement>) => {
+    e.preventDefault()
+    createRoom({ slug: roomSlug })
+      .then((data) =>
+        navigateTo('room', data.slug, {
+          state: { create: true, initialRoomData: data },
+        })
+      )
+      .catch((e) => {
+        const msg =
+          e.statusCode === 400
+            ? t('errors.server.taken')
+            : t('errors.server.unknown')
+        setServerErrors([msg])
+      })
+  }
+
+  const validationErrors = []
+  if (roomSlug.length < 5) {
+    validationErrors.push(t('errors.validation.length'))
+  }
+  if (!roomSlug.match(/^[a-z0-9]+$/)) {
+    validationErrors.push(t('errors.validation.spaceOrSpecialCharacter'))
+  }
+
+  const errors = [...validationErrors, ...serverErrors]
+
+  return (
+    <Dialog isOpen={!!isOpen} {...dialogProps} title={t('heading')}>
+      <RACForm onSubmit={onSubmit}>
+        <TextField
+          isRequired
+          isInvalid={errors.length > 0}
+          value={roomSlug}
+          onChange={(e) => {
+            setServerErrors([])
+            setRoomSlug(e.toLowerCase())
+          }}
+          className={css({
+            display: 'flex',
+            flexDirection: 'column',
+          })}
+        >
+          <Label>{t('label')}</Label>
+          <div
+            className={css({
+              display: 'flex',
+              justifyContent: 'space-between',
+            })}
+          >
+            <Input
+              className={css({
+                height: '46px',
+                border: '1px solid black',
+                padding: '4px 8px',
+                width: '80%',
+                borderRadius: '0.25rem',
+              })}
+              placeholder={t('placeholder')}
+            />
+            <Button type="submit">{t('submit')}</Button>
+          </div>
+          <div
+            className={css({
+              minHeight: '72px',
+            })}
+          >
+            <FieldError>
+              <ul
+                className={css({
+                  listStyle: 'square inside',
+                  color: 'red',
+                  marginLeft: '10px',
+                  marginTop: '10px',
+                })}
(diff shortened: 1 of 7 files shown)


lebaudantoine:
Discussed IRL with Arnaud: with should at least enforce 10char, and at first enables this feature with a feature flag, to only our top users, and gather early feedback.

NathanVss:
Submit button should be disabled if there are errors

Capture d’écran 2025-01-27 à 17 28 21

NathanVss:
Broader remark: I think using API response error names such as "error: NameExists", is much more easier to use on the frontend side instead of a message. 

Using error name makes it easier to make conditions, find translations on the frontend side.

Capture d’écran 2025-01-27 à 17 32 31

src/frontend/src/locales/fr/home.json
```diff
@@ -24,6 +25,23 @@
     "copied": "Lien copié dans le presse-papiers",
     "permissions": "Les personnes disposant de ce lien n'ont pas besoin de votre autorisation pour rejoindre cette réunion."
   },
+  "personalizeMeetingDialog": {
+    "heading": "Votre lien personnalisé",
+    "label": "Votre lien personnalisé",
+    "submit": "Créer",
+    "placeholder": "standupbizdev",
```
NathanVss, line 32: WDYT about French placeholder?

src/frontend/src/features/home/components/PersonalizeMeetingDialog.tsx
```diff
@@ -0,0 +1,134 @@
...
+    e.preventDefault()
+    createRoom({ slug: roomSlug })
+      .then((data) =>
+        navigateTo('room', data.slug, {
+          state: { create: true, initialRoomData: data },
+        })
+      )
+      .catch((e) => {
+        const msg =
+          e.statusCode === 400
```
NathanVss, line 42: Related to my comment about error names: what if there are multiple possible errors with the same http code?

Not to fix now ofc

src/frontend/src/features/home/components/PersonalizeMeetingDialog.tsx
```diff
@@ -0,0 +1,134 @@
...
+}: Omit<DialogProps, 'title'>) => {
+  const { t } = useTranslation('home', {
+    keyPrefix: 'personalizeMeetingDialog',
+  })
+  const { mutateAsync: createRoom } = useCreateRoom()
+  const [roomSlug, setRoomSlug] = useState('')
+  const [serverErrors, setServerErrors] = useState<Array<string>>([])
+
+  const onSubmit = (e: FormEvent<HTMLFormElement>) => {
+    e.preventDefault()
```
NathanVss, line 33: why?

src/frontend/src/features/home/components/PersonalizeMeetingDialog.tsx
```diff
@@ -0,0 +1,134 @@
...
+      })
+  }
+
+  const validationErrors = []
+  if (roomSlug.length < 5) {
+    validationErrors.push(t('errors.validation.length'))
+  }
+  if (!roomSlug.match(/^[a-z0-9]+$/)) {
+    validationErrors.push(t('errors.validation.spaceOrSpecialCharacter'))
+  }
```
NathanVss, line 55: I like how simple this approach is were most of devs wants to use libraries to validate their forms :)

src/frontend/src/features/home/components/PersonalizeMeetingDialog.tsx
```diff
@@ -0,0 +1,134 @@
...
+
+  const errors = [...validationErrors, ...serverErrors]
+
+  return (
+    <Dialog isOpen={!!isOpen} {...dialogProps} title={t('heading')}>
```
NathanVss, line 60: Broader subject (out of scope): on mobile I tend to think having a full screen modal with its content being aligned on top is more UX. Mobile user are more used to panes-like navigation. 

Capture d’écran 2025-01-27 à 17 38 52

src/frontend/src/features/home/routes/Home.tsx
```diff
@@ -209,6 +211,18 @@ export const Home = () => {
...
                       {t('createMenu.laterOption')}
                     </MenuItem>
+                    <MenuItem
+                      className={
+                        menuRecipe({ icon: true, variant: 'light' }).item
+                      }
+                      onAction={() => {
+                        setIsPersonalizeModalOpen(true)
+                      }}
+                      data-attr="create-option-personalize"
```
NathanVss, line 221: What is it used for ?

NathanVss (approved):
Nice job bud

Tbh I still have doubts about real needs for this feature as gmeet does not implement it. But anyway if users request it and it increase the adoption, lez go ;) !!

src/frontend/src/locales/fr/home.json
lebaudantoine: I felt bizdev was kinda french, but would you prefer, "equipecom" or smth else?

lebaudantoine:
> Submit button should be disabled if there are errors

In fact, it is, the form is disabled if it's invalid. But indeed the style is needed.

lebaudantoine:
> Broader remark: I think using API response error names such as "error: NameExists", is much more easier to use on the frontend side instead of a message.
> 
> Using error name makes it easier to make conditions, find translations on the frontend side

I think it's the default Django error formats inherited from Joanie. Broader refactoring, to be honest.
I agree with you

src/frontend/src/features/home/components/PersonalizeMeetingDialog.tsx
lebaudantoine: Agree, we have a poor mobile UX.

src/frontend/src/features/home/routes/Home.tsx
lebaudantoine: It's used by Posthog to scrap event data!

src/frontend/src/features/home/components/PersonalizeMeetingDialog.tsx
lebaudantoine: I may have overlooked something in the React Aria documentation, but this is the approach I found to prevent a HTML form from being submitted and navigating to a URL suffixed with a question mark.

src/frontend/src/locales/fr/home.json
NathanVss: Maybe like "Réunion Hebdo" or smth?

LNGGlago:
Hi,

We are waiting this PR in order to deploy it on PROD our side.

Do you need any help for this PR ? If not, do you know when it will be merged ?

Thanks

lebaudantoine:
@LNGGlago I'll pick it back up in mid-May after my holidays. Looking forward to making progress then!
I prioritize the white label version of visio first.

LNGGlago:
Thanks for the feedback !

If any help is needed on this PR or white label don't hesitate to ask.

vnghia:
Hello, I would like to ask the progress of this PR ? Any help I can provide to make this PR comes true :D

PR #317, opened 2025-01-22
🩹(frontend) fix allowed hosts by Vite using the Tilt stack

lebaudantoine:
I recently got the error:

> Blocked request. This host ("meet.127.0.0.1.nip.io") is not allowed.
> To allow this host, add "meet.127.0.0.1.nip.io" to server.allowedHosts in vite.config.js.

We apparently need to explicitly allow these domains in Vite's server configuration to prevent host security violations.
This check should be ignored when using HTTPS.

I've made local config changes that might create this issue. @NathanVss, can you reproduce?


NathanVss:
I am not using Tilt in local due to some issues, If you want to merge it, go for it. Or if you absolutely want me to reproduce, tell me I will invest some time in it ;)

PR #318, opened 2025-01-22
🩹(frontend) prevent runtime error when Crisp is not initialized

lebaudantoine:
Prevents runtime error when Crisp chat hasn't initialized before component mount Previously caused crash since we assumed $crisp global was always available.

### DIFF
--- a/src/frontend/src/features/rooms/livekit/components/controls/SupportToggle.tsx
+++ b/src/frontend/src/features/rooms/livekit/components/controls/SupportToggle.tsx
@@ -7,7 +7,9 @@ import { ToggleButtonProps } from '@/primitives/ToggleButton'
 
 export const SupportToggle = ({ onPress, ...props }: ToggleButtonProps) => {
   const { t } = useTranslation('rooms', { keyPrefix: 'controls' })
-  const [isOpened, setIsOpened] = useState($crisp.is('chat:opened'))
+  const [isOpened, setIsOpened] = useState(() => {
+    return window?.$crisp?.is?.('chat:opened') || false
+  })
 
   useEffect(() => {
     if (!Crisp) {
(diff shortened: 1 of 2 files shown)


src/frontend/src/features/rooms/livekit/components/controls/SupportToggle.tsx
```diff
@@ -7,7 +7,9 @@ import { ToggleButtonProps } from '@/primitives/ToggleButton'
 
 export const SupportToggle = ({ onPress, ...props }: ToggleButtonProps) => {
   const { t } = useTranslation('rooms', { keyPrefix: 'controls' })
-  const [isOpened, setIsOpened] = useState($crisp.is('chat:opened'))
+  const [isOpened, setIsOpened] = useState(() => {
+    return window?.$crisp?.is?.('chat:opened') || false
+  })
```
lebaudantoine, line 14: afterthought, should we hide the toggle if crisp isn't defined. I'm scared it would cause layout shifts;

NathanVss (commented):
It seems it may sometimes caused by a race condition due to network speed. I do not find it nice to set it closed only because we are not waiting for $crisp to be fully loaded.

But in the same time I can't find a way to "wait for $crisp to be loaded" in the officiel doc here: https://docs.crisp.chat/guides/chatbox-sdks/web-sdk/dollar-crisp/

WDYT?


EDIT: Maybe "https://docs.crisp.chat/guides/chatbox-sdks/web-sdk/dollar-crisp/"  "[Session has loaded](https://docs.crisp.chat/guides/chatbox-sdks/web-sdk/dollar-crisp/#session-has-loaded)" ?

src/frontend/src/features/rooms/livekit/components/controls/SupportToggle.tsx
NathanVss: My review comment reply to this in an indirect way

lebaudantoine:
It closes #321

PR #334, opened 2025-02-04
✨(front) turn on camera when applying effects

NathanVss:
## Enable camera when applying effects

https://github.com/user-attachments/assets/ecda3e64-56a5-4090-8f72-8718652b1f04

## Show loader if it takes times

`useSyncAfterDelay` allows to enable loading indicators only if the loading takes more than a specific time. It prevent blinking effect when the loading time is nearly instant. 

Loading effects:
- Loader on video
- Disable effect on effects buttons

Please see below

https://github.com/user-attachments/assets/43642734-d2dd-40d6-b46e-7ab59502ae13

### DIFF
--- a/src/frontend/src/primitives/Loader.tsx
+++ b/src/frontend/src/primitives/Loader.tsx
@@ -1,0 +1,45 @@
+import { cva } from '@/styled-system/css'
+
+const loader = cva({
+  base: {
+    borderRadius: '50%',
+    position: 'relative',
+    animation: 'rotate 1s linear infinite',
+    '&:before, &:after': {
+      content: '""',
+      boxSizing: 'border-box',
+      position: 'absolute',
+      inset: '0',
+      borderRadius: '50%',
+      borderStyle: 'solid',
+      borderColor: 'white',
+    },
+    _before: {
+      animation: 'prixClipFix 2s linear infinite',
+    },
+    _after: {
+      borderColor: 'white',
+      animation:
+        'prixClipFix 2s linear infinite, rotate 0.5s linear infinite reverse',
+      inset: 6,
+    },
+  },
+  variants: {
+    size: {
+      small: {
+        width: '24px',
+        height: '24px',
+        '&:before, &:after': {
+          borderWidth: '2px',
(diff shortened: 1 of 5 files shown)


src/frontend/src/styles/index.css
```diff
@@ -38,3 +38,34 @@ body:has(.lk-video-conference) #crisp-chatbox > div > a {
...
+  50% {
+    clip-path: polygon(50% 50%, 0 0, 100% 0, 100% 100%, 100% 100%, 100% 100%);
+  }
+  75% {
+    clip-path: polygon(50% 50%, 0 0, 100% 0, 100% 100%, 0 100%, 0 100%);
+  }
+  100% {
+    clip-path: polygon(50% 50%, 0 0, 100% 0, 100% 100%, 0 100%, 0 0);
+  }
+}
```
lebaudantoine: you can define keyframe using pandas css, keyframes are already defined in `panda.config.ts`

src/frontend/src/primitives/Loader.tsx
```diff
@@ -0,0 +1,45 @@
...
+    },
+  },
+  defaultVariants: {
+    size: 'small',
+  },
+})
+
+export const Loader = () => {
+  return <div className={loader()}></div>
+}
```
lebaudantoine, line 45: you could use the same approach of the `checkbox.ts`, using `styled`

lebaudantoine (approved):

src/frontend/src/primitives/Loader.tsx
NathanVss: Yep, but I don't understand why we is cva and styled used here and there and not only one of those? Maybe I'm missing something ofc

src/frontend/src/styles/index.css
NathanVss: Oh yes, future discussion: why adding complexity in this situation by using a css wrapper in `panda.config.ts` file were CSS is just made for that and works well? Especially for animation where the dead-code argument is not viable 🤔

PR #351, opened 2025-02-19
Private room

lebaudantoine:
## Purpose

**I've updated the demo**

[Quick demo with additional explanation](https://www.loom.com/share/81ac5c4d850b41e693a37fdb33e6f56a)


src/backend/core/api/viewsets.py
```diff
@@ -343,6 +348,86 @@ def stop_room_recording(self, request, pk=None):  # pylint: disable=unused-argum
...
+
+        except LobbyParticipantNotFound:
+            return drf_response.Response(
+                {"message": "Participant not found."},
+                status=drf_status.HTTP_404_NOT_FOUND,
+            )
+        except LobbyParticipantParsingError:
+            return drf_response.Response(
+                {"message": "Participant data are corrupted, removed from the lobby."},
+                status=drf_status.HTTP_500_INTERNAL_SERVER_ERROR,
```
lunika: As you are handling the exception you should not return a 500 IMO. But... I don't find an other status code that could match this case. Maybe a 404 again ? It's like the participant does not exists as you are removing it.

src/backend/core/services/lobby_service.py
```diff
@@ -0,0 +1,271 @@
...
+    """Service for managing participant access through a lobby system.
+
+    Handles participant entry requests, status management, and notifications
+    using Redis cache for state management and LiveKit for real-time updates.
+    """
+
+    @staticmethod
+    def _get_cache_key(room_id: str, participant_key: str) -> str:
+        """Generate cache key for participant(s) data."""
+        return f"{settings.LOBBY_KEY_PREFIX}_{room_id}_{participant_key}"
```
lunika:
```suggestion
        return f"{settings.LOBBY_KEY_PREFIX}_{room_id!s}_{participant_key!s}"
```
`!` is the conversion mark in a fstring and `s` is for string. Doing this you don't have to manage the casting in caller methods, don't need anymore to call `str(room.id)` in the viewset for example.

src/backend/core/services/lobby_service.py
```diff
@@ -0,0 +1,271 @@
...
+                LobbyParticipantStatus.ACCEPTED,
+                settings.LOBBY_ACCEPTED_TIMEOUT,
+            )
+        else:
+            self._update_participant_status(
+                room_id,
+                participant_id,
+                LobbyParticipantStatus.DENIED,
+                settings.LOBBY_DENIED_TIMEOUT,
+            )
```
lunika:
```suggestion
            self._update_participant_status(
                room_id,
                participant_id,
                LobbyParticipantStatus.ACCEPTED if allow_entry else LobbyParticipantStatus.DENIED,
                settings.LOBBY_ACCEPTED_TIMEOUT if allow_entry else settings.LOBBY_DENIED_TIMEOUT,
            )
```

src/backend/core/services/lobby_service.py
```diff
@@ -0,0 +1,271 @@
...
+            return []
+
+        data = cache.get_many(keys)
+
+        return [
+            raw_participant
+            for raw_participant in data.values()
+            if raw_participant
+            and raw_participant.get("status") == LobbyParticipantStatus.WAITING.value
+        ]
```
lunika: It's the only method you are not dealing with the `LobbyParticipant` object and so not validating the data coming from the cache.

lunika (commented):
Nice work.
It seems yo don't check the room access level in the `request_entry`, `allow_participant_to_enter` and `list_waiting_participants` actions

src/backend/core/api/viewsets.py
lunika: One more argument about this, system like sentry open an issue if a 500 is return. But you can't fix it as you already handle it.

src/backend/core/migrations/0011_remove_resource_is_public_room_access_level.py
```diff
@@ -0,0 +1,22 @@
...
+
+    dependencies = [
+        ('core', '0010_alter_resourceaccess_options_alter_user_options'),
+    ]
+
+    operations = [
+        migrations.RemoveField(
+            model_name='resource',
+            name='is_public',
+        ),
```
qbey, line 16: This one is not used yet, right? I just wonder if you need to migrate data or not.

src/backend/core/api/viewsets.py
lebaudantoine: Yes, maybe—but if the backend can't parse participant data, something might be wrong, and we'd appreciate being alerted, don't you think so?

lunika: In that case don't catch it. Let the exception bubble, you will have a 500 response and will be handle by a system like sentry.

lebaudantoine: legit

src/backend/core/services/lobby_service.py
```diff
@@ -0,0 +1,271 @@
...
+            cache_key,
+            participant.to_dict(),
+            timeout=settings.LOBBY_WAITING_TIMEOUT,
+        )
+
+        try:
+            self.notify_participants(room_name=room_id)
+        except LobbyNotificationError:
+            # If room not created yet, there is no participants to notify
+            pass
```
lebaudantoine, line 217: v2: Check for LiveKit room existence to differentiate errors from normal scenarios. If the room exists but the notification wasn't delivered, raise an error; otherwise, treat it as expected behavior.

src/backend/core/services/lobby_service.py
```diff
@@ -0,0 +1,271 @@
...
+        data = cache.get_many(keys)
+
+        return [
+            raw_participant
+            for raw_participant in data.values()
+            if raw_participant
+            and raw_participant.get("status") == LobbyParticipantStatus.WAITING.value
+        ]
+
+    def handle_participant_entry(
```
lebaudantoine, line 262: Should I enforce accepting/denying only for users in the "waiting" status?
Non-critical—there's currently a lot of flexibility.

src/backend/core/models.py
```diff
@@ -80,6 +80,13 @@ class RecordingModeChoices(models.TextChoices):
     TRANSCRIPT = "transcript", _("TRANSCRIPT")
 
 
+class RoomAccessLevel(models.TextChoices):
+    """Room access level choices."""
+
+    PUBLIC = "public", _("Public Access")
+    RESTRICTED = "restricted", _("Restricted Access")
```
lebaudantoine, line 87: v2: Introducing the "trusted" level, allowing authenticated users to join a call without waiting in the lobby.

src/backend/core/services/lobby_service.py
```diff
@@ -0,0 +1,325 @@
...
+        2. If waiting, refresh timeout to maintain position
+        3. If unknown, add to waiting list
+        4. If accepted, generate LiveKit config
+        5. If denied, do nothing.
+        """
+
+        participant_id = self._get_or_create_participant_id(request)
+        participant = self._get_participant(room.id, participant_id)
+
+        if room.is_public:
```
lebaudantoine, line 136: v2: Extracting this check into a method to verify if the room is is_public or if the user is authenticated and the room's access level is "trusted"

lebaudantoine (commented):
In v2 : 
- Add a webhook system to reset cache when the room is done.
- Enhance the base UI.
- Introduce quick actions on Notifications

src/backend/core/api/viewsets.py
```diff
@@ -343,6 +347,88 @@ def stop_room_recording(self, request, pk=None):  # pylint: disable=unused-argum
...
+
+        participant, livekit = lobby_service.request_entry(
+            room=room,
+            request=request,
+            **serializer.validated_data,
+        )
+        response = drf_response.Response({**participant.to_dict(), "livekit": livekit})
+        lobby_service.prepare_response(response, participant.id)
+
+        return response
```
lebaudantoine, line 375: how could protect this endpoint from being spammed by an attacker?
Overall, let's brainstorm together @lunika on possible attack scenarios about the lobby system.

lunika: You can use the throttle from DRF: https://www.django-rest-framework.org/api-guide/throttling/

src/backend/core/services/lobby_service.py
lunika: In the viewset, in the method `allow_participant_to_enter`, if the room is public, you are returning a 404. So I don't understand why you are allowing a participant to reach the lobby service if there is no lobby on a public room (you are returning this message in the http response: `Room has no lobby system.`)

src/backend/core/services/lobby_service.py
```diff
@@ -0,0 +1,325 @@
...
+        except (KeyError, ValueError) as e:
+            logger.exception("Error creating Participant from dict:")
+            raise LobbyParticipantParsingError("Invalid participant data") from e
+
+
+class LobbyService:
+    """Service for managing participant access through a lobby system.
+
+    Handles participant entry requests, status management, and notifications
+    using Redis cache for state management and LiveKit for real-time updates.
```
lunika:
```suggestion
    using cache for state management and LiveKit for real-time updates.
```
Your cache is still agnostic of the backend used.

src/backend/core/tests/rooms/test_api_rooms_lobby.py
```diff
@@ -0,0 +1,401 @@
...
+        id="test-id",
+        username="test_user",
+        status=LobbyParticipantStatus.WAITING,
+        color="#123456",
+    )
+
+    with mock.patch.object(
+        LobbyService, "request_entry", return_value=(mock_participant, None)
+    ):
+        with mock.patch.object(LobbyService, "prepare_response"):
```
lunika:
```suggestion
    with mock.patch.object(
        LobbyService, "request_entry", return_value=(mock_participant, None)
    ), mock.patch.object(LobbyService, "prepare_response"):
```

Any reason to not group them ?

lunika: Also, your mock should not patch the `LobbyService` this way I think. The patch should be made on the object where it is used: https://docs.python.org/3/library/unittest.mock.html#where-to-patch
In your case there is no side effect because this class is only used in the `viewset` module. But the day you start using it somewhere else, you can have unwanted side effects in your test.

lunika: And also I see no real reason to mock the `request_entry` method.

src/backend/core/tests/rooms/test_api_rooms_lobby.py
```diff
@@ -0,0 +1,401 @@
...
+            )
+
+    assert response.status_code == 200
+    assert response.json()["username"] == "test_user"
+    assert response.json()["status"] == "waiting"
+    assert response.json()["livekit"] is None
+
+
+def test_request_entry_public_room():
+    """Entry requests to public rooms should return ACCEPTED status with LiveKit config."""
```
lunika, line 196: In this test you are just asserting that your `mock_participant` has its `status` set to `ACCEPTED` because you are mocking the `request_entry` method. And it is in this method that you decide to return earlier the accepted status. So in fact you are testing a mock :slightly_smiling_face:

src/backend/core/tests/rooms/test_api_rooms_lobby.py
```diff
@@ -0,0 +1,401 @@
...
+            )
+
+    assert response.status_code == 200
+    assert response.json()["username"] == "anonymous"
+    assert response.json()["status"] == "accepted"
+    assert response.json()["livekit"] == {"token": "test-token"}
+
+
+def test_request_entry_authenticated_user_public_room():
+    """While authenticated, entry request to public rooms should get accepted."""
```
lunika, line 244: Same thing here.

src/backend/core/tests/rooms/test_api_rooms_lobby.py
```diff
@@ -0,0 +1,401 @@
...
+
+    assert response.status_code == 400
+
+
+def test_request_entry_room_not_found():
+    """Should return 404 for non-existent room."""
+    client = APIClient()
+
+    response = client.post(
+        "/api/v1.0/rooms/00000000-0000-0000-0000-000000000000/request-entry/",
```
lunika: Why not generating a real UUID ?

src/backend/core/tests/rooms/test_api_rooms_lobby.py
```diff
@@ -0,0 +1,401 @@
...
+    response = client.post(
+        f"/api/v1.0/rooms/{room.id}/enter/",
+        {"participant_id": "test-id", "allow_entry": True},
+    )
+
+    mock_lobby_service.handle_participant_entry.assert_called_once_with(
+        room_id=room.id,
+        participant_id="test-id",
+        allow_entry=True,
+    )
```
lunika: IMO you can do the same thing without this mock. Before sending the request you add in the cache the participant in waiting status.
After the request made, the status in the cache should have changed.
You test everything and your test description will be more accurate. You are "just" asserting that the method `handle_participant_entry` has been called. But not that the status has changed.

src/backend/core/tests/rooms/test_api_rooms_lobby.py
```diff
@@ -0,0 +1,401 @@
...
+    user = UserFactory()
+    # Make user the room owner
+    room.accesses.create(user=user, role="owner")
+
+    client = APIClient()
+    client.force_login(user)
+
+    mock_lobby_service.handle_participant_entry.side_effect = LobbyParticipantNotFound(
+        "Participant not found"
+    )
```
lunika: Same here. You can do the same thing without mock.

src/backend/core/tests/rooms/test_api_rooms_lobby.py
```diff
@@ -0,0 +1,401 @@
...
+            "color": "#123456",
+        },
+        {
+            "id": "participant2",
+            "username": "user2",
+            "status": "waiting",
+            "color": "#654321",
+        },
+    ]
+    mock_lobby_service.list_waiting_participants.return_value = mock_participants
```
lunika: Same here. I think you should not use mock in this test

src/backend/core/tests/rooms/test_api_rooms_lobby.py
```diff
@@ -0,0 +1,401 @@
...
+
+
+@pytest.fixture
+def mock_lobby_service_class(mock_lobby_service):
+    """Mock lobby service class."""
+    with mock.patch(
+        "core.api.viewsets.LobbyService",
+        return_value=mock_lobby_service,
+    ) as mock_lobby_service_class:
+        yield mock_lobby_service_class
```
lunika: IMO you can test everything without mocks.

src/backend/core/services/lobby_service.py
lebaudantoine: If the admin changes the room status while participants are waiting, all waiting participants should automatically enter the room. This behavior should be explicitly reflected in my code.

src/backend/core/api/viewsets.py
lebaudantoine: I've added a throttling, wdyt?

src/backend/core/tests/rooms/test_api_rooms_lobby.py
lebaudantoine: Cache used in test environment doesn't implement `keys` method … would you recommend a workaround?

src/backend/core/tests/rooms/test_api_rooms_lobby.py
lebaudantoine: yes, I removed this mock

PR #382, opened 2025-03-06
Livekit webhook

lebaudantoine:
## Purpose

Clear lobby cache when the room is finished.

## Proposal

Handle nip.io CA cert in livekit, and configure the backend and the livekit to talk.


src/backend/core/api/viewsets.py
```diff
@@ -430,6 +434,30 @@ def list_waiting_participants(self, request, pk=None):  # pylint: disable=unused
         participants = lobby_service.list_waiting_participants(room.id)
         return drf_response.Response({"participants": participants})
 
+    @decorators.action(
+        detail=False,
+        methods=["POST"],
+        url_path="webhooks-livekit",
+        permission_classes=[],
+    )
+    def livekit_webhook(self, request, pk=None):  # pylint: disable=unused-argument
```
```suggestion
    def livekit_webhook(self, request): 
```

As detail is set to False you don't need the `pk` argument

src/backend/core/api/viewsets.py
```diff
@@ -430,6 +434,30 @@ def list_waiting_participants(self, request, pk=None):  # pylint: disable=unused
...
+            return drf_response.Response(
+                {"status": "success"}, status=drf_status.HTTP_200_OK
+            )
+        except LiveKitWebhookError as e:
+            status_code = getattr(
+                e, "status_code", drf_status.HTTP_500_INTERNAL_SERVER_ERROR
+            )
+            return drf_response.Response(
+                {"status": "error", "message": str(e)}, status=status_code
+            )
```
lunika: What do you think about reraising the exception if the status code is 500 ?

PR #421, opened 2025-03-28
📝(docs) describe environmental options for meet backend

berrydenhartog:
## Purpose

Give more insight to which options are available for configuration of the backend

## Proposal

Add a table to the installation docs of all the available environmental options.


lebaudantoine:
Thx for your contribution

PR #423, opened 2025-03-28
Wip fix calendar integration

lebaudantoine:
wip


src/backend/meet/settings.py
```diff
@@ -272,6 +272,11 @@ class Base(Configuration):
                 environ_name="REQUEST_ENTRY_THROTTLE_RATES",
                 environ_prefix=None,
             ),
+            "generation_callback": values.Value(
+                default="150/minute",
```
lunika: That's not a lot? I can make more than 2 requests every second.

src/backend/meet/settings.py
```diff
@@ -528,6 +533,18 @@ class Base(Configuration):
         environ_prefix=None,
     )
 
+    # Integrations
+    ROOM_GENERATION_CALLBACK_CACHE_TIMEOUT = values.PositiveIntegerValue(
+        600,  # 6hrs
```
lunika: You mention 10 minutes in the video.

src/backend/core/services/room_generation.py
```diff
@@ -0,0 +1,44 @@
...
+from django.conf import settings
+from django.core.cache import cache
+
+
+class RoomGeneration:
+    """Wip"""
+
+    def _get_cache_key(self, callback_id):
+        """Wip"""
+        return f"{settings.ROOM_GENERATION_CALLBACK_CACHE_KEY}_{callback_id}"
```
lunika: why using a settings here ?

src/backend/meet/settings.py
lebaudantoine: I've increased it to 600/minutes. The frontend fetch the result every 1 second.

PR #436, opened 2025-04-03
🍱(readme) update readme to rename Visio to La Suite Meet

lebaudantoine:
Visio is a MS product. In the official documentation call our product La Suite Meet.


Maciejrybak4:
**Maciejrybak4**

PR #451, opened 2025-04-10
➕(backend) add `django-lasuite` dependency

qbey:
## Purpose

We work on a common library for our project (https://github.com/suitenumerique/django-lasuite), this is the first step on this.

## Proposal

Use the OIDC backends from the new library.

- [x] Override the `OIDCAuthenticationBackend`
- [x] Update code and tests
- [x] Set the proper version of `django-lasuite` in the `pyproject.toml` file


src/backend/core/authentication/backends.py
```diff
@@ -116,16 +70,16 @@ def signup_to_marketing_email(email):
...
             contact_data = ContactData(
                 email=email, attributes={"VISIO_SOURCE": ["SIGNIN"]}
             )
             marketing_service.create_contact(
                 contact_data, timeout=settings.BREVO_API_TIMEOUT
             )
-        except (ContactCreationError, ImproperlyConfigured, ImportError):
-            pass
 
     def get_existing_user(self, sub, email):
```
lebaudantoine, line 84: why not replacing this method by your implementation and thus merging behavior across projects?

src/backend/meet/settings.py
```diff
@@ -346,8 +346,8 @@ class Base(Configuration):
     SESSION_COOKIE_AGE = 60 * 60 * 12
 
     # OIDC - Authorization Code Flow
```
lebaudantoine, line 357: Should we add the `OIDC_OP_USER_ENDPOINT_FORMAT` to allow consumer customizing the behavior without making a PR on the repo.

src/backend/meet/settings.py
```diff
@@ -417,9 +417,13 @@ class Base(Configuration):
     OIDC_REDIRECT_FIELD_NAME = values.Value(
         "returnTo", environ_name="OIDC_REDIRECT_FIELD_NAME", environ_prefix=None
     )
-    OIDC_USERINFO_FULLNAME_FIELDS = values.ListValue(
-        default=["given_name", "usual_name"],
-        environ_name="OIDC_USERINFO_FULLNAME_FIELDS",
+    USER_OIDC_FIELDS_TO_FULLNAME = values.ListValue(
```
lebaudantoine: why all settings related to OIDC are prefixed by `OIDC` except this one? Is it because you want to differentiate mozilla-django-oidc settings from our custom ones

lebaudantoine (commented):
Great jobs, minor comments!

src/backend/core/authentication/backends.py
qbey: As said IRL, the User management logic is highly related to the project (unicity on email and stuff), so I don't want to make `django-lasuite` too opinionated on this.

PR #458, opened 2025-04-14
Support media link to download recording

lebaudantoine:
# Purpose
Enable users to download their recordings seamlessly.

# Proposal
Taking inspiration from the approach used in Docs, this proposal introduces a new endpoint to support recording downloads. To enable the frontend to render a complete information page, additional metadata has been added to the recording serializer.

Looking ahead, both this page and the media endpoint could eventually be replaced by Fichier for a more unified file access experience.

In a future V2, we’ll introduce automatic expiration of recordings after 7 days — stay tuned!

### DIFF
--- a/.github/workflows/meet.yml
+++ b/.github/workflows/meet.yml
@@ -135,6 +135,9 @@ jobs:
       STORAGES_STATICFILES_BACKEND: django.contrib.staticfiles.storage.StaticFilesStorage
       LIVEKIT_API_SECRET: secret
       LIVEKIT_API_KEY: devkey
+      AWS_S3_ENDPOINT_URL: http://localhost:9000
+      AWS_S3_ACCESS_KEY_ID: meet
+      AWS_S3_SECRET_ACCESS_KEY: password
 
     steps:
       - name: Checkout repository
@@ -151,6 +154,33 @@ jobs:
         with:
           path: "src/backend/core/templates/mail"
           key: mail-templates-${{ hashFiles('src/mail/mjml') }}
+
+      - name: Start MinIO
+        run: |
+          docker pull minio/minio
+          docker run -d --name minio \
+            -p 9000:9000 \
+            -e "MINIO_ACCESS_KEY=meet" \
+            -e "MINIO_SECRET_KEY=password" \
+            -v /data/media:/data \
+            minio/minio server --console-address :9001 /data
+
+      # Tool to wait for a service to be ready
+      - name: Install Dockerize
+        run: |
+          curl -sSL https://github.com/jwilder/dockerize/releases/download/v0.8.0/dockerize-linux-amd64-v0.8.0.tar.gz | sudo tar -C /usr/local/bin -xzv
+
+      - name: Wait for MinIO to be ready
+        run: |
+          dockerize -wait tcp://localhost:9000 -timeout 10s
+
+      - name: Configure MinIO
+        run: |
+          MINIO=$(docker ps | grep minio/minio | sed -E 's/.*\s+([a-zA-Z0-9_-]+)$/\1/')
+          docker exec ${MINIO} sh -c \
+            "mc alias set meet http://localhost:9000 meet password && \
+            mc alias ls && \
+            mc mb meet/meet-media-storage"
 
       - name: Install Python
         uses: actions/setup-python@v5
--- a/compose.yml
+++ b/compose.yml
@@ -15,6 +15,37 @@ services:
     ports:
       - "1081:1080"
 
+  minio:
+    user: ${DOCKER_USER:-1000}
+    image: minio/minio
+    environment:
+      - MINIO_ROOT_USER=meet
+      - MINIO_ROOT_PASSWORD=password
+    ports:
+      - '9000:9000'
+      - '9001:9001'
+    healthcheck:
+      test: [ "CMD", "mc", "ready", "local" ]
+      interval: 1s
+      timeout: 20s
+      retries: 300
+    entrypoint: ""
+    command: minio server --console-address :9001 /data
+    volumes:
+      - ./data/media:/data
+
+  createbuckets:
+    image: minio/mc
+    depends_on:
+      minio:
+        condition: service_healthy
+        restart: true
+    entrypoint: >
+      sh -c "
+      /usr/bin/mc alias set meet http://minio:9000 meet password && \
+      /usr/bin/mc mb meet/meet-media-storage && \
+      exit 0;"
+
   app-dev:
     build:
       context: .
@@ -40,6 +71,7 @@ services:
         - redis
         - nginx
         - livekit
+        - createbuckets
   
   celery-dev:
     user: ${DOCKER_USER:-1000}
@@ -73,6 +105,7 @@ services:
       - postgresql
       - redis
       - livekit
+      - minio
 
   celery:
     user: ${DOCKER_USER:-1000}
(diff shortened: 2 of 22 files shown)


.github/workflows/meet.yml
```diff
@@ -152,6 +155,34 @@ jobs:
...
+          dockerize -wait tcp://localhost:9000 -timeout 10s
+
+      - name: Configure MinIO
+        run: |
+          MINIO=$(docker ps | grep minio/minio | sed -E 's/.*\s+([a-zA-Z0-9_-]+)$/\1/')
+          docker exec ${MINIO} sh -c \
+            "mc alias set meet http://localhost:9000 meet password && \
+            mc alias ls && \
+            mc mb meet/meet-media-storage && \
+            mc version enable meet/meet-media-storage"
```
sampaccoud: you should not need versioning as each recording will be written only once.

compose.yml
```diff
@@ -15,6 +15,38 @@ services:
...
+    image: minio/mc
+    depends_on:
+      minio:
+        condition: service_healthy
+        restart: true
+    entrypoint: >
+      sh -c "
+      /usr/bin/mc alias set meet http://minio:9000 meet password && \
+      /usr/bin/mc mb meet/meet-media-storage && \
+      /usr/bin/mc version enable meet/meet-media-storage && \
```
sampaccoud: ditto

src/backend/core/models.py
```diff
@@ -576,6 +578,26 @@ def is_savable(self) -> bool:
...
+        """Check if the recording is in a saved state."""
+        return self.status == RecordingStatusChoices.SAVED
+
+    @property
+    def extension(self):
+        """Get recording extension based on its mode."""
+        extensions = {
+            RecordingModeChoices.TRANSCRIPT: FileExtension.OGG.value,
+            RecordingModeChoices.SCREEN_RECORDING: FileExtension.MP4.value,
+        }
```
sampaccoud, line 592: Belongs to enums.py as it is a constant?

src/backend/core/enums.py
```diff
@@ -2,9 +2,21 @@
...
 
+import re
+
 from django.conf import global_settings, settings
 from django.utils.translation import gettext_lazy as _
 
+UUID_REGEX = (
+    r"[a-fA-F0-9]{8}-[a-fA-F0-9]{4}-[a-fA-F0-9]{4}-[a-fA-F0-9]{4}-[a-fA-F0-9]{12}"
+)
+FILE_EXT_REGEX = r"[a-zA-Z0-9]{1,10}"
```
sampaccoud, line 13: you may want to restrict to ogg and mp4 if it's what you expect?
```suggestion
FILE_EXT_REGEX = r"^(ogg|mp4)$"
```
or 
```suggestion
FILE_EXT_REGEX = r"^(" + "|".join(re.escape(ext.value) for ext in FileExtension) + r")$"
```

src/backend/core/api/viewsets.py
```diff
@@ -617,3 +620,85 @@ def on_storage_event_received(self, request, pk=None):  # pylint: disable=unused
...
+
+        url_params = self._auth_get_url_params(
+            enums.RECORDING_STORAGE_URL_PATTERN, parsed_url.path
+        )
+
+        user = request.user
+        recording_id = url_params["recording_id"]
+
+        extension = url_params["extension"]
+        if extension not in [item.value for item in FileExtension]:
```
sampaccoud, line 681: I would say this works:
```suggestion
        if extension not in FileExtension.values:
```

src/backend/core/tests/test_models_recording.py
```diff
@@ -216,3 +217,71 @@ def test_models_recording_worker_id_very_long():
...
+def test_models_recording_is_saved_false_active():
+    """Test is_saved property returns False for ACTIVE status."""
+    recording = RecordingFactory(status=RecordingStatusChoices.ACTIVE)
+    assert recording.is_saved is False
+
+
+def test_models_recording_is_saved_false_initiated():
+    """Test is_saved property returns False for INITIATED status."""
+    recording = RecordingFactory(status=RecordingStatusChoices.INITIATED)
+    assert recording.is_saved is False
```
sampaccoud, line 273: could be parametrized as well?

sampaccoud (approved):
I only have neatpicks and anyway GTM!
Nice feature coming!

src/backend/core/api/viewsets.py
```diff
@@ -617,3 +620,85 @@ def on_storage_event_received(self, request, pk=None):  # pylint: disable=unused
...
+        except models.Recording.DoesNotExist as e:
+            raise drf_exceptions.NotFound("No recording found for this event.") from e
+
+        if extension != recording.extension:
+            raise drf_exceptions.NotFound("No recording found with this extension.")
+
+        abilities = recording.get_abilities(user)
+
+        if not abilities["retrieve"]:
+            logger.debug("User '%s' lacks permission for attachment", user)
```
```suggestion
            logger.debug("User '%s' lacks permission for attachment", user.id)
```
You should not log the user. The string representation use the email, this is a personal data and you don't want to have them in your logs.

src/backend/core/enums.py
lunika: But remove the `^` and `$` as this regex is part of a bigger one.

src/backend/core/models.py
```diff
@@ -576,6 +578,26 @@ def is_savable(self) -> bool:
...
+    @property
+    def is_saved(self) -> bool:
+        """Check if the recording is in a saved state."""
+        return self.status == RecordingStatusChoices.SAVED
+
+    @property
+    def extension(self):
+        """Get recording extension based on its mode."""
+        extensions = {
+            RecordingModeChoices.TRANSCRIPT: FileExtension.OGG.value,
```
lunika, line 590: `TRANSCRIPT` seems not accurate here.

src/backend/core/models.py
lebaudantoine: I prefer to keep it close to where it's used, since no other code elements rely on it.

src/backend/core/enums.py
lebaudantoine: Actually, I preferred keeping the file extension validation in the viewset ! You don't think so?

src/backend/core/tests/test_models_recording.py
lebaudantoine: gree! But didn’t we decide not to factorize the tests? Not a big deal in my opinion

src/backend/core/models.py
lebaudantoine: why?

lebaudantoine: might outside the scope of my PR

src/backend/core/api/viewsets.py
lebaudantoine: Actually its a debug log, but true, I took it from docs https://github.com/suitenumerique/docs/blob/ecd06560c6dab354b0c42249ee7af31ebfd8ed95/src/backend/core/api/viewsets.py#L1249

src/backend/core/models.py
lunika: Yes it is not in the scope of this PR indeed

src/backend/core/api/viewsets.py
lebaudantoine: 🤔 `.values` doesn't exist on a Python Enum class

src/backend/core/api/viewsets.py
lebaudantoine: Fixed!

src/backend/core/enums.py
sampaccoud: 👌

PR #461, opened 2025-04-14
✨(backend) add email notifications for screen recordings

lebaudantoine:
Implement backend method to send email notifications when screen recordings are ready for download. Enables users to be alerted when their recordings are available. Frontend implementation to follow in upcoming commits.

This service is triggered by the storage hook from Minio.

Add minimal unit test coverage for notification service, addressing previous lack of tests in this area. The notification service was responsible for calling the unstable summary service feature, which was developped way too quickly.

The email template has been reviewed by a LLM, to make it user-friendly and crystal clear.

### DIFF
--- a/src/backend/core/recording/event/notification.py
+++ b/src/backend/core/recording/event/notification.py
@@ -1,8 +1,13 @@
 """Service to notify external services when a new recording is ready."""
 
 import logging
+import smtplib
 
 from django.conf import settings
+from django.core.mail import send_mail
+from django.template.loader import render_to_string
+from django.utils.translation import get_language, override
+from django.utils.translation import gettext_lazy as _
 
 import requests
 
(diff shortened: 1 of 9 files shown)


src/backend/core/recording/event/notification.py
```diff
@@ -33,6 +36,65 @@ def notify_external_services(self, recording):
...
+
+        domain = Site.objects.get_current().domain
+
+        owner_accesses = (
+            models.RecordingAccess.objects.select_related("user")
+            .filter(
+                role=models.RoleChoices.OWNER,
+                recording_id=recording.id,
+            )
+            .all()
```
lunika:
```suggestion
```

I don't think it is needed.

src/backend/core/recording/event/notification.py
```diff
@@ -33,6 +36,65 @@ def notify_external_services(self, recording):
...
+    @staticmethod
+    def _notify_user_by_email(recording) -> bool:
+        """
+        Send an email notification to recording owners when their recording is ready.
+
+        The email includes a direct link that redirects owners to a dedicated download
+        page in the frontend where they can access their specific recording.
+        """
+
+        domain = Site.objects.get_current().domain
```
lunika: I know it is used in other projects but IMO it's not a good idea to use it only to have the domain. Using a settings for this is easier, configurable (see the mock used in your tests).

lebaudantoine: agree

PR #464, opened 2025-04-16
Invite people by email

lebaudantoine:
## Purpose

Feature requested by some users. 
Being able to send email invitation from a room. (Inspired by proprietary solutions)

## Proposal

Frontend will display a pop-in with email selection / search. 
Frontend will be responsible to call the /invite endpoint to trigger the email, and add the invited people to the room as member, if they exist in our user base (we don't have pending invitation mechanism as on docs).

This frontend feature will be added in a next PR. 

This new endpoint is intentionally minimal and simply handles sending emails to all invited users. I haven’t added unit tests for the service, as I’m relying on integration tests at the viewset level instead. Feel free to challenge the implementation — I shipped it fast

### DIFF
--- a/src/backend/core/api/viewsets.py
+++ b/src/backend/core/api/viewsets.py
@@ -39,6 +39,7 @@ from core.recording.worker.factories import (
 from core.recording.worker.mediator import (
     WorkerServiceMediator,
 )
+from core.services.invitation import InvitationService
 from core.services.livekit_events import (
     LiveKitEventsService,
     LiveKitWebhookError,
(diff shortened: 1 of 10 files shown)


src/backend/core/api/viewsets.py
```diff
@@ -495,6 +496,44 @@ def creation_callback(self, request):
...
+        to one or more recipients, inviting them to join the specified room.
+        """
+
+        room = self.get_object()
+
+        serializer = serializers.RoomInviteSerializer(data=request.data)
+        serializer.is_valid(raise_exception=True)
+
+        emails = serializer.validated_data.get("emails")
+        emails = list({email.lower() for email in emails})
```
lunika: I don't think it is needed to lower it. IMO, as the serializer has validated it is a valid email, you should use them as it.

src/backend/core/api/viewsets.py
```diff
@@ -495,6 +496,44 @@ def creation_callback(self, request):
...
+
+        try:
+            InvitationService().invite_to_room(
+                room=room, sender=request.user, emails=emails
+            )
+        except InvitationError:
+            return drf_response.Response(
+                {"error": "Could not send invitation"},
+                status=drf_status.HTTP_500_INTERNAL_SERVER_ERROR,
+            )
```
lunika: you should not catch it and let it bubble to be catch by sentry.

src/backend/core/api/viewsets.py
lebaudantoine: Tbh, I had mixed feelings about the lower(). Christophe Ninucci pointed out that a lowercased email should be treated the same as one with uppercase letters. I’ll remove it, but just noting that this could allow duplicates with different casing to slip through.

I'll add it in the test

PR #476, opened 2025-04-19
🔧(backend) add MEET_DATA_DIR env variable to specify the data dir

fflorent:
## Purpose

When building Visio without Docker, the data directory may not (and probably should not) be placed at the root of the filesystem. 

## Proposal

This commit offers a way to specify where the data directory is placed, and when omitted it is assumed to be at `/data`.

### DIFF
--- a/src/backend/meet/settings.py
+++ b/src/backend/meet/settings.py
@@ -23,7 +23,7 @@ from sentry_sdk.integrations.logging import ignore_logger
 
 # Build paths inside the project like this: BASE_DIR / 'subdir'.
 BASE_DIR = os.path.dirname(os.path.dirname(os.path.abspath(__file__)))
-DATA_DIR = os.path.join("/", "data")
+DATA_DIR = os.environ.get("MEET_DATA_DIR", os.path.join("/", "data"))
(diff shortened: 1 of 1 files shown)


src/backend/meet/settings.py
```diff
@@ -23,7 +23,7 @@
 
 # Build paths inside the project like this: BASE_DIR / 'subdir'.
 BASE_DIR = os.path.dirname(os.path.dirname(os.path.abspath(__file__)))
-DATA_DIR = os.path.join("/", "data")
+DATA_DIR = os.environ.get("MEET_DATA_DIR", os.path.join("/", "data"))
```
fflorent, line 26: I am not very sure how to prefix this variable. Alternatively, I may remove the prefix (`DATA_DIR`) or prefix with `VISIO` (`VISIO_DATA_DIR`)

lebaudantoine:
what do you think of https://github.com/suitenumerique/meet/pull/480 ?

fflorent:
@lebaudantoine, sounds good, I close this PR in favor to the other one you have opened.

PR #480, opened 2025-04-22
🔧(backend) make data directory location configurable via env var

lebaudantoine:
Add environment variable to control data directory location when building outside of Docker. Improves flexibility for non-containerized deployments where storing data at filesystem root is inappropriate or undesirable.

the env variable will be prefixed by `DJANGO_` @fflorent .


fflorent:
Right, so `DJANGO_DATA_DIR`?

PR #508, opened 2025-04-29
YWH reports 

lebaudantoine:
Respond to YesWeHack reports:
- Report #YWH-PGM14336-4
- Report #YWH-PGM14336-5

Goal:
- Ensure members are not trusted by default; do not expose the list of accessible resources to room members.
- Disable the browsable API to prevent leakage of sensitive information.

### DIFF
--- a/src/backend/meet/settings.py
+++ b/src/backend/meet/settings.py
@@ -263,6 +263,9 @@ class Base(Configuration):
             "rest_framework.parsers.JSONParser",
             "nested_multipart_parser.drf.DrfNestedParser",
         ],
+        "DEFAULT_RENDERER_CLASSES": [
+            "rest_framework.renderers.JSONRenderer",
+        ],
         "EXCEPTION_HANDLER": "core.api.exception_handler",
         "DEFAULT_PAGINATION_CLASS": "rest_framework.pagination.PageNumberPagination",
         "PAGE_SIZE": 20,
--- a/src/backend/core/api/serializers.py
+++ b/src/backend/core/api/serializers.py
@@ -120,7 +120,7 @@ class RoomSerializer(serializers.ModelSerializer):
         role = instance.get_role(request.user)
         is_admin = models.RoleChoices.check_administrator_role(role)
 
-        if role is not None:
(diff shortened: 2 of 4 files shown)


src/backend/meet/settings.py
```diff
@@ -263,6 +263,9 @@ class Base(Configuration):
             "rest_framework.parsers.JSONParser",
             "nested_multipart_parser.drf.DrfNestedParser",
         ],
+        "DEFAULT_RENDERER_CLASSES": [
+            "rest_framework.renderers.JSONRenderer",
+        ],
```
qbey, line 268: @sampaccoud & @lunika we need to do that on all our projects to prevent Django from rendering a form which have access to all objects when using BrowsableAPIRenderer (which is enabled by default)

sampaccoud: Since this is the default setting, are we sure this is enough and the other renderers can't be called via a querystring?

qbey: The default is 

``` python
    'DEFAULT_RENDERER_CLASSES': [
        'rest_framework.renderers.JSONRenderer',
        'rest_framework.renderers.BrowsableAPIRenderer',
    ],
```
So if `BrowsableAPIRenderer` is not listed, it cannot be used, except if the view `renderer_classes` adds it manually.

sampaccoud: Maybe add a non-regression test with explanations in the docstring why we should not allow this?

qbey: IMHO, we should even disable all `DEFAULT_RENDERER_CLASSES` : we don't use it and our APIs are not supposed to be used in a browser.

lunika: Oh we have the same report on docs but it was not accepted. As it is a feature of DRF and the project is open source, we are not leaking any sensitive information.
BTW I think we should remove indeed, in production, the browsableAPIRenderer.

lebaudantoine: we were leaking sensitive information in the select options, rendered in the browsableAPIRenderer

PR #536, opened 2025-05-13
✨(backend) add 10-digit PIN codes on rooms for telephony

lebaudantoine:
## Product

We aim at enabling users to join rooms via SIP telephony by:
- Dialing the SIP trunk number
- Entering the room's PIN followed by '#'

The PIN code must be generated immediately upon room creation. This allows the
room owner to invite participants in advance with the correct PIN and dial-in
number, ensuring a smooth join experience.

## Collision

With 10-digit PINs (10^10 combinations) and a large number of rooms (e.g., 1M), collisions become statistically inevitable. A retry mechanism helps reduce the chance of repeated collisions but doesn't eliminate the overall risk.

With 100K generated PINs, the probability of at least one collision exceeds 39%, due to the birthday paradox.

## Scale

To scale safely, we’ll later propose using multiple trunks. Each trunk will handle a separate PIN namespace, and the combination of trunk_id and PIN will ensure uniqueness. Room assignment will be evenly distributed across trunks to balance load and minimize collisions.

Following XP principles, we’ll ship the simplest working version of this feature. The goal is to deliver value quickly without over-engineering.

We’re not solving scaling challenges we don’t currently face. Our production load is around 10,000 rooms — well within safe limits for the initial implementation.

## Discussion points

- The `while` loop should be reviewed. Should we add rate limiting for failed attempts?
- A systematic existence check before `INSERT` is more costly for a rare event and doesn't prevent race conditions, whereas retrying on integrity errors is more efficient overall.
- Should we add logging or monitoring to track and analyze collisions?

I tried to balance performance and simplicity while ensuring the robustness of the PIN generation process.


src/backend/core/models.py
```diff
@@ -414,6 +425,26 @@ def is_public(self):
         """Check if a room is public"""
         return self.access_level == RoomAccessLevel.PUBLIC
 
+    def save(self, *args, **kwargs):
+        """Generate a unique n-digit pin code for new rooms."""
+        if not self.pin_code:
+            self.pin_code = self.generate_unique_pin_code(length=settings.ROOM_PIN_CODE_LENGTH)
+        return super().save(*args, **kwargs)
```
lebaudantoine:
```suggestion
        super().save(*args, **kwargs)
```

src/backend/core/models.py
```diff
@@ -383,6 +384,14 @@ class Room(Resource):
         verbose_name=_("Visio room configuration"),
         help_text=_("Values for Visio parameters to configure the room."),
     )
+    pin_code = models.CharField(
+        max_length=100,
```
lebaudantoine:
```suggestion
        max_length=None,
```

src/backend/meet/settings.py
```diff
@@ -572,6 +572,13 @@ class Base(Configuration):
         environ_prefix=None,
     )
 
+    # SIP Telephony
+    ROOM_PIN_CODE_LENGTH = values.PositiveIntegerValue(
+        10,  # this value cannot exceed 100 digits due to database constraints
```
lebaudantoine:
```suggestion
        10,
```

src/backend/core/models.py
```diff
@@ -393,6 +402,14 @@ class Meta:
     def __str__(self):
         return capfirst(self.name)
 
+    def save(self, *args, **kwargs):
+        """Generate a unique n-digit pin code for new rooms."""
+        if not self.pin_code:
```
lebaudantoine: telephony flag

src/backend/core/models.py
```diff
@@ -414,6 +431,20 @@ def is_public(self):
...
+    @staticmethod
+    def generate_unique_pin_code(length):
+        """Generate a unique n-digit PIN code"""
+
+        if length < 1:
+            raise ValueError("PIN code length must be at least 1")
+
+        max_value = 10**length
+
+        while True:
```
lebaudantoine: range instead of while

src/backend/core/models.py
```diff
@@ -414,6 +431,20 @@ def is_public(self):
...
+            pin_code = str(secrets.randbelow(max_value)).zfill(length)
+            if not Room.objects.filter(pin_code=pin_code).exists():
+                return pin_code
```
lebaudantoine, line 449: return None if no success

src/backend/core/models.py
```diff
@@ -407,13 +424,38 @@ def clean_fields(self, exclude=None):
...
+        for _ in range(settings.ROOM_TELEPHONY_PIN_MAX_RETRIES):
+            pin_code = str(secrets.randbelow(max_value)).zfill(length)
+            if not Room.objects.filter(pin_code=pin_code).exists():
+                return pin_code
+
+        logger.warning(
+            "Failed to generate unique PIN code of length %s after %s attempts",
+            length,
+            settings.ROOM_TELEPHONY_PIN_MAX_RETRIES,
+        )
```
lebaudantoine, line 456: This is temporary, as soon as we add observability, we would refactor this part. I'll add a comment.

src/backend/core/tests/test_models_rooms.py
```diff
@@ -175,3 +181,169 @@ def test_models_rooms_is_public_property():
...
+
+    # Assert default value is 10 for collision reasons
+    assert len(room.pin_code) == 10
+
+    settings.ROOM_TELEPHONY_PIN_LENGTH = 5
+
+    room = RoomFactory()
+
+    # Assert custom size
+    assert len(room.pin_code) == 5
```
lebaudantoine, line 215: I wanted to test the number of SQL queries, but it wasn't trivial. Let's discuss it.

src/backend/meet/settings.py
```diff
@@ -572,6 +572,23 @@ class Base(Configuration):
...
+    ROOM_TELEPHONY_PIN_LENGTH = values.PositiveIntegerValue(
+        10,
+        environ_name="ROOM_TELEPHONY_PIN_LENGTH",
+        environ_prefix=None,
+    )
+    ROOM_TELEPHONY_PIN_MAX_RETRIES = values.PositiveIntegerValue(
+        3,
+        environ_name="ROOM_TELEPHONY_PIN_MAX_RETRIES",
+        environ_prefix=None,
+    )
```
lebaudantoine: (I am rusty in probabilities)

With 100K rooms generated and 10-digits pin codes, there is a 39% chance of collision when generating a new pin code. 

With 3 retries, the probability of not-being able to generate a pin code would be 5%. 
Should we increase it to 5 retries? chance are down to less than 1%.

src/backend/core/models.py
```diff
@@ -383,6 +384,14 @@ class Room(Resource):
...
         help_text=_("Values for Visio parameters to configure the room."),
     )
+    pin_code = models.CharField(
+        max_length=None,
+        unique=True,
+        blank=True,
+        null=True,
+        verbose_name=_("Room PIN code"),
+        help_text=_("Unique n-digit code that identifies this room in telephony mode."),
+    )
```
lebaudantoine, line 394: I considered adding data validation for this field during the `clean_fields` call. In this initial draft, there's currently no validation applied when the room object is updated.

To keep things lean, I’d avoid introducing strict constraints at this stage. We can progressively enforce stronger validation once the feature is in production. It's also worth noting that this field won't be exposed for editing through the API.

src/backend/meet/settings.py
lunika: 5 retries seems not so big. I would be in favor to increase it.

lunika (approved):

PR #541, opened 2025-05-14
Adding additional support in the install documentation in regards to the kind method

keda82:
## Purpose

The documentation was missing some bits and pieces in regards to the kind method of deploying a local test/dev instance

## Proposal

Added the following

- the requirement of kubectl
- that the script has to be downloaded from the repo and made executable
- that mkcert, docker etc needs to be executable from the user context running the script
- the kind method only works for local access


lebaudantoine:
Thanks a lot for your contribution @keda82, excellent as a first PR. I've merged your update in https://github.com/suitenumerique/meet/pull/547, and renamed the commit to pass the CI.

PR #554, opened 2025-05-19
Fix the path of the backend ingress.

bmassemin:
## Purpose
The current path is `/api/v`, and it doesn't work with `ingress-nginx`. 

## Proposal
I'm not sure if other ingress controllers work with this prefix, but changing it to `/api/` will work for `ingress-nginx` and likely for others as well.

### DIFF
--- a/src/helm/meet/templates/ingress.yaml
+++ b/src/helm/meet/templates/ingress.yaml
@@ -60,7 +60,7 @@ spec:
               serviceName: {{ include "meet.frontend.fullname" . }}
               servicePort: {{ .Values.frontend.service.port }}
             {{- end }}
-          - path: /api/v
+          - path: /api/
             {{- if semverCompare ">=1.18-0" $.Capabilities.KubeVersion.GitVersion }}
             pathType: Prefix
             {{- end }}
@@ -96,7 +96,7 @@ spec:
               serviceName: {{ include "meet.frontend.fullname" $ }}
               servicePort: {{ $.Values.frontend.service.port }}
             {{- end }}
-          - path: /api/v
+          - path: /api/
             {{- if semverCompare ">=1.18-0" $.Capabilities.KubeVersion.GitVersion }}
             pathType: Prefix
             {{- end }}


rouja:
Hi,

Could you show us how to reproduce the issue? Which ingress are you using exactly? This one: https://docs.nginx.com/nginx-ingress-controller/ or this one: https://github.com/kubernetes/ingress-nginx?

Because we are using the community one and it works as it does.

bmassemin:
I use https://github.com/kubernetes/ingress-nginx with the chart version 4.11.5.

With `/api/v`, calls are routed to the frontend. With `/api/`, calls are correctly routed to the backend.

Here is my Helm configuration:
```
ingress:
  enabled: true
  className: nginx
  host: meet.[REDACTED]
  annotations:
    cert-manager.io/cluster-issuer: zerossl      
  tls:
    enabled: true
```

rouja:
Hi,

I investigated this a bit. The official documentation states that this shouldn't work: https://kubernetes.io/docs/concepts/services-networking/ingress/#examples 
However, in our case it works because the ingress generates two locations:

```
		location ~* "^/api/v/" {
			
			set $namespace      "meet";
			set $ingress_name   "meet";
			set $service_name   "meet-backend";
			set $service_port   "80";
			set $location_path  "/api/v";
```
So this particular location block will not work.
However, there is a second one:
```
		location ~* "^/api/v" {
			
			set $namespace      "meet";
			set $ingress_name   "meet";
			set $service_name   "meet-backend";
			set $service_port   "80";
			set $location_path  "/api/v";
```

I'm not sure why you don't have the second location block in your case. To avoid any confusion in the future, we will proceed with merging your PR.

rouja (approved):

lebaudantoine:
Thx @bmassemin for your contribution.

lebaudantoine:
We have a (strict) git linter, as an organization rule @suitenumerique/developers. I'll rewrite your commit, and bump the chart version, to merge your contribution.

lebaudantoine:
Merged in #572

lebaudantoine:
Merged in #572

PR #560, opened 2025-05-20
✏️(backend) fix typo in FRONTEND_IS_SILENT_LOGIN_ENABLED env var

### LINKED ISSUE #556: Typo in environment variable name: FRONTEND_IS_SILENT_LOGING_ENABLED should be LOGIN
Basically just that. See https://github.com/suitenumerique/meet/blob/c8772bb1ad9b9df97416a48c182b05518c24d39d/src/backend/meet/settings.py#L317.

Not sure if it's OK to break this without backwards compatibility so not sending a PR.

lebaudantoine:
Correct spelling in environment variable name as identified by @K900. This is technically a breaking change for existing deployments using this setting, but acceptable as we haven't released an official version yet. Will personally notify known users of this setting about the change to minimize disruption.

It closes #556

### DIFF
--- a/src/backend/meet/settings.py
+++ b/src/backend/meet/settings.py
@@ -314,7 +314,7 @@ class Base(Configuration):
             False, environ_name="FRONTEND_SILENCE_LIVEKIT_DEBUG", environ_prefix=None
         ),
         "is_silent_login_enabled": values.BooleanValue(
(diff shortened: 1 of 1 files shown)


rouja (approved):

fflorent (commented):
LGTM

PR #561, opened 2025-05-20
🔒️(oidc) disable OIDC authentication on API

qbey:
## Purpose

Our authentication flow uses the Django authentication which creates a session for the User. Then the session is used to make API calls, therefore we don't need to accept OIDC tokens directly on the API.

Accepting the OIDC token on the API can allow to bypass the "resource server mode" which allows to restrict provided information according to the Service Provider which makes the request.

## Proposal

- [x] remove OIDC backend authentication on API


qbey:
This changes 401 -> 403 (see https://www.django-rest-framework.org/api-guide/authentication/#sessionauthentication) so we won't merge it now.

PR #575, opened 2025-05-26
Enhance the UX admin

lebaudantoine:
## Purpose

Debugging user is painful. Make my life easier.

## Proposal

Enhance UX while using the Django admin.


src/backend/core/admin.py
```diff
@@ -120,4 +125,24 @@ class RecordingAdmin(admin.ModelAdmin):
     """Recording admin interface declaration."""
 
     inlines = (RecordingAccessInline,)
-    list_display = ("id", "status", "room", "created_at", "worker_id")
+    search_fields = ["status", "=id", "worker_id", "room__slug", "=room__id"]
+    list_display = ("id", "status", "room", "get_owner", "created_at", "worker_id")
```
qbey, line 129: I'm afraid it will generate N+1 queries

qbey (approved):

PR #578, opened 2025-05-27
Catch request timeout when adding contact to brevo

lebaudantoine:
Handle unhandled exceptions to prevent UX impact while authenticating. Marketing email operations are optional and should not disrupt core functionality.

Wrapping the timeout exception in a ContactCreationError allows the authentication backend to ignore it appropriately.

My first implementation was imperfect, raising errors in sentry.


src/backend/core/services/marketing.py
```diff
@@ -120,7 +121,7 @@ def create_contact(self, contact_data: ContactData, timeout=None) -> dict:
 
         try:
             response = contact_api.create_contact(contact, **api_configurations)
-        except brevo_python.rest.ApiException as err:
+        except (brevo_python.rest.ApiException, requests.exceptions.ReadTimeout) as err:
             logger.exception("Failed to create contact in Brevo")
```
lebaudantoine: Should I raise a warning to avoid spam sentry? @qbey

qbey: It should be a warning because I'm not sure a developer can do anything we the error occurs + the stacktrace won't help you.

PR #586, opened 2025-06-03
Remove some DINUM-specific UI

manuhabitela:
## Purpose

Clean the UI of some DINUM specific things:

- the Marianne font should not be used for instances that are not French government
- the footer is French government specific and should basically not be included by default
- the French government-related logos (Marianne, motto) in the header should not be shown by default

## Proposal

- remove the Marianne font
- remove the government-related logos
- add a way to load a custom CSS file on a meet instance. It allows some very basic customization of specific DOM nodes for the header, to add back the government-related logos on a specific instance. And allows to change the color variables and other design tokens configured through panda-css
- hide the French government-specific footer behind an env var.

This proposal is not (yet) ideal as we still host some DINUM-specific assets, and having the footer still in this code doesn't make that much sense; but, it's better than nothing :)

Here an example of `custom.css` we could use on the DINUM instance:

```css
.Header-beforeLogo {
  display: block;
}

.Header-beforeLogo::before {
  content: '';
  display: block;
  background-image: url(/assets/marianne.svg);
  background-position: 0 -0.046875rem;
  background-size: 2.0625rem 0.84375rem;
  height: 0.75rem;
  margin-bottom: 0.1rem;
  width: 2.0625rem;
}

.Header-beforeLogo::after {
  content: '';
  display: block;
  background-image: url(/assets/gouvernement.svg), url(/assets/devise.svg);
  background-repeat: no-repeat, no-repeat;
  background-size: 108.8px 10px, 40px 29px;
  background-position: 0 3px, 0 18.9px;
  width: 108.8px;
  height: 48px;
  margin-top: 0.25rem;
}
```


manuhabitela:
I also just added a way to not use the ProConnect UI for the login button.

@lebaudantoine I think you can check it out now, this should be ready.

manuhabitela:
I also removed the source sans default font that wasn't really useful. People can use their own font if they want with custom CSS until we decide more carefully on some default design.

lebaudantoine:
Hey @manuhabitela, thanks a lot for your contribution, much appreciated!
I’ve added a few commits to help further generalize the distribution of La Suite Meet. Let me know what you think!
The logo still references “Visio”, so I’ll need to update that separately; once that’s done, we should be all set.

src/frontend/index.html
```diff
@@ -4,7 +4,7 @@
     <meta charset="UTF-8" />
     <link rel="icon" type="image/svg+xml" href="/assets/icon.svg" />
     <meta name="viewport" content="width=device-width, initial-scale=1.0" />
-    <title>Visio</title>
+    <title>La Suite Meet</title>
```
manuhabitela: As you say in the commit message I also think this is not ideal.

One way to resolve this for some people would be to use build-time env vars, but it would not solve the issue for the pre-built docker image.

We could have `Loading…` as default title but it wouldn't be translated, so not an option.

We could have an empty `<title></title>` as default, but it means having a tab that is not titled at all while the config api call is not done. Not ideal, but acceptable?

So, is this a bad idea?:

- in the frontend, reuse the FRONTEND_APP_TITLE env var, or have some dedicated env var, for the default title
- use it as default here: `<title>%FRONTEND_APP_TITLE%</title>`
- keep the system setting the `<title>` from the backend config

This allows people building their own frontend to have the title already set before any api calls. For others, it still allows changing the title, without having "La Suite Meet" shown for a sec.

src/frontend/src/hooks/useAppTitle.ts
```diff
@@ -0,0 +1,10 @@
+import { useEffect } from 'react'
+
+export const useAppTitle = (title?: string, fallback = 'La Suite Meet') => {
```
manuhabitela: I would not have any fallback in order to prevent any flash in case the title is customized

src/frontend/src/components/NotFoundScreen.tsx
```diff
@@ -10,8 +10,7 @@ export const NotFoundScreen = () => {
     <Screen layout="centered">
       <CenteredContent title={t('notFound.heading')} withBackButton>
         <Text centered>
-          {t('notFound.body')}{' '}
-          <Bold>https://visio.numerique.gouv.fr/xxx-yyyy-zzz.</Bold>
+          {t('notFound.body')} <Bold>{window.origin}/xxx-yyyy-zzz.</Bold>
```
manuhabitela, line 13: Thanks! Way better than before. Not blocking for merging but this makes me think: is meet something we intend to be hosted only at the root of a domain? I guess at some point we should support the case were we want to host it in a subfolder.

manuhabitela (commented):
:ok_hand: :ok_hand: :ok_hand:

src/frontend/src/components/NotFoundScreen.tsx
mosa-riel: Why not have the baseurl configurable on deployment?

src/frontend/index.html
fflorent: I am not sure "La Suite Meet" is meant for community instances. Maybe just `Meet`?

lebaudantoine: I think I would have to stick to LaSuite Meet @fflorent for trademark reasons.

src/frontend/src/hooks/useAppTitle.ts
lebaudantoine: I removed this part, and used the env variable approach.

src/frontend/src/components/NotFoundScreen.tsx
lebaudantoine: Actually, to support such case @manuhabitela I would need to rework [apiUrl.ts ](https://github.com/suitenumerique/meet/blob/d1bb414cb4643593f396a20e6280058ff8872b0d/src/frontend/src/api/apiUrl.ts#L4)… I've opened an issue to track it [here](https://github.com/suitenumerique/meet/issues/610).

PR #597, opened 2025-06-16
support rnn suppression

lebaudantoine:
## Purpose

Description...

## Proposal

Description...

- [] item 1...
- [] item 2...


lebaudantoine:
Tried in sandbox env with few users!

PR #602, opened 2025-06-23
⬆️(docker) upgrade backend image to python 3.13

lebaudantoine:
Python 3.13 is now stable, our libraries are compatible with it. We also
upgrade the alpine version and node one used in the backend.

### DIFF
--- a/src/summary/Dockerfile
+++ b/src/summary/Dockerfile
@@ -1,4 +1,4 @@
-FROM python:3.12-slim AS builder
+FROM python:3.13-slim AS builder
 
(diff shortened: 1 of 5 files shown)


src/summary/Dockerfile
```diff
@@ -1,12 +1,12 @@
-FROM python:3.12-slim AS builder
+FROM python:3.13-slim AS builder
```
lunika: why not having `FROM python:3.13-slim as base` anf then use `FROM base as ...` ?

PR #603, opened 2025-06-24
Security auditors feedbacks

lebaudantoine:
Please check commits.

### DIFF
--- a/src/backend/core/api/serializers.py
+++ b/src/backend/core/api/serializers.py
@@ -1,5 +1,7 @@
 """Client serializers for the Meet core app."""
 
+import uuid
+
 from django.utils.translation import gettext_lazy as _
 
 from rest_framework import serializers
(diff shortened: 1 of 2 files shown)


src/backend/core/api/serializers.py
```diff
@@ -219,6 +221,14 @@ class ParticipantEntrySerializer(serializers.Serializer):
     participant_id = serializers.CharField(required=True)
     allow_entry = serializers.BooleanField(required=True)
 
+    def validate_participant_id(self, value):
+        """Validate that the participant_id is a valid UUID hex string."""
+        try:
+            uuid.UUID(hex=value)
```
lunika: you should not validate that it's UUID version 4 ?

lebaudantoine: I did not want to refactor the code, as the point of the auditor was especially to enforce a stricter validation of user entry. I can in an upcoming commit if you feel it's more natural to manipulate uuid v4 string instead of their hex representation.

lunika (approved):

PR #614, opened 2025-07-01
Support telephony

lebaudantoine:
Wip.

## Summary by CodeRabbit

* **New Features**
  * Exposed room PIN code in API responses and frontend room data.
  * Introduced telephony dispatch rule management for rooms, enabling direct SIP call routing using room ID and PIN code.
  * Added telephony configuration and phone number display with PIN code in room info UI.
  * Included localized labels for phone call and PIN code in multiple languages.
  * Added a React hook to access telephony configuration and phone number formatting utilities.
  * Added telephony settings for phone number and default country configuration.
  * Added phone number parsing and PIN code formatting utilities.
  * Added new frontend dependency for phone number parsing.

* **Bug Fixes**
  * Improved handling of falsy metadata values when parsing raised hand state in rooms.

* **Chores**
  * Enhanced error handling and logging for telephony rule operations and room lifecycle events.
  * Expanded test coverage for telephony services and room lifecycle event handling.

### DIFF
--- a/src/backend/core/services/telephony.py
+++ b/src/backend/core/services/telephony.py
@@ -1,0 +1,124 @@
+"""Telephony service for managing SIP dispatch rules for room access."""
+
+from logging import getLogger
+
+from asgiref.sync import async_to_sync
+from livekit.api import TwirpError
+from livekit.protocol.sip import (
+    CreateSIPDispatchRuleRequest,
+    DeleteSIPDispatchRuleRequest,
+    ListSIPDispatchRuleRequest,
+    SIPDispatchRule,
+    SIPDispatchRuleDirect,
+)
+
+from core import utils
+
+logger = getLogger(__name__)
+
+
+class TelephonyException(Exception):
+    """Exception raised when telephony operations fail."""
+
+
+class TelephonyService:
+    """Service for managing participant access through the telephony system (SIP)."""
+
+    def _rule_name(self, room_id):
+        """Generate the rule name for a room based on its ID."""
+        return f"SIP_{str(room_id)}"
+
+    @async_to_sync
+    async def create_dispatch_rule(self, room):
+        """Create a SIP inbound dispatch rule for direct room routing.
+
+        Configures telephony to route incoming SIP calls directly to the specified room
+        using the room's ID and PIN code for authentication.
+        """
+
+        direct_rule = SIPDispatchRule(
+            dispatch_rule_direct=SIPDispatchRuleDirect(
+                room_name=str(room.id), pin=str(room.pin_code)
+            )
+        )
+
+        request = CreateSIPDispatchRuleRequest(
+            rule=direct_rule, name=self._rule_name(room.id)
+        )
+
+        lkapi = utils.create_livekit_client()
+
+        try:
+            await lkapi.sip.create_sip_dispatch_rule(create=request)
+        except TwirpError as e:
+            logger.exception(
+                "Unexpected error creating dispatch rule for room %s", room.id
+            )
+            raise TelephonyException("Could not create dispatch rule") from e
+
+        finally:
+            await lkapi.aclose()
+
+    async def _list_dispatch_rules_ids(self, room_id):
+        """List SIP dispatch rule IDs for a specific room.
+
+        Fetches all existing SIP dispatch rules and filters them by room name
+        since LiveKit API doesn't support server-side filtering by 'room_name'.
+        This approach is acceptable for moderate scale but may need refactoring
+        for high-volume scenarios.
+
+        Note:
+            Feature request for server-side filtering: livekit/sip#405
+        """
+
+        lkapi = utils.create_livekit_client()
+
+        try:
+            existing_rules = await lkapi.sip.list_sip_dispatch_rule(
+                list=ListSIPDispatchRuleRequest()
+            )
+        except TwirpError as e:
+            logger.exception("Failed to list dispatch rules for room %s", room_id)
+            raise TelephonyException("Could not list dispatch rules") from e
+        finally:
+            await lkapi.aclose()
+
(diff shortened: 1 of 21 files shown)


lebaudantoine:
@coderabbitai review

lebaudantoine:
@coderabbitai full review

src/backend/core/services/telephony.py
```diff
@@ -0,0 +1,122 @@
...
+        except TwirpError as e:
+            logger.exception(
+                "Unexpected error creating dispatch rule for room %s", room.id
+            )
+            raise TelephonyException("Could not create dispatch rule") from e
+
+        finally:
+            await lkapi.aclose()
+
+    async def _list_dispatch_rules_ids(self, room_id):
```
lunika, line 62: Don't really the interest to have this method is used only once.

src/backend/core/services/telephony.py
```diff
@@ -0,0 +1,122 @@
...
+            await lkapi.aclose()
+
+        if not existing_rules or not existing_rules.items:
+            return []
+
+        return [
+            existing_rule.sip_dispatch_rule_id
+            for existing_rule in existing_rules.items
+            if existing_rule.name == self._rule_name(room_id)
+        ]
```
lunika: you can compute `self._rule_name(room_id)` once and use it in this for loop

src/backend/core/tests/services/test_livekit_events.py
```diff
@@ -56,42 +58,145 @@ def test_initialization(
...
+@mock.patch.object(TelephonyService, "delete_dispatch_rule")
+def test_handle_room_finished_clears_cache_and_deletes_dispatch_rule(
+    mock_delete_dispatch_rule, mock_clear_cache, service
+):
+    """Should clear lobby cache and delete telephony dispatch rule when room finishes."""
     mock_room_name = uuid.uuid4()
+    mock_data = mock.MagicMock()
+    mock_data.room.name = str(mock_room_name)
+
+    service._handle_room_finished(mock_data)
```
lunika, line 71: Private method should not be tested directly but through the public methods using them.

src/backend/core/services/telephony.py
lebaudantoine: It encapsulates the logic the current LiveKit SDK is missing. When they would release it, we would have to just trash this method.

Without this method, I was ending having all the logic in a single try catch to avoid re-instantiating Lkapi, which wasn't ideal to me.

src/backend/core/services/telephony.py
lebaudantoine: fixed

src/backend/core/tests/services/test_livekit_events.py
lebaudantoine: Testing through the public API, would force me to mock parts of the private method implementations, which breaks encapsulation again, or I simply don’t test them at all.

I feel it's important to test those internal implementations directly because the process is sequential: if one step fails, the next doesn’t run. That’s a flaw in the current design.

If I rely solely on testing through the public methods, the tests end up full of boilerplate. I get your point, and I can definitely see the value in sticking strictly to OOP principles. But part of me wanted to apply those principles while keeping things pragmatic. At the end of the day, working code matters more to me than theoretical rules, and making sure each part is thoroughly tested is my top priority.

It might be a pain to refactor, but since the public method essentially acts as a dispatcher, it feels more natural to focus tests on the individual pieces being dispatched.

Another reason is that handle_room_finished and handle_room_started have different internal logic. One requires an existing room in database, while the other only needs to validate a UUID for the room name. I want to test those specific behaviors.

I’m not doing strict TDD or adhering fully to classical OOP, just trying to find a balance that works.

No strong opinion on this comment, if you think it's really better for the technical debt of the project to refactor the tests, I'll do it.

src/backend/core/services/telephony.py
```diff
@@ -0,0 +1,122 @@
...
+        """Delete all SIP inbound dispatch rules associated with a specific room."""
+
+        rules_ids = await self._list_dispatch_rules_ids(room_id)
+
+        if not rules_ids:
+            logger.info("No dispatch rules found for room %s", room_id)
+            return False
+
+        if len(rules_ids) > 1:
+            logger.error("Multiple dispatch rules found for room %s", room_id)
```
qbey, line 108: It's more like a warning, because the code manages it well, no?

src/backend/core/services/telephony.py
qbey: I guess the fix would be removing the call from the loop:

```suggestion
        _rule_name = self._rule_name(room_id)
        return [
            existing_rule.sip_dispatch_rule_id
            for existing_rule in existing_rules.items
            if existing_rule.name == _rule_name
        ]
```

src/frontend/src/api/useConfig.ts
```diff
@@ -30,6 +30,11 @@ export interface ApiConfig {
     available_modes?: RecordingMode[]
     expiration_days?: number
   }
+  telephony: {
+    enabled: boolean
+    phone_number?: string
```
qbey, line 35: For my understanding, should it be `phone_number?: string` or `phone_number: string | null`?

src/frontend/src/features/rooms/api/ApiRoom.ts
```diff
@@ -14,6 +14,7 @@ export type ApiRoom = {
   id: string
   name: string
   slug: string
+  pin_code: string
```
qbey, line 17: Always a string, or might be null?

src/frontend/src/api/useConfig.ts
lebaudantoine: I think it's fine, phone_number is either undefined or a string.

PR #634, opened 2025-07-11
Push ynrrlluwnnro

CobaltCause:
When I was setting up Meet I just guessed a random value and then had to fish out the real URL from the logs of my OIDC provider to configure it properly. I would like to obviate the need for doing this for other people trying to set up Meet by documenting it *somewhere*. Not sure if this is the best place though.


CobaltCause:
I was about to leave a very different comment but it looks like it's getting pre-emptively addressed, so instead I'll just say thank you for that

lebaudantoine:
Thank you for your contribution!  The documentation is clearer.

I apologize for the automated review from Coderabbit, we were running some internal tests and had it misconfigured. We've since disabled it to avoid any confusion.

Hope to see you around again. If you have any thoughts or feedback, even nitpicks, feel free to share. Everything help.

Issue #632, opened 2025-07-11
Support PKCE for OIDC


CobaltCause:
I am not super familiar with this stuff but my understanding is that PKCE has security benefits so it would probably be good to support it. My main reference is this: <https://kanidm.github.io/kanidm/stable/frequently_asked_questions.html#why-is-disabling-pkce-considered-insecure>.

PR #635, opened 2025-07-14
Add environment variable for PKCE setting

K900:
Defaulting to false for now to keep compatibility.

## Purpose

PKCE is good and everyone should use it.

## Proposal

Use PKCE.


CobaltCause (commented):
This would fix #632. You might also consider updating `docs/installation.md` to include this new env var.

Tom-Hubrecht:
While at it, you could also add support for `OIDC_PKCE_CODE_CHALLENGE_METHOD` and `OIDC_PKCE_CODE_VERIFIER_SIZE` c.f. https://github.com/suitenumerique/docs/pull/886/files

lebaudantoine:
thank you for your contribution @K900! I'll update the `docs/installation.md`

PR #644, opened 2025-07-18
🐛(backend) fix lobby notification type error breaking participant alerts

lebaudantoine:
Correct data type issue that was preventing lobby notifications from being sent to other participants in the room.


lebaudantoine:
@coderabbitai full review

PR #660, opened 2025-07-31
:construction_worker:(ci) use variables in pipelines for easy fork an…

f-ld:
## Purpose

Introduce some variables in pipelines to make easier customisation of forks for testing

## Proposal

It is common to fork a project and test it locally. But for whoever who wants to test it at scale, it could be interesting to use environment variables and change only what is required to get started.

Two variables are introduced:

- one for the docker registry (and not try to push docker images to `lasuite/...`)
- one for the main branch from which pipelines should be executed (and keep `main` clean of changes for easy sync from upstream while keeping some changes local).

### DIFF
--- a/.github/workflows/docker-hub.yml
+++ b/.github/workflows/docker-hub.yml
@@ -1,19 +1,21 @@
 name: Docker Hub Workflow
 run-name: Docker Hub Workflow
 
+env:
+  DOCKER_USER: 1001:127
+  DOCKER_CONTAINER_REGISTRY: lasuite
+  MAIN_BRANCH: main
+
 on:
   workflow_dispatch:
   push:
     branches:
-      - 'main'
+      - '${{ MAIN_BRANCH }}'
     tags:
       - 'v*'
   pull_request:
     branches:
-      - 'main'
-
-env:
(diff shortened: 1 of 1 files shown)


lebaudantoine:
Hi,
Thanks for your contribution!

I noticed that your PR description mentions two environment variables, but only one appears to have been introduced in the actual diff. Could you double-check if the description is up to date?

Also, following a suggestion from @rouja (we discussed your proposal in person), we'd like to propose adding a variable for the registry URL (e.g., docker.io) to allow for more configurability.

Thanks again!

f-ld:
> Hi, Thanks for your contribution!
> 
> I noticed that your PR description mentions two environment variables, but only one appears to have been introduced in the actual diff. Could you double-check if the description is up to date?

Sorry, I expected to do more that what I finally committed and description ended up to be out of sync

> Also, following a suggestion from @rouja (we discussed your proposal in person), we'd like to propose adding a variable for the registry URL (e.g., docker.io) to allow for more configurability.

Ok, I updated my commit that way.

> Thanks again!

Sorry for the late reply, I've been quite side tracked lately. Thank you for looking at it.

PR #662, opened 2025-08-01
🐛(frontend) use feature detection for adaptiveStream and dynacast

lebaudantoine:
Replace hardcoded true values with supportsAdaptiveStream() and supportsDynacast() checks. LiveKit SDK supports broad browser range but requires specific APIs - modern features need explicit compatibility checks.

Prevents enabling unsupported WebRTC features on incompatible browsers, which could led to a poor user experience.

One alternative solution could be to install polyfills.


lebaudantoine:
@coderabbitai full review

PR #663, opened 2025-08-01
🩹(backend) allow enforcing WSS protocol to resolve browser compatibility

lebaudantoine:
The LiveKit API URL is necessary to interact with the API. It uses https
protocol.

Eplicit wss protocol is necessary in Websocket constructor for some
older browsers.

This resolves critical compatibility issues with legacy browsers
(notably Firefox <124, Chrome <125, Edge <125) that lack support
for HTTPS URLs in the WebSocket() constructor. Without explicit WSS
URLs, WebSocket signaling connections may fail, crash, or be blocked
entirely in these environments.

The setting is optional and defaults to the current behavior when
not specified, ensuring zero breaking changes for existing deployments.


lebaudantoine:
@coderabbitai full review

lebaudantoine:
reviewed by @rouja discussed irl

PR #664, opened 2025-08-01
🩹(frontend) avoid video glitch on the prejoin screen while initiating a processor

lebaudantoine:
We've introduced simplifications that improve performance, enhance ux, and contribute to the overall perception of experience quality.

Previously, our processor handling was overly complex. LiveKit allows us to set a processor before starting the local video preview track, which eliminates the black blink glitch that appeared when loading the join component with a default processor.

This change prevents the unnecessary stopping and restarting of the local video track.

I'm glad this issue is now resolved.

We also simplified component communication by avoiding props drilling. Now, we use a single flag to indicate when the user is ready to enter the room. This significantly reduces the complexity of props passed through components.


lebaudantoine:
@coderabbitai full review

lebaudantoine:
example : 

https://github.com/user-attachments/assets/e2637dbb-f4ac-4bd6-9422-3fcad0716271

PR #668, opened 2025-08-08
🚀(paas) add PaaS deployment scripts, tested on Scalingo

sylvinus:
This PR adds a new deployment option for Heroku-like platforms based on buildpacks, like Scalingo.

Similar deployments options are already supported in Messages, Drive and soon Docs https://github.com/suitenumerique/docs/pull/1020

### DIFF
--- a/bin/buildpack_postfrontend.sh
+++ b/bin/buildpack_postfrontend.sh
@@ -1,0 +1,15 @@
+#!/bin/bash
+
+set -o errexit    # always exit on error
+set -o pipefail   # don't ignore exit codes when piping output
+
+echo "-----> Running post-frontend script"
+
+# Move the frontend build to the nginx root and clean up
+mkdir -p build/
+mv src/frontend/dist build/frontend-out
+
+mv src/backend/* ./
+mv src/nginx/* ./
+
+echo "3.13" > .python-version
--- a/src/backend/meet/settings.py
+++ b/src/backend/meet/settings.py
@@ -11,11 +11,12 @@ https://docs.djangoproject.com/en/3.1/ref/settings/
 """
 
 import json
-from os import path
+from os import environ, path
 from socket import gethostbyname, gethostname
 
 from django.utils.translation import gettext_lazy as _
 
+import dj_database_url
 import sentry_sdk
 from configurations import Configuration, values
 from lasuite.configuration.values import SecretFileValue
@@ -86,7 +87,9 @@ class Base(Configuration):
 
     # Database
     DATABASES = {
-        "default": {
+        "default": dj_database_url.config()
+        if environ.get("DATABASE_URL")
+        else {
             "ENGINE": values.Value(
                 "django.db.backends.postgresql_psycopg2",
                 environ_name="DB_ENGINE",
--- a/src/backend/setup.py
+++ b/src/backend/setup.py
@@ -1,0 +1,7 @@
+#!/usr/bin/env python
+"""Setup file for the impress module. All configuration stands in the setup.cfg file."""
+# coding: utf-8
+
+from setuptools import setup
+
+setup()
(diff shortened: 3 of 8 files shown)


src/backend/setup.py
```diff
@@ -0,0 +1,7 @@
+#!/usr/bin/env python
+"""Setup file for the impress module. All configuration stands in the setup.cfg file."""
```
lebaudantoine, line 2:
```suggestion
"""Setup file for the meet module. All configuration stands in the setup.cfg file."""
```

src/backend/meet/settings.py
```diff
@@ -86,7 +87,9 @@ class Base(Configuration):
 
     # Database
     DATABASES = {
-        "default": {
+        "default": dj_database_url.config()
+        if environ.get("DATABASE_URL")
```
lebaudantoine, line 91: What do you think about using Django config values (e.g., `values.Value(...)`) to keep settings.py consistent?

bin/buildpack_postfrontend.sh
```diff
@@ -0,0 +1,15 @@
...
+echo "-----> Running post-frontend script"
+
+# Move the frontend build to the nginx root and clean up
+mkdir -p build/
+mv src/frontend/dist build/frontend-out
+
+mv src/backend/* ./
+mv src/nginx/* ./
+
+echo "3.13" > .python-version
```
lebaudantoine, line 15: nit: missing EOF

src/nginx/servers.conf.erb
```diff
@@ -0,0 +1,50 @@
+
```
lebaudantoine, line 1: Quick review, I agree with @lunika's [comment](https://github.com/suitenumerique/docs/pull/1020/files#r2221477451). The src/nginx folder name doesn’t really communicate the purpose of this Nginx config. Could we at least add a short README in this folder to explain the overall PaaS deployment?

Also, just a note: `/media` and `/media-auth` are required in Meet for handling recordings!

src/backend/meet/settings.py
lebaudantoine: Nit: I prefer `DATABASE_`, but for consistency, should we stick with `DB_` for all database-related settings?

src/backend/setup.py
lebaudantoine: is the addition of this file necessary for Scalingo? 

https://setuptools.pypa.io/en/latest/userguide/pyproject_config.html

for reference :  

> If compatibility with legacy builds or versions of tools that don’t support certain packaging standards (e.g. [PEP 517](https://peps.python.org/pep-0517/) or [PEP 660](https://peps.python.org/pep-0660/)), a simple setup.py script can be added to your project [[1]](https://setuptools.pypa.io/en/latest/userguide/pyproject_config.html#setupcfg-caveats) (while keeping the configuration in pyproject.toml):

Discussed IRL with @lunika

src/nginx/servers.conf.erb
sylvinus: @lebaudantoine what about making a `deploy` folder and moving files from `docker` in there? https://github.com/suitenumerique/docs/pull/1020#discussion_r2287862143

src/backend/meet/settings.py
sylvinus: `DATABASE_URL` is a standard name in most PaaS platforms https://doc.scalingo.com/databases/postgresql/getting-started/connecting https://devcenter.heroku.com/changelog-items/438

src/backend/setup.py
sylvinus: I know it fails without it. Not sure if there's another easier way...

src/nginx/servers.conf.erb
lebaudantoine: sounds good

PR #672, opened 2025-08-12
♻️(frontend) refactor preview track to avoid unnecessary resets

lebaudantoine:
Replace usePreviewTrack approach with manual track management inspired by deprecated LiveKit implementation that only toggles what's needed instead of resetting entire preview track on config changes.

Previous approach created new videoTrack when updating audio track, causing video preview to blink black during audio configuration changes.

It was also acquiring video streams even when disabled.

New implementation:
* Initializes tracks once with user permissions for both devices
* Manually handles muting/unmuting and device updates
* Preserves muted state when changing device IDs
* Uses exact device constraints on Chrome for proper device selection

Prevents unnecessary track recreation and eliminates preview blinking.


lebaudantoine:
@coderabbitai full review

PR #679, opened 2025-08-23
👷(ci) add pip caching and upgrade setup-python action to v5

lebaudantoine:
Implement pip dependency caching across all CI jobs requiring package installation and upgrade actions/setup-python from v4 to v5.

The setup-python action is able to cache the dependencies and reuse this cache while the pyproject file has not changed. It is easy to setup, just the package manager used has to be declared in the cache settings


lebaudantoine:
@coderabbitai full review

PR #680, opened 2025-08-28
Refactor admin

lebaudantoine:
## Purpose

This PR marks a **significant improvement in room admin capabilities**. It strengthens security, improves permission handling, and provides admins with new tools to manage meetings more effectively. These changes lay the groundwork for future enhancements and bring us closer to enterprise-grade conferencing solutions.

## Proposal

This update introduces new admin actions and security improvements.
Admins can now disable user permissions in real time:

https://github.com/user-attachments/assets/ce1479df-dc32-4ca6-9701-9b0c58300fb0

https://github.com/user-attachments/assets/3b1c1d18-9e0e-4b74-8bab-d5d482253ac3

The room configuration is also persisted using the legacy `configuration` attribute. This ensures that when generating new participant tokens, users automatically receive the correct permissions.

At the request of @arnaud-robin, we have introduced a “mute all” feature for admins. This allows the host to mute all unprivileged users in a meeting at once, giving better control during disruptions:

https://github.com/user-attachments/assets/890adad7-f6a4-44e1-8e37-2adb0ca8c35f

The host of the meeting is now clearly identified in the participant list. In addition, a new participant menu has been added, giving admins the ability to exclude or remove a participant from a meeting. This action also clears the lobby cache properly:

https://github.com/user-attachments/assets/8e1d1d6a-5d28-4863-b277-d1fe9d01b639

A major security enhancement has also been implemented: not all participants are LiveKit room admins anymore. This addresses security audit recommendations and reduces unnecessary privileges. As a result, not everyone can mute another participant as before—this functionality is now restricted to admins. It will be reintroduced soon as a configurable option.


lebaudantoine:
@coderabbitai full review

lebaudantoine:
@coderabbitai full review

src/backend/core/api/serializers.py
```diff
@@ -192,75 +210,85 @@ class StartRecordingSerializer(serializers.Serializer):
...
+
+    track_sid = serializers.CharField(
+        max_length=255, help_text="LiveKit track SID to mute"
+    )
+
+
+class UpdateParticipantSerializer(BaseParticipantsManagementSerializer):
+    """Validate participant update data."""
+
+    metadata = serializers.JSONField(
```
qbey:
```suggestion
    metadata = serializers.DictField(
```

src/backend/core/api/serializers.py
```diff
@@ -192,75 +210,85 @@ class StartRecordingSerializer(serializers.Serializer):
...
+    )
+    name = serializers.CharField(
+        max_length=255,
+        required=False,
+        allow_blank=True,
+        allow_null=True,
+        help_text="Display name for the participant",
+    )
+
+    def validate(self, attrs):
```
qbey, line 280: You may want to validate the `permission` field against `livekit.api.ParticipantPermission` (and or return a ParticipantPermission directly?)

src/backend/core/api/serializers.py
```diff
@@ -1,6 +1,6 @@
 """Client serializers for the Meet core app."""
 
-import uuid
+# pylint: disable=W0223
```
qbey: NIT: Using pylint name instead of codes makes it easier to understand => we prefer the name :)

```suggestion
# pylint: disable=abstract-method
```

src/backend/core/api/viewsets.py
```diff
@@ -530,6 +535,107 @@ def invite(self, request, pk=None):  # pylint: disable=unused-argument
...
+    def remove_participant(self, request, pk=None):  # pylint: disable=unused-argument
+        """Remove a participant from the room."""
+        room = self.get_object()
+
+        serializer = serializers.BaseParticipantsManagementSerializer(data=request.data)
+        serializer.is_valid(raise_exception=True)
+
+        try:
+            ParticipantsManagement().remove(
+                room_name=str(room.id),
```
qbey: NIT: even if the Python object ID is the same as the database object Primary Key, I personally prefer to use `room.pk` instead of `room.id`

src/backend/core/api/serializers.py
qbey: Or, you may catch the bad request directly in the view, when calling `update_participant` ? This would keep the serializer "livekit agnostic" (if that's what you want)

src/backend/core/api/viewsets.py
lebaudantoine: done over all the project's sources

src/backend/core/api/serializers.py
lebaudantoine: done all over the project's sources

src/backend/core/api/serializers.py
lebaudantoine: Fixed!

PR #683, opened 2025-09-02
♻️(tilt) use hem dev-backend chart

rouja:
Use the helm dev-backend chart to share the backend across all projects


lebaudantoine:
@coderabbitai full review

src/helm/helmfile.yaml
```diff
@@ -65,3 +68,61 @@ releases:
...
+          username: dinum
+          password: password
+          bucket: meet-media-storage
+          size: 1Gi
+      - keycloak:
+          enabled: true
+          image: quay.io/keycloak/keycloak:20.0.1
+          name: keycloak
+          #serviceNameOverride: keycloak
+          hostname: keycloak.127.0.0.1.nip.io
```
lunika, line 111: You should add a suffix here to not collapse with other projects running in a different namespace.

Also, we should find a solution to have one keycloak for all our projects in the dev stack ? Maybe running it in a dedicated namespace ? And we use a convention to use this namespace accross all our projects ? We will need this once we will implement the resource server protocol on our projects.

PR #690, opened 2025-09-08
♻️(backend) replace Django permissions with custom feature flag decorator

lebaudantoine:
Refactor feature flag mechanism from Django permission classes to custom decorator that returns 404 Not Found when features are disabled instead of exposing API structure through permission errors.

Improves security by preventing information disclosure about disabled features and provides more appropriate response semantics. Custom decorator approach is better suited for feature toggling than Django's permission system which is designed for authorization.

### DIFF
--- a/src/backend/core/api/feature_flag.py
+++ b/src/backend/core/api/feature_flag.py
@@ -1,0 +1,45 @@
+"""Feature flag handler for the Meet core app."""
+
+from functools import wraps
+
+from django.conf import settings
+from django.http import Http404
+
+
+class FeatureFlag:
+    """Check if features are enabled and return error responses."""
+
+    FLAGS = {
+        "recording": "RECORDING_ENABLE",
+        "storage_event": "RECORDING_STORAGE_EVENT_ENABLE",
+        "subtitle": "ROOM_SUBTITLE_ENABLED",
+    }
+
+    @classmethod
+    def flag_is_active(cls, flag_name):
+        """Check if a feature flag is active."""
+
+        setting_name = cls.FLAGS.get(flag_name)
+
+        if setting_name is None:
+            return False
+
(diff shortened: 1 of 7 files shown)


lebaudantoine:
@coderabbitai full review

lebaudantoine:
@coderabbitai full review

src/backend/core/api/feature_flag.py
```diff
@@ -0,0 +1,43 @@
...
+    }
+
+    @classmethod
+    def check(cls, feature_name):
+        """Return 404 response if feature is disabled, None if enabled."""
+        setting_name = cls.FEATURES.get(feature_name)
+        if not getattr(settings, setting_name, False):
+            return drf_response(
+                {"error": "Not found."}, status=drf_status.HTTP_404_NOT_FOUND
+            )
```
lunika: What about raising a `Http404` ? https://github.com/django/django/blob/main/django/http/response.py#L718

src/backend/core/api/feature_flag.py
```diff
@@ -0,0 +1,41 @@
...
+        """Decorator to check feature at the beginning of endpoint methods."""
+
+        if feature_name not in cls.FEATURES:
+            raise ValueError(f"Unknown feature flag: {feature_name}")
+
+        def decorator(view_func):
+            @wraps(view_func)
+            def wrapper(self, request, *args, **kwargs):
+                if response := cls.check(feature_name):
+                    return response
```
lunika:
```suggestion
                cls.check(feature_name):
```

lunika: The `check` function is now really small, you can maybe remove it and copy the 3 lines inside the `wrapper` function ?

PR #716, opened 2025-09-22
Setup langfuse

GuittenyMartin:
## Setup first observability object with langfuse

Create traces for process-audio and summarize-transcription
Create span for minio_get_object
Declare generations for LLM calls
Add variables in end.d, config and pyproject.toml


lebaudantoine:
@coderabbitai full review

env.d/development/summary.dist
```diff
@@ -21,3 +21,7 @@ WEBHOOK_URL="https://configure-your-url.com"
 
 POSTHOG_API_KEY="your-posthog-key"
 POSTHOG_ENABLED="False"
+
+LANGFUSE_SECRET_KEY="your-secret-key"
+LANGFUSE_PUBLIC_KEY="your-public-key"
+LANGFUSE_HOST="https://cloud.langfuse.com"
```
lebaudantoine: nitpick: add a blank line at the end of your file.

src/summary/summary/core/celery_worker.py
```diff
@@ -319,7 +350,7 @@ def process_audio_transcribe_summarize_v2(
     metadata_manager.capture(task_id, settings.posthog_event_success)
 
     if (
-        analytics.is_feature_enabled("summary-enabled", distinct_id=sub)
+        analytics.is_feature_enabled("summary-enabled", distinct_id=1756993468466)
```
lebaudantoine: wrongly committed

src/summary/summary/core/config.py
```diff
@@ -75,6 +75,11 @@ class Settings(BaseSettings):
     task_tracker_redis_url: str = "redis://redis/0"
     task_tracker_prefix: str = "task_metadata:"
 
+    # Langfuse
+    langfuse_secret_key: Optional[str] = None
+    langfuse_public_key: Optional[str] = None
+    langfuse_host: Optional[str] = "https://cloud.langfuse.com"
```
lebaudantoine: It might be better to keep these as None by default, so the SDK can fall back to its own defaults. Defining them here suggests they’re meant to override via environment variables, but since they’re not passed explicitly to the client, it’s a bit confusing.

src/summary/summary/core/celery_worker.py
```diff
@@ -222,6 +227,21 @@ def process_audio_transcribe_summarize_v2(
     logger.info("Notification received")
     logger.debug("filename: %s", filename)
 
+    try:
+        langfuse.update_current_trace(
+            user_id=sub or email,
```
lebaudantoine, line 246: I guess sub is always set
```suggestion
            user_id=sub,
```
It applies for next calls

**Question:** Should we use `owner_id` as the distinct identifier in Langfuse to stay consistent with how we handle it in PostHog?

src/summary/summary/core/celery_worker.py
GuittenyMartin: Resolved

PR #717, opened 2025-09-22
🐛(summary) fix featureflag

GuittenyMartin:
fix the featureflag to allow chosen users to use
the summary feature by using the owner_id instead of sub because 
users are identified with owner_id as distinct_id in posthog. It's necessary
to use the distinct_id to check the feature flag.

### DIFF
--- a/src/backend/core/recording/event/notification.py
+++ b/src/backend/core/recording/event/notification.py
@@ -131,8 +131,8 @@ class NotificationService:
         if not owner_access:
             logger.error("No owner found for recording %s", recording.id)
             return False
-
         payload = {
+            "owner_id": str(owner_access.user.id),
             "filename": recording.key,
             "email": owner_access.user.email,
             "sub": owner_access.user.sub,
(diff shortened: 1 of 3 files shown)


src/backend/core/recording/event/notification.py
```diff
@@ -131,8 +132,9 @@ def _notify_summary_service(recording):
         if not owner_access:
             logger.error("No owner found for recording %s", recording.id)
             return False
-
+        owner_id = owner_access.user.id
         payload = {
+            "owner_id": str(owner_id) if isinstance(owner_id, UUID) else owner_id,
```
lebaudantoine:
```suggestion
            "owner_id": str(owner_id),
```

lebaudantoine (approved):
**LGTM 👍**
You just need to lint your summary files with **ruff**.

* Create a virtual environment
* Install the Python dev dependencies
* Run:


  ```bash
  ruff check . --fix
  ruff format .
  ```

The CI is currently failing on the frontend due to the Trivy scan. This is unrelated to your changes, so no worries.

PR #732, opened 2025-09-30
🐛(summary) fix webhook minio

GuittenyMartin:
fix the makefile to start the minio webhook
to start a transcription when an audio is posted

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -45,7 +45,6 @@ COMPOSE_RUN         = $(COMPOSE) run --rm
 COMPOSE_RUN_APP     = $(COMPOSE_RUN) app-dev
 COMPOSE_RUN_CROWDIN = $(COMPOSE_RUN) crowdin crowdin
 WAIT_DB             = @$(COMPOSE_RUN) dockerize -wait tcp://$(DB_HOST):$(DB_PORT) -timeout 60s
-WAIT_MINIO          = @$(COMPOSE_RUN) dockerize -wait tcp://minio:9000 -timeout 60s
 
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -141,6 +142,11 @@ run-summary: ## start only the summary application and all needed services
 run:
 run: ## start the wsgi (production) and development server
 	@$(MAKE) run-backend
+	@echo "Waiting for app-dev to be up..."
+	@$(WAIT_APP)
+	@echo "Configuring MinIO webhook…"
+	@$(MAKE) minio-webhook-setup
+	@$(MAKE) minio-webhook-setup
```
lebaudantoine: I think this might be a duplicate. Should we add a cleanup step to prevent the issue you experienced, where local data gets created and affects the hook’s recreation?

lebaudantoine: Why are we calling the webhook setup in the run command? It seems more like an initialization step that should only be called once during project bootstrapping. What do you think?

Makefile
```diff
@@ -60,7 +61,7 @@ MINIO_ENDPOINT      ?= http://127.0.0.1:9000
 MINIO_ACCESS_KEY    ?= meet
 MINIO_SECRET_KEY    ?= password
 MINIO_BUCKET        ?= meet-media-storage
-MINIO_WEBHOOK_NAME  ?= recording
+MINIO_WEBHOOK_NAME  ?= summary
```
lebaudantoine: I guess the webhook is useful for both recording (video recording) and summary

PR #734, opened 2025-10-03
Add external application api endpoints

lebaudantoine:
(Please refer to my commits) 

Introduce integration with external applications that are not yet connected to our
authorization server, requiring an alternative approach for authentication and
authorization.

These integrations are limited in scope and should be considered a last resort.
The principle of least privilege is applied whenever possible, restricting actions
based on token scopes and, optionally, domains.

In the future, we plan to support full resource server integration to enable more
secure connections for smaller external applications.


src/backend/core/external_api/viewsets.py
```diff
@@ -0,0 +1,182 @@
...
+    acting on behalf of users. All operations are scope-based and filtered to the
+    authenticated user's accessible rooms.
+
+    Supported operations:
+    - list: List rooms the user has access to (requires 'rooms:list' scope)
+    - retrieve: Get room details (requires 'rooms:retrieve' scope)
+    - create: Create a new room owned by the user (requires 'rooms:create' scope)
+    """
+
+    authentication_classes = [authentication.ApplicationJWTAuthentication]
```
lebaudantoine, line 152: I plan to support Resource Server as an alternative authentication on this ViewSet.

lebaudantoine:
@coderabbitai full review

src/backend/core/external_api/permissions.py
```diff
@@ -0,0 +1,61 @@
...
+        """
+        # Get the current action (e.g., 'list', 'create')
+        action = getattr(view, "action", None)
+        if not action:
+            return True
+
+        required_scope = self.scope_map.get(action)
+        if not required_scope:
+            # Action not in scope_map, allow by default
+            return True
```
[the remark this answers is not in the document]
lebaudantoine: legit

src/backend/core/external_api/serializers.py
```diff
@@ -0,0 +1,70 @@
...
+
+
+class ApplicationJwtSerializer(BaseValidationOnlySerializer):
+    """Validate OAuth2 JWT token request data."""
+
+    client_id = serializers.CharField(write_only=True)
+    client_secret = serializers.CharField(write_only=True)
+    grant_type = serializers.ChoiceField(choices=["client_credentials"])
+    scope = serializers.EmailField()
+
```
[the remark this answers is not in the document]
lebaudantoine: legit, I longly hesitated whether validating the email string inside my ViewSet instead.

src/backend/core/external_api/serializers.py
```diff
@@ -0,0 +1,70 @@
...
+        output["url"] = f"{settings.APPLICATIONS_BASE_URL}/{instance.slug}"
+
+        # Add telephony information if enabled
+        if settings.ROOM_TELEPHONY_ENABLED:
+            output["telephony"] = {
+                "enabled": True,
+                "phone_number": settings.ROOM_TELEPHONY_PHONE_NUMBER,
+                "pin_code": instance.pin_code,
+                "default_country": settings.ROOM_TELEPHONY_DEFAULT_COUNTRY,
+            }
```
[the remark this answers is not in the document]
lebaudantoine: legit

src/backend/core/fields.py
```diff
@@ -0,0 +1,44 @@
...
+            )
+        except ValueError:
+            logger.debug(
+                "%s: %s is not hashed; hashing it now.", model_instance, self.attname
+            )
+            hashed_secret = make_password(secret, hasher=settings.SECRET_FIELD_HASHER)
+            setattr(model_instance, self.attname, hashed_secret)
+            return hashed_secret
+
+        return super().pre_save(model_instance, add)
```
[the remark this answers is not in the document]
lebaudantoine: ignored

src/backend/core/external_api/permissions.py
```diff
@@ -0,0 +1,61 @@
...
+        token_scopes = getattr(request.user, "token_scopes", [])
+
+        # Ensure scopes is a list (handle both list and space-separated string)
+        if isinstance(token_scopes, str):
+            token_scopes = token_scopes.split()
+
+        if required_scope not in token_scopes:
+            raise exceptions.PermissionDenied("Insufficient permissions.")
+
+        return True
```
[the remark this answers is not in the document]
lebaudantoine: I prefer keeping Minimal information disclosure and follow conservative security principle. It won't generate much support as few developers will integrate this API. If support increases, I will reconsider my position.

env.d/development/common.dist
```diff
@@ -65,3 +65,8 @@ ROOM_TELEPHONY_ENABLED=True
 
 FRONTEND_USE_FRENCH_GOV_FOOTER=False
 FRONTEND_USE_PROCONNECT_BUTTON=False
+
+# External Applications
+APPLICATIONS_JWT_AUDIENCE=http://localhost:8071/external-api/v1.0/
+APPLICATIONS_JWT_SECRET_KEY=devKey
+APPLICATIONS_BASE_URL=http://localhost:3000
```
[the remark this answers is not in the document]
lebaudantoine: dev environment, not relevant

src/backend/core/models.py
```diff
@@ -717,3 +719,103 @@ def get_abilities(self, user):
...
+    def __str__(self):
+        return f"{self.name!s}"
+
+    def can_delegate_email(self, email):
+        """Check if this application can delegate the given email."""
+
+        if not self.allowed_domains.exists():
+            return True  # No domain restrictions
+
+        domain = email.split("@")[-1]
```
qbey: Potential security issue here, you should use https://github.com/suitenumerique/django-lasuite/blob/c771bae725f30bea75f2965d795a368627afe19b/src/lasuite/tools/email.py#L7 I think

src/helm/meet/templates/ingress.yaml
```diff
@@ -74,6 +74,20 @@ spec:
               serviceName: {{ include "meet.backend.fullname" . }}
               servicePort: {{ .Values.backend.service.port }}
             {{- end }}
+          - path: /external-api/
```
qbey, line 77: Don't you need to also add it few lines bellow?

src/backend/meet/settings.py
```diff
@@ -375,6 +375,13 @@ class Base(Configuration):
     CELERY_BROKER_URL = values.Value("redis://redis:6379/0")
     CELERY_BROKER_TRANSPORT_OPTIONS = values.DictValue({})
 
+    # Secret Field
+    SECRET_FIELD_HASHER = values.Value(
+        "default",
+        environ_name="SECRET_FIELD_HASHER",
+        environ_prefix=None,
+    )
```
lunika: This can not rely on the django internal hashers list system ? Did you see this documentation ? https://docs.djangoproject.com/en/5.2/topics/auth/passwords/

src/helm/meet/Chart.yaml
```diff
@@ -1,4 +1,4 @@
 apiVersion: v2
 type: application
 name: meet
-version: 0.0.12
+version: 0.0.13
```
lunika: While you didn't make a new release you should probably use pre version annotation. In docs and drive I use `0.0.13-beta.1` for example.

src/backend/core/external_api/serializers.py
```diff
@@ -0,0 +1,16 @@
...
+
+from core.api.serializers import BaseValidationOnlySerializer
+
+
+class ApplicationJwtSerializer(BaseValidationOnlySerializer):
+    """Validate OAuth2 JWT token request data."""
+
+    client_id = serializers.CharField(write_only=True)
+    client_secret = serializers.CharField(write_only=True)
+    grant_type = serializers.ChoiceField(choices=["client_credentials"])
```
lunika: There is only one value but using a constant should be better I think.

src/backend/core/external_api/viewsets.py
```diff
@@ -0,0 +1,121 @@
...
+from rest_framework import decorators, viewsets
+from rest_framework import (
+    exceptions as drf_exceptions,
+)
+from rest_framework import (
+    response as drf_response,
+)
+from rest_framework import (
+    status as drf_status,
+)
```
lunika: You can't group them in one ?

src/backend/core/external_api/viewsets.py
```diff
@@ -0,0 +1,121 @@
...
+class ApplicationViewSet(viewsets.GenericViewSet):
+    """API endpoints for application authentication and token generation."""
+
+    @decorators.action(
+        detail=False,
+        methods=["post"],
+        url_path="token",
+        url_name="token",
+    )
+    def generate_token(self, request, *args, **kwargs):
```
lunika: `generate_jwt_access_token` ?

src/backend/core/external_api/viewsets.py
```diff
@@ -0,0 +1,121 @@
...
+            )
+
+        try:
+            user = models.User.objects.get(email=email)
+        except models.User.DoesNotExist as e:
+            raise drf_exceptions.NotFound(
+                {
+                    "error": "User not found.",
+                }
+            ) from e
```
lunika, line 101: What about a settings, like in the mozilla-oidc library, allowing to create a user if not existing ?

src/backend/meet/settings.py
```diff
@@ -683,6 +683,34 @@ class Base(Configuration):
...
+    APPLICATIONS_JWT_EXPIRATION_SECONDS = values.PositiveIntegerValue(
+        3600,
+        environ_name="APPLICATIONS_JWT_EXPIRATION_SECONDS",
+        environ_prefix=None,
+    )
+    APPLICATIONS_JWT_TOKEN_TYPE = values.Value(
+        "Bearer",
+        environ_name="APPLICATIONS_JWT_TOKEN_TYPE",
+        environ_prefix=None,
+    )
```
lunika: Why using the plural form for this settings ?

src/backend/core/external_api/authentication.py
```diff
@@ -0,0 +1,115 @@
...
+
+        if not user.is_active:
+            logger.warning("Inactive user attempted authentication: %s", user_id)
+            raise exceptions.AuthenticationFailed("User account is disabled.")
+
+        user.token_client_id = client_id
+        user.token_scopes = payload.get(
+            "scope", ""
+        ).split()  # Convert space-separated string to list
+        user.is_delegated = is_delegated
```
lunika: You are adding property not present on the `user` object ? If yes I think it is not a good practice. As you are returning the user and the token, both will be present in the request after, so you can use the token to have this information.

src/backend/core/external_api/permissions.py
```diff
@@ -0,0 +1,61 @@
...
+class HasRequiredScope(permissions.BasePermission):
+    """Enforce scope-based access control for application-delegated requests.
+
+    Maps ViewSet actions to required OAuth2-style scopes. JWT tokens issued
+    to applications must contain the appropriate scope to perform operations,
+    implementing the principle of least privilege.
+    """
+
+    scope_map = {
+        "list": models.ApplicationScope.ROOMS_LIST,
```
lunika: You are coupling this permission class to the room object, you should probably not use a generic name for this class. `HasRequiredRoomScope` ? Also you should probably create a mixin holding the `has_permission` method and every permission class will use it and just declare the `scope_map` property. WDYT ?

src/backend/core/external_api/viewsets.py
```diff
@@ -119,3 +119,64 @@ def generate_token(self, request, *args, **kwargs):
...
+    authenticated user's accessible rooms.
+
+    Supported operations:
+    - list: List rooms the user has access to (requires 'rooms:list' scope)
+    - retrieve: Get room details (requires 'rooms:retrieve' scope)
+    - create: Create a new room owned by the user (requires 'rooms:create' scope)
+    """
+
+    authentication_classes = [authentication.ApplicationJWTAuthentication]
+    permission_classes = [api.permissions.IsAuthenticated, permissions.HasRequiredScope]
```
lunika: I suggest to use a bitwise operator here. It will be explicit on the intended purpose. Is it `api.permissions.IsAuthenticated & permissions.HasRequiredScope` or `api.permissions.IsAuthenticated | permissions.HasRequiredScope` ?

src/backend/core/external_api/serializers.py
```diff
@@ -2,8 +2,11 @@
 
 # pylint: disable=abstract-method
 
+from django.conf import settings
```
lunika, line 5: Comment about your commit body. Will would you want to limite the list of returned rooms ? The application is a user delegation, don't you think we can trust it and then return eveything belonging to the user ?

lunika (commented):
Really love what you made. I will implement it later in both docs and drive. Great job.

src/helm/meet/Chart.yaml
lebaudantoine: legit!

src/backend/core/external_api/viewsets.py
lebaudantoine: In the next PR, I will enable external services to create pending users, who can then complete their registration by connecting to the webapp.

src/backend/core/external_api/authentication.py
lebaudantoine: legit yes

src/backend/core/external_api/permissions.py
lebaudantoine: legit, I should follow open-closed principle!

src/backend/core/external_api/serializers.py
lebaudantoine: Yes, you might be right. I think that should at least be configurable per application, to provide a minimum level of trust if needed. With our current providers, though, it feels like YAGNI.

src/helm/meet/templates/ingress.yaml
lebaudantoine: true

src/backend/core/models.py
lebaudantoine: great, I'll use it.

src/backend/core/external_api/viewsets.py
lebaudantoine: Actually, linter lint and reformat imports using this format; I'll ignore this comment.

src/backend/meet/settings.py
lebaudantoine: I’m not sure I fully understand your question. From my understanding, it does. I allow the hashing algorithm for the secret field to be overridden if needed.

src/backend/core/external_api/permissions.py
```diff
@@ -0,0 +1,70 @@
...
+
+        Returns:
+            bool: True if permission granted
+
+        Raises:
+            PermissionDenied: If required scope is missing from token
+        """
+        # Get the current action (e.g., 'list', 'create')
+        action = getattr(view, "action", None)
+        if not action:
```
qbey, line 36: Could you just add a comment to say why we need to manage this case ? (automatic scheme documentation if I remember correctly?)

src/backend/core/external_api/authentication.py
qbey: Yes, could be fixed after merge if you want to deploy fast.

src/backend/core/tests/test_external_api_rooms.py
```diff
@@ -0,0 +1,390 @@
...
+from core.factories import (
+    RoomFactory,
+    UserFactory,
+)
+from core.models import ApplicationScope, RoleChoices, Room
+
+pytestmark = pytest.mark.django_db
+
+
+def generate_test_token(user, scopes):
```
qbey, line 24: Nit: This could have been factorized to be the same function used here and in `generate_jwt_access_token`, but can remain like this

src/backend/core/tests/test_external_api_rooms.py
```diff
@@ -0,0 +1,390 @@
...
+        "client_id": "test-client",
+        "scope": "rooms:list",
+        "user_id": str(user.id),
+        "delegated": True,
+    }
+    token = jwt.encode(
+        payload,
+        settings.APPLICATION_JWT_SECRET_KEY,
+        algorithm=settings.APPLICATION_JWT_ALG,
+    )
```
qbey: Nit: Same can be done by simply setting `settings.APPLICATION_JWT_EXPIRATION_SECONDS = 0`? or maybe use [freezegun](https://github.com/spulec/freezegun)

My point is just to try to remove too much logic in tests. But can remain like this

src/backend/core/tests/test_external_api_rooms.py
```diff
@@ -0,0 +1,390 @@
...
+
+    token = generate_test_token(user, [ApplicationScope.ROOMS_RETRIEVE])
+
+    client = APIClient()
+    client.credentials(HTTP_AUTHORIZATION=f"Bearer {token}")
+    response = client.get(f"/external-api/v1.0/rooms/{room.id}/")
+
+    assert response.status_code == 200
+    assert response.data["id"] == str(room.id)
+    assert response.data["slug"] == room.slug
```
qbey: I would have tested the whole content, to be sure only the wanted data are present (and not super secret stuff).

src/backend/core/tests/test_external_api_rooms.py
```diff
@@ -0,0 +1,390 @@
...
+    token = generate_test_token(user, [ApplicationScope.ROOMS_CREATE])
+
+    client = APIClient()
+    client.credentials(HTTP_AUTHORIZATION=f"Bearer {token}")
+    response = client.post("/external-api/v1.0/rooms/", {}, format="json")
+
+    assert response.status_code == 201
+
+    room = Room.objects.get(id=response.data["id"])
+    assert room.access_level == "trusted"
```
qbey: I'm not sure this requires a dedicated test, it could be tested in `test_api_rooms_create_success` to check the room is in the expected state after creation.

src/backend/core/external_api/serializers.py
```diff
@@ -0,0 +1,72 @@
...
+                "default_country": settings.ROOM_TELEPHONY_DEFAULT_COUNTRY,
+            }
+
+        return output
+
+    def create(self, validated_data):
+        """Create room with secure defaults for application delegation."""
+
+        # Set secure defaults
+        validated_data["name"] = utils.generate_room_slug()
```
qbey, line 70: What if the slug is already existing?

src/backend/meet/settings.py
lebaudantoine: I've removed this extra settings

src/backend/core/external_api/authentication.py
lebaudantoine: nice, will probably do your option @qbey

src/backend/core/tests/test_external_api_rooms.py
lebaudantoine: In my opinion, it was more explicit, but I’ll follow your advice.

src/backend/core/tests/test_external_api_rooms.py
lebaudantoine: legit!

src/backend/core/tests/test_external_api_rooms.py
lebaudantoine: I'll ignore this comment

src/backend/core/external_api/serializers.py
lebaudantoine: I have set this case aside for now. Why? With the current slug pattern, there are approximately 10^14 possible combinations.

To put this in perspective: you would need to generate roughly 1.7 million slugs to have just a 1% chance of encountering at least one collision. For a 50% chance, the number rises to around 14 million slugs.

It's relevant, however I think it can wait for later.

src/backend/core/external_api/permissions.py
lebaudantoine: I have updated this check to return false, if I cannot determine the action being made.

lebaudantoine:
@coderabbitai full review

PR #737, opened 2025-10-08
✨ Add UI/UX improvements and accessibility features

sivamurthy30:
- Enhanced loading screen with animated spinner
- Improved error screen with retry functionality
- Added accessibility helper with skip links and announcements
- Added performance monitor for development
- Added theme toggle with light/dark/auto modes
- Updated CSS with theme variables
- Updated documentation with new features

## Purpose

Description...

## Proposal

Description...

- [] item 1...
- [] item 2...

### DIFF
--- a/src/frontend/src/App.tsx
+++ b/src/frontend/src/App.tsx
@@ -9,6 +9,9 @@ import { Switch, Route } from 'wouter'
 import { I18nProvider } from 'react-aria-components'
 import { Layout } from './layout/Layout'
 import { NotFoundScreen } from './components/NotFoundScreen'
+import { AccessibilityHelper } from './components/AccessibilityHelper'
+import { PerformanceMonitor } from './components/PerformanceMonitor'
+import { ThemeToggle } from './components/ThemeToggle'
 import { routes } from './routes'
 import './i18n/init'
 import { queryClient } from '@/api/queryClient'
@@ -23,16 +26,21 @@ function App() {
 
   return (
     <QueryClientProvider client={queryClient}>
+      <AccessibilityHelper />
+      <PerformanceMonitor enabled={process.env.NODE_ENV === 'development'} />
+      <ThemeToggle />
       {!isSDKContext && <AppInitialization />}
       <Suspense fallback={null}>
         <I18nProvider locale={i18n.language}>
           <Layout>
-            <Switch>
-              {Object.entries(routes).map(([, route], i) => (
-                <Route key={i} path={route.path} component={route.Component} />
-              ))}
-              <Route component={NotFoundScreen} />
-            </Switch>
+            <main id="main-content">
+              <Switch>
+                {Object.entries(routes).map(([, route], i) => (
+                  <Route key={i} path={route.path} component={route.Component} />
+                ))}
+                <Route component={NotFoundScreen} />
+              </Switch>
+            </main>
           </Layout>
           <ReactQueryDevtools
             initialIsOpen={false}
--- a/src/frontend/src/components/AccessibilityHelper.tsx
+++ b/src/frontend/src/components/AccessibilityHelper.tsx
@@ -1,0 +1,79 @@
+import { useEffect, useState } from 'react'
+import { useTranslation } from 'react-i18next'
+import { css } from '@/styled-system/css'
+
+const skipLinkStyles = css({
+  position: 'absolute',
+  top: '-40px',
+  left: '6px',
+  background: '#000',
+  color: '#fff',
+  padding: '8px',
+  textDecoration: 'none',
+  borderRadius: '4px',
+  zIndex: 1000,
+  '&:focus': {
+    top: '6px',
+  },
+})
+
+const announcementStyles = css({
+  position: 'absolute',
+  left: '-10000px',
+  width: '1px',
+  height: '1px',
+  overflow: 'hidden',
+})
+
+export const AccessibilityHelper = () => {
+  const { t } = useTranslation()
+  const [announcement, setAnnouncement] = useState('')
+
+  useEffect(() => {
+    // Listen for route changes and announce them
+    const handleRouteChange = () => {
+      const title = document.title
+      setAnnouncement(`${t('accessibility.navigatedTo', 'Navigated to')} ${title}`)
+      
+      // Clear announcement after screen reader has time to read it
+      setTimeout(() => setAnnouncement(''), 1000)
+    }
+
+    // Listen for title changes (which happen on route changes)
+    const observer = new MutationObserver((mutations) => {
+      mutations.forEach((mutation) => {
+        if (mutation.type === 'childList' && mutation.target === document.querySelector('title')) {
+          handleRouteChange()
+        }
+      })
+    })
+
+    const titleElement = document.querySelector('title')
+    if (titleElement) {
+      observer.observe(titleElement, { childList: true })
+    }
+
+    return () => observer.disconnect()
+  }, [t])
+
+  return (
+    <>
+      {/* Skip to main content link */}
+      <a 
+        href="#main-content" 
+        className={skipLinkStyles}
+      >
+        {t('accessibility.skipToMain', 'Skip to main content')}
+      </a>
+      
+      {/* Screen reader announcements */}
+      <div 
+        className={announcementStyles}
+        aria-live="polite"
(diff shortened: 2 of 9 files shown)


src/frontend/src/components/LoadingScreen.tsx
```diff
@@ -2,15 +2,35 @@ import { Screen, type ScreenProps } from '@/layout/Screen'
 import { DelayedRender } from './DelayedRender'
 import { CenteredContent } from '@/layout/CenteredContent'
 import { useTranslation } from 'react-i18next'
-import { Center } from '@/styled-system/jsx'
+import { Center, VStack } from '@/styled-system/jsx'
+import { css } from '@/styled-system/css'
+
+const spinnerStyles = css({
```
lebaudantoine, line 8: You should instead use [the spinner component](https://github.com/suitenumerique/meet/blob/main/src/frontend/src/primitives/Spinner.tsx).

src/frontend/src/components/ErrorScreen.tsx
```diff
@@ -1,27 +1,69 @@
...
+                {body}
+              </Text>
+            )}
+            {showRetry && (
+              <Button className={retryButtonStyles} onPress={handleRetry}>
+                {t('error.retry', 'Try Again')}
+              </Button>
+            )}
+          </VStack>
+        </Center>
```
lebaudantoine, line 66: Could you clarify why you would add this and perhaps provide an example?

src/frontend/src/components/ThemeToggle.tsx
```diff
@@ -0,0 +1,114 @@
...
+  },
+  '&:focus': {
+    outline: '2px solid #fff',
+    outlineOffset: '2px',
+  },
+})
+
+type Theme = 'light' | 'dark' | 'auto'
+
+export const ThemeToggle = () => {
```
lebaudantoine, line 36: Great idea! However, I don’t think we’re ready to implement a theme toggle yet. We’d like to keep the room in dark mode for contrast, while the outside room screens remain in light mode.

src/frontend/src/components/PerformanceMonitor.tsx
```diff
@@ -0,0 +1,97 @@
...
+  },
+})
+
+interface PerformanceMetrics {
+  fps: number
+  memory: number
+  connectionQuality: 'excellent' | 'good' | 'poor' | 'unknown'
+}
+
+export const PerformanceMonitor = ({ enabled = false }: { enabled?: boolean }) => {
```
lebaudantoine, line 27: This seems related to a developer experience component.
Could you explain when and why you would use it to help understand whether to accept your changes, and perhaps isolate it in a dedicated commit?

lebaudantoine (commented):
Awesome, thanks for the contribution!
Maybe you could split your changes into focused commits and add a short note on why each one’s important. It would help the reviewer

src/frontend/src/components/AccessibilityHelper.tsx
```diff
@@ -0,0 +1,79 @@
...
+
+const announcementStyles = css({
+  position: 'absolute',
+  left: '-10000px',
+  width: '1px',
+  height: '1px',
+  overflow: 'hidden',
+})
+
+export const AccessibilityHelper = () => {
```
manuhabitela, line 28: Hey, thanks for this :)

What do you think about splitting this in two different components? I admit it feels a bit weird to have skip links and route announcer in the same component, they are not really related from my pov.

src/frontend/src/components/AccessibilityHelper.tsx
```diff
@@ -0,0 +1,79 @@
...
+  const [announcement, setAnnouncement] = useState('')
+
+  useEffect(() => {
+    // Listen for route changes and announce them
+    const handleRouteChange = () => {
+      const title = document.title
+      setAnnouncement(`${t('accessibility.navigatedTo', 'Navigated to')} ${title}`)
+      
+      // Clear announcement after screen reader has time to read it
+      setTimeout(() => setAnnouncement(''), 1000)
```
manuhabitela, line 39: Is there a specific reason you do that? I'm worried doing that would skip the vocalization if it takes more than one second. And I don't see the need of cleaning this, since it would get vocalized only when changes happen, leaving the text as is should not be a problem?

src/frontend/src/components/AccessibilityHelper.tsx
```diff
@@ -0,0 +1,79 @@
...
+      </a>
+      
+      {/* Screen reader announcements */}
+      <div 
+        className={announcementStyles}
+        aria-live="polite"
+        aria-atomic="true"
+      >
+        {announcement}
+      </div>
```
manuhabitela, line 76: What do you think of having the page change announcement div actually be after the whole app in the dom, rather than before? The idea of this is mostly to notify the user, we don't want him risking focusing this manually by mistake.

src/frontend/src/components/AccessibilityHelper.tsx
```diff
@@ -0,0 +1,79 @@
...
+      })
+    })
+
+    const titleElement = document.querySelector('title')
+    if (titleElement) {
+      observer.observe(titleElement, { childList: true })
+    }
+
+    return () => observer.disconnect()
+  }, [t])
```
manuhabitela, line 57: I'll admit I don't quite understand this whole part, having `t` as the effect dependency, and using a mutation observer to detect changes.

Can you explain a little more? 🙏 

Would a logic more tied to the app routes be simpler maybe? Like listening to actual route changes and then updating the announcement state with the current title content.

src/frontend/src/components/AccessibilityHelper.tsx
```diff
@@ -0,0 +1,79 @@
...
+
+  return (
+    <>
+      {/* Skip to main content link */}
+      <a 
+        href="#main-content" 
+        className={skipLinkStyles}
+      >
+        {t('accessibility.skipToMain', 'Skip to main content')}
+      </a>
```
manuhabitela, line 67: Question I have when reading this: do we really want skip links on the meeting page? Where there is basically nothing to skip before main content. I guess thinking about meeting-page specific skip links, to skip directly to the toolbar, or right pane, might be interesting?

So creating a SkipLinks component where we can pass a specific set of links to generate might be interesting?

src/frontend/src/App.tsx
```diff
@@ -23,16 +26,21 @@ function App() {
...
       <Suspense fallback={null}>
         <I18nProvider locale={i18n.language}>
           <Layout>
-            <Switch>
-              {Object.entries(routes).map(([, route], i) => (
-                <Route key={i} path={route.path} component={route.Component} />
-              ))}
-              <Route component={NotFoundScreen} />
-            </Switch>
+            <main id="main-content">
```
manuhabitela, line 36: This is certainly a mistake?

From my understanding this would make the whole app wrapped in a main tag, always. The header component exposes header/nav elements, and we already use a main tag in the content components. And, adding this would make the skip links basically useless as they would skip to the whole app that is rendered right after the skip links.

src/frontend/src/components/ThemeToggle.tsx
```diff
@@ -0,0 +1,114 @@
...
+      default: return t('theme.auto', 'Auto theme')
+    }
+  }
+
+  return (
+    <Button
+      className={themeToggleStyles}
+      onPress={cycleTheme}
+      aria-label={`${t('theme.toggle', 'Toggle theme')} - ${getThemeLabel()}`}
+      title={getThemeLabel()}
```
manuhabitela, line 109: Not sure we'll merge the theme toggle feature but if we do:

- I'd avoid using a `title` attribute as its support is not ideal in assistive technologies
- I'd make the `aria-label` more explicit. Like if it says `Toggle theme - Light theme`, does it mean we currently are on the light theme, or that if we click the button, we'll end up with the light theme? For example, having `Switch to light theme` as `aria-label` and `Current theme: dark` as `aria-description` might be better: we describe the button action first, and remind the user about current state second.

lebaudantoine:
This PR has gone stale, so I'll close it for now @sivamurthy30. Please feel free to reopen it and split it into smaller parts to make review and merging easier.

sivamurthy30:
yeah ok ,i will recheck it and i will make a pr

PR #742, opened 2025-10-10
🔧(backend) add configurable language settings for WhisperX transcription

lebaudantoine:
Make WhisperX language detection configurable through FastAPI settings to handle empty audio start scenarios where automatic detection fails and incorrectly defaults to English despite 99% French usage.

Quick fix acknowledging long-term solution should allow dynamic per-recording language selection configured by users through web interface rather than global server settings.


lebaudantoine:
@coderabbitai full review

PR #752, opened 2025-10-20
✨(helm) Add extraManifests support for custom Kubernetes resources

gigi206:
## Summary

This PR adds the ability to inject custom Kubernetes manifests through the Helm chart values, allowing users to deploy additional resources without modifying chart templates.

## Changes

- Add `templates/extra-objects.yaml` template with flexible rendering engine
- Add `extraManifests` parameter in `values.yaml` with comprehensive documentation
- Support multiple input formats: list of objects, map of named objects, and raw YAML strings
- Enable Helm template variables (e.g., `{{ .Release.Namespace }}`) in injected manifests
- Bump chart version to `0.0.15`

## Use Cases

Users can now deploy additional Kubernetes resources alongside the main application without forking or modifying the chart:

- Custom ConfigMaps for feature flags or configuration
- Additional Services (external services)
- NetworkPolicies for security rules
- PersistentVolumeClaims for debugging storage
- CronJobs for cleanup tasks
- ServiceAccounts and RBAC rules

### List format (recommended)

```yaml
extraManifests:
  - apiVersion: v1
    kind: ConfigMap
    metadata:
      name: {{ .Release.Name }}-custom-config
      namespace: {{ .Release.Namespace }}
    data:
      config.json: |
        {
          "feature": "enabled"
        }

  - apiVersion: v1
    kind: Service
    metadata:
      name: minio
      namespace: {{ .Release.Namespace }}
    spec:
      type: ExternalName
      externalName: minio.storage.svc.cluster.local
```

### Map format (named objects)

```yaml
extraManifests:
  customConfig:
    apiVersion: v1
    kind: ConfigMap
    metadata:
      name: custom-config
    data:
      key: value
```

### String format (raw YAML)

```yaml
extraManifests:
  - |
    apiVersion: v1
    kind: Secret
    metadata:
      name: {{ .Release.Name }}-credentials
    stringData:
      password: {{ .Values.password }}
```

### DIFF
--- a/src/helm/meet/Chart.yaml
+++ b/src/helm/meet/Chart.yaml
@@ -1,4 +1,4 @@
 apiVersion: v2
 type: application
 name: meet
-version: 0.0.14
+version: 0.0.15


src/helm/meet/values.yaml
```diff
@@ -822,3 +822,22 @@ agents:
...
+## @extra extraManifests[].apiVersion API version of the resource
+## @extra extraManifests[].kind Kind of the resource (Deployment, Service, ConfigMap, etc.)
+## @extra extraManifests[].metadata Resource metadata
+## @extra extraManifests[].spec Resource specification
+extraManifests: []
+#  - apiVersion: v1
+#    kind: ConfigMap
+#    metadata:
+#      name: my-custom-config
+#      namespace: {{ .Release.Namespace }}
```
rouja: The value of .Release.Namespace can't be in the values.yaml as it will probably lead to this error:

```
Error: cannot load values.yaml: error converting YAML to JSON: yaml: invalid map key: map[interface {}]interface {}{".Release.Namespace":interface {}(nil)}
```

I think just removing this from the example should be ok.

gigi206:
Thanks for catching this! You're absolutely right - Helm template syntax like `{{ .Release.Namespace }}` cannot be used in `values.yaml` as it causes YAML parsing errors.

I've removed the `namespace` field from the example in commit 3a9a1ba5. This is actually better since Kubernetes will automatically use the release namespace when the field is not explicitly specified in the manifest.

Verified with:

- ✅ `helm lint` - passes without errors
- ✅ `helm template` - renders correctly
- ✅ `extraManifests` example - works as expected

The chart is now ready to use without any YAML parsing issues.

gigi206:
Hi,
If it's okay with you, could we please merge this pull request?

PR #757, opened 2025-10-22
🐛(celery) fix metadata manager task_args order broken by signal sender argument

lebaudantoine:
Restore correct task_args ordering in metadata manager after commit f0939b6f added sender argument to Celery signals for transcription task scoping, unexpectedly shifting positional arguments and breaking metadata creation.

Issue went undetected due to missing staging analytics deployment, silently losing production observability on microservice without blocking transcription job execution, highlighting need for staging analytics activation.


lebaudantoine:
@coderabbitai full review

Issue #753, opened 2025-10-21
False alert: "Your camera is in use in another tab"


lebaudantoine:
## Bug Report

**Problematic behavior**
User could not activate camera.
Only happens on Firefox

**Expected behavior/code**

<img width="457" height="227" alt="Image" src="https://github.com/user-attachments/assets/27aa108a-6ad2-4d74-a49a-53499c1bdfa0" />

<img width="718" height="257" alt="Image" src="https://github.com/user-attachments/assets/ad6d9b4b-4750-47b4-8878-f3da395baa5f" />

Vincentparisgithub:
Precision : For the moment, it happens only on the PC of the ofb.gouv.fr
https://app.crisp.chat/website/58ea6697-8eba-4492-bc59-ad6562585041/inbox/session_da6b113d-47f4-42d1-9ac8-7752a1dd6a8c/
And see a second cas treated directly by email  (search the email with the object "Bug/firefox" 22/10)

Please note that Teams are installed on their PC.
Maybe we shoud contact their DSI

Their DSI contact is in the same crips file : https://app.crisp.chat/website/58ea6697-8eba-4492-bc59-ad6562585041/inbox/session_da6b113d-47f4-42d1-9ac8-7752a1dd6a8c/

PR #762, opened 2025-10-22
FIX: issue #753

sarthak503:
## Purpose
Reg issue : "Your camera is in use in another tab" 

Description...
Firefox can show a false "Your camera is in use in another tab" alert and block camera activation.

## Proposal
Add an opt‑in Firefox suppression flag.

Description...
- Neutralize locale copy for camera DeviceInUse.
- Add opt-in suppression: when [apiConfig.livekit.suppress_firefox_deviceinuse_alert] is true and client is Firefox, do not show the DeviceInUse dialog.

### DIFF
--- a/src/frontend/src/features/rooms/components/Conference.tsx
+++ b/src/frontend/src/features/rooms/components/Conference.tsx
@@ -232,7 +232,17 @@ export const Conference = ({
             }
           }}
           onMediaDeviceFailure={(e, kind) => {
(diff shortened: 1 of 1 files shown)


lebaudantoine:
This does not resolve the issue. It neither explains nor addresses the false positive of 'device used in another tab'; it merely disables the feature via a backend setting, which is entirely unacceptable. Additionally, the frontend build is broken. I will close the issue for now — feel free to reopen it whenever you believe it’s ready for reevaluation.

PR #776, opened 2025-11-14
Wip enhance accessibility

lebaudantoine:
Wip research works.

### DIFF
--- a/src/frontend/src/primitives/ToggleButton.tsx
+++ b/src/frontend/src/primitives/ToggleButton.tsx
@@ -1,10 +1,10 @@
+import { forwardRef, ReactNode } from 'react'
 import {
   ToggleButton as RACToggleButton,
   ToggleButtonProps as RACToggleButtonProps,
 } from 'react-aria-components'
 import { type ButtonRecipeProps, buttonRecipe } from './buttonRecipe'
 import { TooltipWrapper, TooltipWrapperProps } from './TooltipWrapper'
-import { ReactNode } from 'react'
 
 export type ToggleButtonProps = RACToggleButtonProps &
   ButtonRecipeProps &
@@ -16,24 +16,25 @@ export type ToggleButtonProps = RACToggleButtonProps &
 /**
  * React aria ToggleButton with our button styles, that can take a tooltip if needed
  */
-export const ToggleButton = ({
-  tooltip,
-  tooltipType,
-  ...props
-}: ToggleButtonProps) => {
-  const [variantProps, componentProps] = buttonRecipe.splitVariantProps(props)
+export const ToggleButton = forwardRef<HTMLButtonElement, ToggleButtonProps>(
+  ({ tooltip, tooltipType, ...props }, ref) => {
+    const [variantProps, componentProps] = buttonRecipe.splitVariantProps(props)
 
-  return (
-    <TooltipWrapper tooltip={tooltip} tooltipType={tooltipType}>
-      <RACToggleButton
-        {...componentProps}
-        className={[buttonRecipe(variantProps), props.className].join(' ')}
-      >
-        <>
-          {componentProps.children as ReactNode}
-          {props.description && <span>{tooltip}</span>}
-        </>
-      </RACToggleButton>
-    </TooltipWrapper>
-  )
-}
+    return (
+      <TooltipWrapper tooltip={tooltip} tooltipType={tooltipType}>
+        <RACToggleButton
+          ref={ref}
+          {...componentProps}
+          className={[buttonRecipe(variantProps), props.className].join(' ')}
+        >
+          <>
+            {componentProps.children as ReactNode}
+            {props.description && <span>{tooltip}</span>}
+          </>
+        </RACToggleButton>
+      </TooltipWrapper>
+    )
+  }
+)
+
+ToggleButton.displayName = 'ToggleButton'


src/frontend/src/primitives/TooltipWrapper.tsx
```diff
@@ -27,7 +27,7 @@ export const TooltipWrapper = ({
   return tooltip ? (
     <TooltipTrigger delay={tooltipType === 'instant' ? 150 : 1000}>
       {children}
-      <Tooltip>{tooltip}</Tooltip>
+      <Tooltip aria-hidden="true">{tooltip}</Tooltip>
```
lebaudantoine: I have the feeling that the tooltip is still being announced by the screen reader. As a result, the label is read out twice (see screenshot below).

Hiding the tooltip might have a broader impact, and I’m not sure that disabling it would be a neutral change from an accessibility standpoint.

Capture d’écran 2025-12-15 à 17 59 36

Ovgodd: hey @lebaudantoine 

You’re right: even with aria-hidden="true", the tooltip gets announced because the React Aria TooltipTrigger adds an aria-describedby to the button, which points to the tooltip. That’s why the label is read twice. Disabling or hiding the tooltip isn’t neutral for accessibility—we need a different approach (e.g., a visual-only tooltip not referenced by aria-describedby) to avoid duplicating the announcement while keeping the visual hover/focus help.

src/frontend/src/features/rooms/livekit/components/effects/EffectsConfiguration.tsx
```diff
@@ -353,12 +390,15 @@ export const EffectsConfiguration = ({
                       <ToggleButton
                         key={i}
                         variant="bigSquare"
-                        aria-label={tooltipLabel(ProcessorType.VIRTUAL, {
-                          imagePath,
-                        })}
-                        tooltip={tooltipLabel(ProcessorType.VIRTUAL, {
-                          imagePath,
-                        })}
```
lebaudantoine, line 361: no tooltip is displayed anymore for regular user, I guess something isn't working as expected.
Capture d’écran 2026-01-06 à 00 04 41

Another comment I had while looking at this screenshot: do you think the visual contrast of the actual selected ring is strong enough to be clearly noticed and understood?

Ovgodd: Hey @lebaudantoine, I re-add the tooltip ! nice catch.

 
Regarding the visual contrast of the current selection ring, this is OK.
The contrast ratio is 3.14:1, which passes the “Graphical Objects and User Interface Components” criterion, requiring a minimum of 3:1.

src/frontend/src/features/rooms/livekit/components/effects/EffectsConfiguration.tsx
```diff
@@ -69,16 +70,46 @@ export const EffectsConfiguration = ({
...
+      cancelAnimationFrame(rafId)
+    }
+  }, [])
+
   const clearEffect = async () => {
     await videoTrack.stopProcessor()
     onSubmit?.(undefined)
   }
 
+  const updateBlurStatusMessage = (
```
lebaudantoine, line 121: Unselecting a bacckground the label feels like something is missing, should the name of the background be right after the `: ` ?

Image

What do you think so ?

lebaudantoine: Same, trainling `:` Arrière plan appliqué should be removed or completed with the name of the background?

Image

lebaudantoine, line 121: Same, trailing `:` Arrière plan appliqué should be removed or completed with the name of the background?

Image

src/frontend/src/features/rooms/livekit/components/effects/EffectsConfiguration.tsx
```diff
@@ -153,8 +186,16 @@ export const EffectsConfiguration = ({
     )
   }
 
-  const tooltipLabel = (type: ProcessorType, options: BackgroundOptions) => {
-    return t(`${type}.${isSelected(type, options) ? 'clear' : 'apply'}`)
+  const tooltipBlur = (type: ProcessorType, options: BackgroundOptions) => {
```
lebaudantoine, line 226: I guess same issue, the bluring option is vocalized twice on voice over

Image

src/frontend/src/features/rooms/livekit/components/effects/EffectsConfiguration.tsx
```diff
@@ -131,6 +162,8 @@ export const EffectsConfiguration = ({
         // We want to trigger onSubmit when options changes so the parent component is aware of it.
         onSubmit?.(processor)
       }
+
+      updateBlurStatusMessage(type, options, wasSelectedBeforeToggle)
```
lebaudantoine, line 203: When selecting and unselecting the blur options quickly, the status update is announced too late.
For example, if I select Blur and then unselect it quickly, the screen reader finishes announcing all the toggle labels, and only afterward I hear “Background heavy blur” and “Background light blur.”
This feedback does not reflect the final state. I should receive audio feedback only for the last selected background state. Wdyt?

https://github.com/user-attachments/assets/e06ea746-e1fe-4da3-a2db-6a02efd01de1

src/frontend/src/features/rooms/livekit/components/effects/EffectsConfiguration.tsx
Ovgodd: Yes good catch, this is a good things to have

Ovgodd: Of course !

src/frontend/src/features/rooms/livekit/components/effects/EffectsConfiguration.tsx
Ovgodd: Hey @lebaudantoine, 

I’m not able to reproduce this on my side. When I quickly select and unselect a blur option, the screen reader switches directly to the correct, final feedback. I don’t hear delayed announcements for intermediate states (like “Background heavy blur” or “Background light blur”) once the final state is reached.

It might be related to the screen reader ? we can check together 

https://github.com/user-attachments/assets/29732190-c44d-4cdb-9e3c-7d17dd5d97ad

PR #782, opened 2025-11-24
✨(backend) allow prefixing resource server scopes

lebaudantoine:
When declaring scopes with our OIDC provider, they require us to prefix each scope with our application name. This is to prevent reserving generic scopes like rooms:list for only our app, as they manage a large federation.

I’m proposing a workaround where, if a resource server prefix is detected in the scope, it’s stripped out. This solution is simple and sufficient in my opinion.

Since the scopes are defined in the database, I don’t want to update them directly. Additionally, each self-hosted instance may have a different application name, so the prefix should be configurable via a Django setting.

### DIFF
--- a/src/backend/core/external_api/permissions.py
+++ b/src/backend/core/external_api/permissions.py
@@ -3,6 +3,8 @@
 import logging
 from typing import Dict
 
+from django.conf import settings
+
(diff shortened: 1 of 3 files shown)


lebaudantoine:
@coderabbitai full review

Issue #806, opened 2025-12-15
No keyboard focus management for chat panel (open / close)[/-]


Ovgodd:
## Bug Report
 **Problematic behavior** 

Chat focus management is not keyboard-friendly.

When opening the chat panel:
- Focus does **not** move to the message input.
- Keyboard users must manually tab through unrelated elements to reach the input.

**When closing the chat panel (Esc, close button, keyboard action):**
- Focus does **not** return to the element that originally opened the chat.
- Focus is lost or jumps to an unrelated area of the UI.

This breaks expected keyboard navigation and creates confusion for screen reader and keyboard-only users.

---

### Expected behavior/code

**On chat opening:**
- Focus should automatically move to the **message input field**, allowing immediate typing.

**On chat closing:**
- Focus should return **exactly to the triggering element** (button, shortcut, etc.) that opened the chat panel.

---

### Steps to Reproduce

1. Join a meeting.
2. Navigate using keyboard only (Tab / Shift+Tab).
3. Open the chat panel via keyboard.
4. Observe that focus does not land on the message input.
5. Close the chat panel using Esc or the close button.
6. Observe that focus does not return to the triggering element.

PR #807, opened 2025-12-15
♿️ (frontend) Improve focus management when opening and closing chat

Ovgodd:
## Purpose

Improve keyboard accessibility and focus management for the chat side panel.

This change ensures that users can start typing immediately when the chat is opened, and that focus behavior remains predictable when the panel is closed.

issue : [806](https://github.com/suitenumerique/meet/issues/806)

https://github.com/user-attachments/assets/36deec4d-83b5-4c78-bc78-3e9f299a93af

---

## Proposal

Ensure consistent, dialog-like focus management for the chat panel:

- [x] Automatically move keyboard focus to the message input when the chat panel opens
- [x] Restore keyboard focus to the element that opened the chat panel when it closes

### DIFF
--- a/src/frontend/src/features/rooms/livekit/prefabs/Chat.tsx
+++ b/src/frontend/src/features/rooms/livekit/prefabs/Chat.tsx
@@ -48,14 +48,17 @@ export function Chat({ ...props }: ChatProps) {
     // Chat just opened
     if (!wasChatOpen && isChatPanelOpen) {
       chatTriggerRef.current = document.activeElement as HTMLElement | null
-      inputRef.current?.focus()
+      // Avoid layout "jump" during the side panel slide-in animation.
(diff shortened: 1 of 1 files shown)


Ovgodd:
> Great, thank you, that would be a game changer! I’ve noticed a small regression in the CSS animation of the side panel.
> 
> Before :
> 
>  Enregistrement.de.l.ecran.2025-12-15.a.17.50.59.mov 
> After :
> 
>  Enregistrement.de.l.ecran.2025-12-15.a.17.49.00.mov

oh good catch, I fixed ( this is caused by the focus, it can trigger the "scroll into view" et "break" an animation


https://github.com/user-attachments/assets/7c9f55e0-237f-49a7-ac1e-da9a8d7d5866

PR #810, opened 2025-12-15
♿️(frontend) Improve SR announcements and focus for transcript and recording

Ovgodd:
## Purpose

Improve accessibility feedback and keyboard usability for transcription and recording actions.

---

## Proposal

Improve screen reader announcements and focus management for transcription and recording:

- [x] Announce transcript and recording events to screen readers
- [x] Move keyboard focus to the transcript or recording button when the related panel opens

### DIFF
--- a/src/frontend/src/features/rooms/livekit/components/SidePanel.tsx
+++ b/src/frontend/src/features/rooms/livekit/components/SidePanel.tsx
@@ -164,7 +164,7 @@ export const SidePanel = () => {
       <Panel isOpen={isChatOpen} keepAlive={true}>
         <Chat />
       </Panel>
-      <Panel isOpen={isToolsOpen}>
+      <Panel isOpen={isToolsOpen} keepAlive={true}>
         <Tools />
       </Panel>
       <Panel isOpen={isAdminOpen}>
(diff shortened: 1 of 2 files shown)


lebaudantoine:
I've quickly read it, I'll dig into it tomorrow. I heavily refactored this part on Friday; we might have conflicts. Let's discuss it.

Ovgodd:
> I've quickly read it, I'll dig into it tomorrow. I heavily refactored this part on Friday; we might have conflicts. Let's discuss it.

Yes of course let's talk about it ;)

Ovgodd:
> LGTM.
> 
> I have two minor comments:
> 
> 1. When opening the recording/transcript side panel, the button does receive focus — I can activate it using Enter or Space — however, there is no visual indication that the element is focused, as there would be when focusing via Tab or other keyboard interactions. I assume this is expected when focus is set programmatically, but I wanted to check if you see the same behavior.
> 2. I’m also seeing the same flickering issue with the side panel animation — similar to the one you addressed in the chat panel.

Yes great catch again, I fixed it, 

for the visual focus, I don't have this issue as discussed with you, this is weird

PR #811, opened 2025-12-15
Allow provision users through the external API

lebaudantoine:
## Purpose

Enable user acquisition and adoption through third-party platforms integrating with our videoconference tool via the external API.

## Description

Adoption of our videoconference tool increasingly depends on external platforms that integrate with us through the public/external API. These platforms may initiate interactions on behalf of users who are not yet known by our system.

Today, this creates friction: without a pre-existing user account, these integrations cannot fully provision access, which limits usage and slows adoption from third-party platforms.

This change allows provisional user creation based solely on an email address when a user does not yet exist in our platform. Full user identification (sub) is completed when the user logs in for the first time, at which point the identity can be reliably reconciled.

Although email-only user creation is not ideal, it provides a practical mechanism to unlock integrations and support growth through external platforms.

## Proposal

- Support provisioning of users via the external API for third-party platforms.
- Create user records using email only when no existing user is found.
- Finalize user identification during the user’s first login to the videoconference tool.

## Risks & Considerations

- This implementation is intentionally naïve and should be treated as an initial approach.
- It depends on implicit behaviors in authentication and user-resolution logic that are not immediately obvious.
- These implicit dependencies may increase complexity for new contributors and introduce technical debt if not clearly documented.

The approach requires team discussion to validate long-term suitability and define boundaries.

### DIFF
--- a/src/backend/core/external_api/viewsets.py
+++ b/src/backend/core/external_api/viewsets.py
@@ -100,7 +100,15 @@ class ApplicationViewSet(viewsets.GenericViewSet):
                 and settings.OIDC_FALLBACK_TO_EMAIL_FOR_IDENTIFICATION
                 and not settings.OIDC_USER_SUB_FIELD_IMMUTABLE
             ):
-                # Create a pending user without sub, but with an email.
+                # Create a provisional user without `sub`, identified by email only.
+                #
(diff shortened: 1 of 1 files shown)


lebaudantoine:
@coderabbitai full review

PR #829, opened 2026-01-04
📌(egress) pin egress version to v1.11.0

lebaudantoine:
Pin egress to the production version, which uses a more recent release than the default chart value (1.9.0).

Using the default could have led to issues; hopefully this change avoids them.


lebaudantoine:
@coderabbitai full review

PR #830, opened 2026-01-05
🐛(backend) fix ignore recording webhook events

lebaudantoine:
Fix an unexpected behavior where filtering LiveKit webhook events sometimes failed because the room name was not reliably extracted from the webhook data, causing notifications to be ignored.

Configure the same filtering logic locally to avoid missing this kind of issue in the future.


lebaudantoine:
@coderabbitai full review

PR #831, opened 2026-01-05
🐛(summary) fix MinIO endpoint handling in constructor

lebaudantoine:
Fix MinIO client configuration: I was incorrectly passing a full URL instead of an endpoint, which caused errors in staging. Local development values did not reflect the staging setup and were also out of sync with the backend.


lebaudantoine:
@coderabbitai full review

PR #839, opened 2026-01-06
📈(frontend) track new recording's modes

lebaudantoine:
I introduced transcript + screen recording modes but forgot to properly track them in PostHog. Fix this issue.


lebaudantoine:
@coderabbitai full review

PR #840, opened 2026-01-06
✨(summary) add dutch and german languages

### LINKED ISSUE #837: Add german(+dutch) to transcription language select box
Right now only french and english can be selected:

Image

When I set it to "automatic" it works OK with German so please add this to the selection box.

Dutch probably also works.

lebaudantoine:
Based on a request from our European partners, introduce new languages for the transcription feature. Dutch and German are now supported, which is a great addition.

It closes #837.

WhisperX is expected to support both languages.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,10 @@ and this project adheres to
 
 ## [Unreleased]
 
+### Added
+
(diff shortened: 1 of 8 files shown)


lebaudantoine:
@coderabbitai full review

Issue #809, opened 2025-12-15
No keyboard focus on transcription / recording controls when opened


Ovgodd:
### Problematic behavior

**Transcription and recording controls do not receive keyboard focus when opened.**

When opening the transcription or recording feature, focus is **not moved to the associated control button**.

---

### Expected behavior/code

- When transcription or recording is opened, keyboard focus should automatically move to the corresponding control button.

---

### Steps to Reproduce

1. Join a meeting.
2. Navigate using keyboard only (Tab / Shift+Tab).
3. Open transcription or start recording via keyboard.
4. Observe that focus does not move to the related control button.
5. Press `Tab` repeatedly to locate the control.

---

PR #843, opened 2026-01-06
Access/issue 808 rebased

lebaudantoine:
Refer to commits

issue [https://github.com/suitenumerique/meet/issues/809](809)

### DIFF
--- a/src/frontend/src/features/rooms/livekit/components/Tools.tsx
+++ b/src/frontend/src/features/rooms/livekit/components/Tools.tsx
@@ -2,8 +2,9 @@ import { A, Div, Text } from '@/primitives'
 import { css } from '@/styled-system/css'
 import { Button as RACButton } from 'react-aria-components'
 import { useTranslation } from 'react-i18next'
-import { ReactNode, useEffect, useRef } from 'react'
+import { ReactNode } from 'react'
 import { SubPanelId, useSidePanel } from '../hooks/useSidePanel'
+import { useRestoreFocus } from '@/hooks/useRestoreFocus'
 import {
   useIsRecordingModeEnabled,
   RecordingMode,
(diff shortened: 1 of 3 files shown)


src/frontend/src/features/rooms/livekit/components/Tools.tsx
```diff
@@ -94,10 +94,49 @@ const ToolButton = ({
...
+    // Tools just opened
+    if (!wasOpen && isOpen) {
+      const activeEl = document.activeElement as HTMLElement | null
+
+      // If the active element is a MenuItem (DIV) that will be unmounted when the menu closes,
+      // find the "Plus d'options" button that opened the menu
+      if (activeEl?.tagName === 'DIV') {
+        toolsTriggerRef.current = document.querySelector<HTMLButtonElement>(
+          'button[aria-label="Plus d\'options"]'
+        )
```
lebaudantoine: This feels fragile and language-dependent (French only). Should we use an ID or a dedicated selector instead? Keep the inline comment with the English version of “plus d’options”.

src/frontend/src/features/rooms/livekit/components/Tools.tsx
```diff
@@ -94,10 +94,49 @@ const ToolButton = ({
...
   const { data } = useConfig()
-  const { openTranscript, openScreenRecording, activeSubPanelId } =
+  const { openTranscript, openScreenRecording, activeSubPanelId, isToolsOpen } =
     useSidePanel()
   const { t } = useTranslation('rooms', { keyPrefix: 'moreTools' })
 
+  // Restore focus to the element that opened the Tools panel
+  // following the same pattern as Chat.
+  const toolsTriggerRef = useRef<HTMLElement | null>(null)
+  const prevIsToolsOpenRef = useRef(false)
```
lebaudantoine: nit: do you think at some point we should encapsulate this logic in a reusable hook to use anywhere in the code?

lebaudantoine:
@coderabbitai full review

src/frontend/src/features/rooms/livekit/components/Tools.tsx
Ovgodd: I think this is a very good idea !

src/frontend/src/features/rooms/livekit/components/Tools.tsx
Ovgodd: Of course, that make sense

PR #845, opened 2026-01-07
🐛(frontend) remove unexpected F2 tooltip when clicking video screen

Ovgodd:
## Purpose

Prevent confusion caused by the "F2 : raccourci barre d'outils en bas." tooltip when interacting with the visio (video call) button via mouse click.

## Proposal

Ensure the tooltip is only shown when navigating via keyboard (Tab), not on click.

- [x] Restrict tooltip visibility to keyboard navigation only  
- [x] Confirm tooltip does not appear on mouse interaction with the visio button

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,10 @@ and this project adheres to
 
 - 📈(frontend) track new recording's modes
 
+### Fixed
+
+- 🐛(frontend) remove unexpected F2 tooltip when clicking video screen
+
 ## [1.2.0] - 2026-01-05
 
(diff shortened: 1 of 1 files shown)


CHANGELOG.md
```diff
@@ -16,6 +16,10 @@ and this project adheres to
 
 - 📈(frontend) track new recording's modes
 
+### Fixed
+
+🐛(frontend) remove unexpected F2 tooltip when clicking video screen
```
lebaudantoine: nit: missing `- `

PR #848, opened 2026-01-07
🩹(frontend) close subtitles on room disconnections

lebaudantoine:
Subtitles were still visible when leaving and rejoining a meeting, even though the backend API call to start them was not triggered again.

Introduce a hook that closes the subtitles layout on unmount, ensuring users must explicitly click the button to restart subtitles when they rejoin a room.


lebaudantoine:
@coderabbitai full review

PR #854, opened 2026-01-08
🔧(agents) make Silero VAD optional

lebaudantoine:
Allow configuring whether a VAD model runs before calling an external ASR API. Running VAD can save API calls (and costs) when no audible sound is detected, but comes with the trade-off of additional computational overhead.


lebaudantoine:
@coderabbitai full review

PR #860, opened 2026-01-13
🚸(frontend) hide back button when a user is ejected by an admin

lebaudantoine:
Avoid showing a back button when a user is kicked out of a meeting by an admin, to prevent them from repeatedly rejoining the room.


lebaudantoine:
@coderabbitai full review

PR #878, opened 2026-01-16
♿️(frontend) align close dialog label in rooms locale

Ovgodd:
## Purpose

Align the close dialog label in rooms with global wording.

## Proposal

Update the FR rooms locale to use the same close dialog label.

- [x] Confirm translation consistency across locales
- [x] Quick check of the share dialog labels


lebaudantoine (commented):
I’m not clear on the exact proposal in this PR.  Let's discuss it IRL

PR #879, opened 2026-01-16
♿️(frontend) improve background effect announcements

Ovgodd:
## Purpose

Fix sr feedback for background effects by ensuring the active or cleared
state is always announced.

## Proposal

Add a unified screen‑reader announcement path for blur and virtual backgrounds,
including the "no effect" case.
 
- [x] validate SR announces "background clear" when effect is removed
- [x] validate SR announces virtual background name when applied


lebaudantoine:
@coderabbitai full review

PR #881, opened 2026-01-16
♿️(frontend) add accessible back button in side panel

Ovgodd:
## Purpose

Fix the missing accessible label on the back button in side panels and ensure
the heading structure is semantic for screen readers.

## Proposal

Add a localized “Back to meeting tools” label, mark the arrow as decorative,
and move the back button outside the H1.

- [x] verify SR announces back button label in subpanels
- [x] verify H1 reads only the panel title (no button inside)

### DIFF
--- a/src/frontend/src/features/rooms/livekit/components/SidePanel.tsx
+++ b/src/frontend/src/features/rooms/livekit/components/SidePanel.tsx
@@ -13,6 +13,7 @@ import { Effects } from './effects/Effects'
 import { Admin } from './Admin'
 import { Tools } from './Tools'
 import { Info } from './Info'
+import { HStack } from '@/styled-system/jsx'
 
 type StyledSidePanelProps = {
   title: string
@@ -23,6 +24,7 @@ type StyledSidePanelProps = {
   closeButtonTooltip: string
   isSubmenu: boolean
   onBack: () => void
+  backButtonLabel: string
 }
 
 const StyledSidePanel = ({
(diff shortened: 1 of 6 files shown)


src/frontend/src/features/rooms/livekit/components/SidePanel.tsx
```diff
@@ -63,31 +65,34 @@ const StyledSidePanel = ({
...
-        paddingLeft: '1.5rem',
-        paddingTop: '1rem',
-        display: isClosed ? 'none' : 'flex',
-        justifyContent: 'start',
-        alignItems: 'center',
-      }}
-    >
+    <Div display="flex">
       {isSubmenu && (
         <Button
```
lebaudantoine, line 71: Hey! I noticed a couple of things while reviewing:

The background on hover has changed from a square to the new shape. Just wanted to flag this in case it was intentional:

Image

The layout looks a bit different now, which might be unexpected if the main goal of this PR is to add an aria-label. Here's a comparison:

Now:
Image

Before:
Image

Just wanted to make sure these changes were expected!

src/frontend/src/features/rooms/livekit/components/SidePanel.tsx
```diff
@@ -158,6 +162,7 @@ export const SidePanel = () => {
       })}
       isClosed={!isSidePanelOpen}
       isSubmenu={isSubPanelOpen}
+      backButtonLabel={t('backToTools')}
```
lebaudantoine, line 166: nit: This isn’t easily extensible if we add back buttons to other side panels, but it’s fine for now.

PR #885, opened 2026-01-20
🐛(brevo) use `django-lasuite` for marketing management

qbey:
## Purpose

When the user is updated their list are overwritten with the new value: this removes lists from other products. We switch to the common lib implementation which manages this.

Warning: Deployment will need an update...

```
    LASUITE_MARKETING_BACKEND: lasuite.marketing.backends.brevo.BrevoBackend
    LASUITE_MARKETING_PARAMETERS:
      secretKeyRef:
        name: backend
        key: LASUITE_MARKETING_PARAMETERS
```

and in the secrets add `LASUITE_MARKETING_PARAMETERS ` :

```
{"api_key": "XXX", "api_contact_list_ids": [1], "api_contact_attributes":{"VISIO_USER": True}}
```

## Proposal

- [x] replace custom implementation with the `django-lasuite` one
- [x] update documentation


lebaudantoine:
@coderabbitai full review

src/backend/core/authentication/backends.py
```diff
@@ -1,21 +1,14 @@
 """Authentication Backends for the Meet core app."""
 
-import contextlib
-
 from django.conf import settings
-from django.core.exceptions import ImproperlyConfigured, SuspiciousOperation
-from django.utils.translation import gettext_lazy as _
+from django.core.exceptions import SuspiciousOperation
 
+from lasuite.marketing.tasks import create_or_update_contact
```
lebaudantoine, line 6: I don't have any Celery configured along with my backend in production currently.
Could it be an issue?

I guess I would need to add one …

src/helm/env.d/dev/values.meet.yaml.gotmpl
```diff
@@ -87,11 +82,6 @@ backend:
     SUMMARY_SERVICE_API_TOKEN: password
     RECORDING_DOWNLOAD_BASE_URL: https://meet.127.0.0.1.nip.io/recording
     SIGNUP_NEW_USER_TO_MARKETING_EMAIL: True
-    BREVO_API_KEY:
-      secretKeyRef:
-        name: backend
-        key: BREVO_API_KEY
-    BREVO_API_CONTACT_LIST_IDS: 8
```
lebaudantoine: Why not add the new settings to this stack?

lebaudantoine (commented):
Thanks for the addition; it's straightforward, so let's merge it quickly.
In your PR description, you included an example where the contact list ID is set to 1. Should we keep it as 8 instead?

Should I release a new major version, given that the updated settings are not backward compatible?

It feels bad to do so for such minimal changes, especially since they may not even affect self-hosters in practice.

PR #887, opened 2026-01-20
🩹(backend) use case-insensitive email matching in the external api

lebaudantoine:
Fix a minor issue in the external API where users were matched using case-sensitive email comparison, while authentication treats emails as case-insensitive. This caused inconsistencies that are now resolved.

Spotted by T. Lemeur from Centrale.


lebaudantoine:
@coderabbitai full review

lebaudantoine:
@coderabbitai full review

PR #897, opened 2026-01-22
Authentication add in

lebaudantoine:
Wip

### DIFF
--- a/src/backend/core/urls.py
+++ b/src/backend/core/urls.py
@@ -42,6 +42,10 @@ external_router.register(
     basename="external_room",
 )
 
+
+addons_urls = addons_router.urls if settings.ADDONS_ENABLED else []
+
+
 urlpatterns = [
     path(
         f"api/{settings.API_VERSION}/",
@@ -49,21 +53,25 @@ urlpatterns = [
             [
                 *router.urls,
                 *oidc_urls,
-                *addons_router.urls,
+                *addons_urls,
                 path("config/", get_frontend_configuration, name="config"),
             ]
         ),
     ),
-    path(
-        "addons/",
-        include(
-            [
-                path("transit/", addons_views.transit_page, name="transit_page"),
-                path("redirect/", addons_views.redirect_page, name="redirect_page"),
-            ]
-        ),
-    ),
 ]
+
+if settings.ADDONS_ENABLED:
+    urlpatterns.append(
+        path(
+            "addons/",
+            include(
+                [
+                    path("transit/", addons_views.transit_page, name="transit_page"),
+                    path("redirect/", addons_views.redirect_page, name="redirect_page"),
+                ]
+            ),
+        ),
+    )
 
 if settings.EXTERNAL_API_ENABLED:
     urlpatterns.append(
--- a/src/backend/meet/settings.py
+++ b/src/backend/meet/settings.py
@@ -792,6 +792,11 @@ class Base(Configuration):
     )
 
     # Addons
+    ADDONS_ENABLED = values.BooleanValue(
+        False,
+        environ_name="ADDONS_ENABLED",
+        environ_prefix=None,
+    )
     ADDONS_SESSION_ID_LENGTH = values.PositiveIntegerValue(
         32,
         environ_name="ADDONS_SESSION_ID_LENGTH",


lebaudantoine:
@coderabbitai full review

lebaudantoine:
@coderabbitai full review

lebaudantoine:
@coderabbitai full review

lebaudantoine:
@coderabbitai full review

lebaudantoine:
@coderabbitai full review

lebaudantoine:
@coderabbitai full review

lebaudantoine:
@coderabbitai full review

src/backend/core/addons/viewsets.py
```diff
@@ -0,0 +1,53 @@
...
+
+    scope = "addons_auth_sessions"
+
+
+class AuthSessionViewSet(viewsets.ViewSet):
+    """ViewSet for managing add-on authentication sessions via token exchange."""
+
+    authentication_classes = []
+    permission_classes = []
+    throttle_classes = [AuthSessionThrottle]
```
qbey, line 27: Might be ok for a v0, did you consider having a burst rate and a sustained one? Also I'm wondering if the DRF's throttling won't hit the cache to read a value (but I guess it's ok).

We need to check if ou deployment configuration is properly set: https://www.django-rest-framework.org/api-guide/throttling/#how-clients-are-identified and maybe add a "legit IP list" to allow users from "known places" to not be blocked because their outgoing IP address.

qbey, line 27: Also, note the current throttling applies on the three endpoints (not a problem).

src/backend/core/services/jwt_token_service.py
```diff
@@ -0,0 +1,97 @@
...
+        """
+        Initialize the token service with custom settings.
+
+        Args:
+            secret_key: Secret key for JWT encoding/decoding
+            algorithm: JWT algorithm (default: HS256)
+            issuer: Token issuer identifier
+            audience: Token audience identifier
+            expiration_seconds: Token expiration time in seconds (default: 3600)
+            token_type: Token type (default: Bearer)
```
qbey, line 35: Default values are defined in settings, should it be here instead?

src/backend/core/services/jwt_token_service.py
```diff
@@ -0,0 +1,97 @@
...
+
+        Raises:
+            ImproperlyConfigured: If secret_key is None or empty
+        """
+        if not secret_key:
+            raise ImproperlyConfigured("Secret key is required.")
+
+        self._key = secret_key
+        self._alg = algorithm
```
qbey, line 44: I'm wondering what happens if `algorithm` is None, maybe we should prevent that?

Globally, should we raise `ImproperlyConfigured` if any value is falsy?

src/backend/core/addons/viewsets.py
qbey: We need to know when the throttle rate is hit, maybe you can use the same from Docs: https://github.com/suitenumerique/docs/blob/325c7d978661ccafa4554e373640ec5a7b19f8a9/src/backend/core/api/throttling.py#L10-L12

src/backend/core/urls.py
```diff
@@ -39,10 +49,20 @@
             [
                 *router.urls,
                 *oidc_urls,
+                *addons_router.urls,
```
qbey: Could you add a specific "switch" to be able to remove these routes in case of emergency?

src/backend/core/addons/viewsets.py
lebaudantoine: yes that's wanted.

src/backend/core/services/jwt_token_service.py
lebaudantoine: Yes you totally right; I would go for the `ImproperlyConfigured` + validation approach.

src/backend/core/urls.py
lebaudantoine: Absolutely agree. Having a kill switch is essential, we need the ability to immediately disable these endpoints if any issues arise.

src/backend/core/addons/viewsets.py
lebaudantoine: I did the same as docs as you suggested; it's now in production

PR #904, opened 2026-01-25
✨ (frontend) add configurable external redirect for unauthenticated users

lebaudantoine:
Offer a way to redirect unauthenticated users to an external home page when they visit the app, allowing a more marketing-focused entry point with a clearer value proposition.

In many self-hosted deployments, the default unauthenticated home page is not accessible or already redirects elsewhere. To ensure resilience, the client briefly checks that the target page is reachable and falls back to the default page if not.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,10 @@ and this project adheres to
 
 ## [Unreleased]
 
+### Added
+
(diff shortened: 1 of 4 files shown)


lebaudantoine:
@coderabbitai full review

lebaudantoine:
@coderabbitai full review

PR #907, opened 2026-01-26
✨(all) add/remove owners to your meeting

nono:
## Purpose

Allow room owners to promote other authenticated participants to owner role (and demote them) directly from the participants panel, without requiring them to leave the meeting.

This enables dynamic delegation of moderation powers during a meeting, which is useful when the original owner needs to leave or wants to share administrative responsibilities.

https://github.com/orgs/suitenumerique/projects/11?pane=issue&itemId=140746470

## Proposal

Add promote/demote functionality accessible via the 3-dot menu on participants:

  - [x] Backend: Add promote_participant and demote_participant endpoints with owner-only permissions
  - [x] Backend: Add is_owner field to room API response to distinguish owners from admins
  - [x] Backend: Add authenticated attribute to LiveKit tokens to identify authenticated vs anonymous users
  - [x] Backend: Send roleChanged notification to trigger real-time UI updates
  - [x] Frontend: Add "Make Owner" and "Remove Owner" menu items in participant menu
  - [x] Frontend: Add hooks to check owner status (useIsOwner, useIsParticipantOwner)
  - [x] Frontend: Listen for role change notifications to refresh permissions without page reload
  - [x] Tests: Add comprehensive test coverage for promote/demote endpoints (19 tests)
  
Screenshot 2026-01-26 at 09-31-33 LaSuite Meet

### DIFF
--- a/src/backend/core/api/permissions.py
+++ b/src/backend/core/api/permissions.py
@@ -106,3 +106,13 @@ class HasLiveKitRoomAccess(permissions.BasePermission):
         if not request.auth or not hasattr(request.auth, "video"):
             return False
         return request.auth.video.room == str(obj.id)
+
+
+class IsRoomOwner(IsAuthenticated):
+    """Check if user is an owner of the room."""
+
+    message = "You must be an owner of the room to perform this action."
+
+    def has_object_permission(self, request, view, obj):
+        """Determine if user is an owner of the room."""
+        return obj.is_owner(request.user)
--- a/src/backend/core/api/serializers.py
+++ b/src/backend/core/api/serializers.py
@@ -163,6 +163,7 @@ class RoomSerializer(serializers.ModelSerializer):
             del output["pin_code"]
 
         output["is_administrable"] = is_admin_or_owner
+        output["is_owner"] = models.RoleChoices.check_owner_role(role)
 
         return output
 
@@ -309,3 +310,37 @@ class UpdateParticipantSerializer(BaseParticipantsManagementSerializer):
                 ) from e
 
         return attrs
+
+
+class PromoteParticipantSerializer(BaseValidationOnlySerializer):
+    """Validate promote participant to owner request."""
+
+    participant_identity = serializers.CharField(
+        help_text="LiveKit participant identity (user sub)"
+    )
+
+    def validate_participant_identity(self, value):
+        """Validate participant_identity is a valid authenticated user sub."""
+        # Verify the user exists - anonymous users don't have User records
+        if not models.User.objects.filter(sub=value).exists():
+            raise serializers.ValidationError(
+                "No user found with this identity. Anonymous participants cannot be promoted."
+            )
+
+        return value
+
+
+class DemoteParticipantSerializer(BaseValidationOnlySerializer):
+    """Validate demote participant from owner request."""
+
+    participant_identity = serializers.CharField(
+        help_text="LiveKit participant identity (user sub)"
+    )
+
+    def validate_participant_identity(self, value):
+        """Validate participant_identity corresponds to a valid user."""
+        # Verify the user exists
+        if not models.User.objects.filter(sub=value).exists():
+            raise serializers.ValidationError("No user found with this identity.")
+
+        return value
--- a/src/backend/core/api/viewsets.py
+++ b/src/backend/core/api/viewsets.py
@@ -670,6 +670,162 @@ class RoomViewSet(
             {"status": "success"}, status=drf_status.HTTP_200_OK
         )
 
+    @decorators.action(
+        detail=True,
+        methods=["post"],
+        url_path="promote-participant",
+        url_name="promote-participant",
+        permission_classes=[permissions.IsRoomOwner],
+    )
+    def promote_participant(self, request, pk=None):  # pylint: disable=unused-argument
+        """Promote a participant to room owner.
+
+        This endpoint allows room owners to grant owner privileges to authenticated
+        participants currently in the meeting. Anonymous participants cannot be promoted.
+        """
+        room = self.get_object()
+
+        serializer = serializers.PromoteParticipantSerializer(data=request.data)
+        serializer.is_valid(raise_exception=True)
+
+        participant_identity = serializer.validated_data["participant_identity"]
+
+        # Get the user by their sub (identity)
+        try:
+            user = models.User.objects.get(sub=participant_identity)
+        except models.User.DoesNotExist:
+            return drf_response.Response(
+                {"error": "User not found"},
+                status=drf_status.HTTP_404_NOT_FOUND,
+            )
+
+        # Create or update ResourceAccess with OWNER role
+        _, created = models.ResourceAccess.objects.update_or_create(
+            resource=room,
+            user=user,
+            defaults={"role": models.RoleChoices.OWNER},
+        )
+
+        # Update LiveKit participant's room_admin attribute
+        try:
+            ParticipantsManagement().update(
+                room_name=str(room.pk),
+                identity=participant_identity,
+                attributes={"room_admin": "true"},
+            )
+        except ParticipantsManagementException:
+            # Log but don't fail - the database update succeeded
+            logger.warning(
+                "Failed to update LiveKit attributes for promoted participant %s",
+                participant_identity,
+            )
+
+        # Notify participants about role change so they can refresh their data
+        try:
+            utils.notify_participants(
+                room_name=str(room.pk),
+                notification_data={"type": "roleChanged"},
+            )
+        except utils.NotificationError:
+            logger.warning("Failed to notify participants about role change")
+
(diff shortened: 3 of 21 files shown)


lebaudantoine:
@coderabbitai full review

lebaudantoine:
That looks super promising. I'm quite busy this week, but I've started reading the code. @nono should we jump on a call?

src/backend/core/api/serializers.py
```diff
@@ -309,3 +310,37 @@ def validate(self, attrs):
...
+        help_text="LiveKit participant identity (user sub)"
+    )
+
+    def validate_participant_identity(self, value):
+        """Validate participant_identity corresponds to a valid user."""
+        # Verify the user exists
+        if not models.User.objects.filter(sub=value).exists():
+            raise serializers.ValidationError("No user found with this identity.")
+
+        return value
```
lebaudantoine, line 346: These two serializers are highly duplicated and could clearly be factorized. I would also keep serializers thin: validate that the identity is a UUID rather than a generic CharField, and move the data querying logic to the viewset.

src/backend/core/api/permissions.py
```diff
@@ -106,3 +106,13 @@ def has_object_permission(self, request, view, obj):
...
+
+
+class IsRoomOwner(IsAuthenticated):
+    """Check if user is an owner of the room."""
+
+    message = "You must be an owner of the room to perform this action."
+
+    def has_object_permission(self, request, view, obj):
+        """Determine if user is an owner of the room."""
+        return obj.is_owner(request.user)
```
lebaudantoine, line 118: I’d add this to the RoomPermissions class instead, which results in less code. The demote and promote actions would be treated as delete operations. You could probably retrieve it from the view object.

src/backend/core/api/viewsets.py
```diff
@@ -670,6 +670,162 @@ def remove_participant(self, request, pk=None):  # pylint: disable=unused-argume
...
         )
 
+    @decorators.action(
+        detail=True,
+        methods=["post"],
+        url_path="promote-participant",
+        url_name="promote-participant",
+        permission_classes=[permissions.IsRoomOwner],
+    )
+    def promote_participant(self, request, pk=None):  # pylint: disable=unused-argument
```
lebaudantoine, line 680: We usually try to avoid putting too much logic in the viewset. Instead, we prefer creating a dedicated service with clear responsibilities. We could even consider adding promote and demote methods to an existing service if that fits, perhaps the participant_management service.

src/backend/core/api/viewsets.py
```diff
@@ -670,6 +670,162 @@ def remove_participant(self, request, pk=None):  # pylint: disable=unused-argume
...
+
+        serializer = serializers.PromoteParticipantSerializer(data=request.data)
+        serializer.is_valid(raise_exception=True)
+
+        participant_identity = serializer.validated_data["participant_identity"]
+
+        # Get the user by their sub (identity)
+        try:
+            user = models.User.objects.get(sub=participant_identity)
+        except models.User.DoesNotExist:
```
lebaudantoine, line 696: That’s what I meant. In your implementation, two SQL queries are executed: one in the serializer to check existence, and another in the viewset to retrieve the user. Let’s reduce this to a single query.

src/backend/core/api/viewsets.py
```diff
@@ -670,6 +670,162 @@ def remove_participant(self, request, pk=None):  # pylint: disable=unused-argume
...
+            return drf_response.Response(
+                {"error": "User not found"},
+                status=drf_status.HTTP_404_NOT_FOUND,
+            )
+
+        # Create or update ResourceAccess with OWNER role
+        _, created = models.ResourceAccess.objects.update_or_create(
+            resource=room,
+            user=user,
+            defaults={"role": models.RoleChoices.OWNER},
```
lebaudantoine, line 706: brainsto: I might go with administrator instead of owner to avoid dealing with “last owner” edge cases in demote and to avoid having to differentiate between the original owner and administrators.

src/backend/core/api/viewsets.py
```diff
@@ -670,6 +670,162 @@ def remove_participant(self, request, pk=None):  # pylint: disable=unused-argume
...
+        )
+
+        # Update LiveKit participant's room_admin attribute
+        try:
+            ParticipantsManagement().update(
+                room_name=str(room.pk),
+                identity=participant_identity,
+                attributes={"room_admin": "true"},
```
lebaudantoine, line 714: this would likely overwrite all existing participant attributes, if any.

Also, relying on participant attributes isn’t safe. Participants are allowed to update their own metadata and could call the LiveKit API to set their attributes to "room_admin": "true".

src/backend/core/tests/rooms/test_api_rooms_demote.py
```diff
@@ -0,0 +1,266 @@
+"""
+Test rooms API endpoints in the Meet core app: demote participant from owner.
```
lebaudantoine, line 2: wip no review

src/backend/core/tests/rooms/test_api_rooms_promote.py
```diff
@@ -0,0 +1,245 @@
+"""
+Test rooms API endpoints in the Meet core app: promote participant to owner.
```
lebaudantoine, line 2: wip no review

src/frontend/src/features/rooms/api/ApiRoom.ts
```diff
@@ -10,15 +10,35 @@ export enum ApiAccessLevel {
...
+  role: ApiAccessRole
+}
+
 export type ApiRoom = {
   id: string
   name: string
   slug: string
   pin_code: string
   is_administrable: boolean
+  is_owner: boolean
```
lebaudantoine, line 37: If we now return accesses and consume them on the frontend, we probably shouldn’t also return a boolean flag alongside them. That creates a risk of desynchronization. The boolean should be derived from the accesses instead.

src/frontend/src/features/rooms/livekit/components/ParticipantMenu/ParticipantMenuButton.tsx
```diff
@@ -12,6 +12,7 @@ export const ParticipantMenuButton = ({
 }) => {
   const { t } = useTranslation('rooms', { keyPrefix: 'participants' })
   const isAdminOrOwner = useIsAdminOrOwner()
+
```
lebaudantoine, line 15: wip useless diff

src/backend/core/api/viewsets.py
```diff
@@ -670,6 +670,162 @@ def remove_participant(self, request, pk=None):  # pylint: disable=unused-argume
...
+                "Failed to update LiveKit attributes for promoted participant %s",
+                participant_identity,
+            )
+
+        # Notify participants about role change so they can refresh their data
+        try:
+            utils.notify_participants(
+                room_name=str(room.pk),
+                notification_data={"type": "roleChanged"},
+            )
```
lebaudantoine, line 728: all participants are notified of a role change, but anonymous users will refresh the room data without being able to retrieve updated access rights. They’ll refetch the room data for no reason. We should avoid these unnecessary API calls.

src/backend/core/utils.py
lebaudantoine, line 93: To safely support this feature (and others), we probably need to reconsider this setting and better understand what’s tied to can_update_own_metadata = False. We would start with a dedicated PR focused on updating this value to False and addressing its implications.

src/backend/core/api/viewsets.py
```diff
@@ -670,6 +670,162 @@ def remove_participant(self, request, pk=None):  # pylint: disable=unused-argume
...
+        # Create or update ResourceAccess with OWNER role
+        _, created = models.ResourceAccess.objects.update_or_create(
+            resource=room,
+            user=user,
+            defaults={"role": models.RoleChoices.OWNER},
+        )
+
+        # Update LiveKit participant's room_admin attribute
+        try:
+            ParticipantsManagement().update(
```
lebaudantoine, line 711: The most important missing piece is the LiveKit video grants.

The tricky part is that updating roomAdmin via the LiveKit API isn’t possible. Admin rights are only granted through the JWT token’s `video.roomAdmin` claim at join time, or by issuing a brand-new token with that claim.

That means we can’t promote a participant to room admin dynamically through the API. The only viable approach would be to update the user state in the database and force the participant to refresh their token, so they rejoin with a newly issued token that includes roomAdmin=true.

I guess we should make sure, if they refresh their token, they won't disconnect from the room, and the operation happens seamlessly

lebaudantoine:
@nono Do you still want to work on this PR? If not I'm going to close it and work on it in another PR.

nono:
@lebaudantoine let's close the PR. I don't have time to work on it for the moment, and I misunderstood the level of changes that this feature requires. I thought it would be easier.

PR #908, opened 2026-01-26
♿(frontend) adjust sr announcements for idle disconnect timer

Ovgodd:
## Purpose

Keep the idle timeout announcements helpful without overwhelming SR users.

## Proposal

Announce the countdown only at key moments (120/90/60/30 and last 10s).

- [x] confirm thresholds with UX/accessibility
- [x] verify announcements in NVDA/JAWS/VoiceOver


Ovgodd:
@coderabbitai full review

PR #911, opened 2026-01-27
🔒️(frontend) fix an XSS vulnerability on the recording page

lebaudantoine:
An XSS vulnerability was identified by an open-source contributor. While the impact was limited, only a room owner could inject the content and then view the recording page, it is important to address, especially before introducing multi-owner support.


lebaudantoine:
@coderabbitai full review

PR #923, opened 2026-01-28
Fix documentation rendering

lebaudantoine:
## Purpose

Description...

## Proposal

Description...

- [] item 1...
- [] item 2...

### DIFF
--- a/docs/installation/compose.md
+++ b/docs/installation/compose.md
@@ -181,9 +181,9 @@ You will need to uncomment the environment and network sections in compose file 
 Expose the Frontend port to the host
 ```yaml
 frontend:
-…
-	ports:
-		- “8086:8086”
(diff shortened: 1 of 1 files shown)


rexbron:
Interesting. My markdown editor replaces ... with the unicode triple dot glyph and it seems that github chokes on that?

lebaudantoine:
Actually, GitHub chokes on the end of file you introduced. I'll revert my changes on the ellipsis.

rexbron:
Very odd! I'm using a Markdown editor called Paper on MacOS. Do you know what character caused the issue? 

I can file a bug report with the developer.

lebaudantoine:
You can compare the last two commits to spot the characters that caused the issue.

PR #931, opened 2026-01-30
♿️(frontend) Improve spinner reduced‑motion fallback

Ovgodd:
## Purpose

Improve loading feedback for users with reduced motion by providing a clear, static fallback.

cssreduce

## Proposal

Hide the animated SVG spinner when `prefers-reduced-motion: reduce` and show an hourglass instead; move animation styles into a reusable CSS class.

- [x] Verify reduced‑motion fallback shows ⏳

### DIFF
--- a/src/frontend/src/primitives/Spinner.tsx
+++ b/src/frontend/src/primitives/Spinner.tsx
@@ -1,5 +1,12 @@
 import { ProgressBar } from 'react-aria-components'
-import { css } from '@/styled-system/css'
+import { RiHourglassFill } from '@remixicon/react'
+import { css, cx } from '@/styled-system/css'
+
+const rotatingArcClassName = css({
+  animation: 'rotate 1s ease-in-out infinite',
+  transformOrigin: 'center',
+  transition: 'transform 16ms linear',
+})
 
 export const Spinner = ({
   size = 56,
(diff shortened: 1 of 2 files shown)


src/frontend/src/primitives/Spinner.tsx
```diff
@@ -13,44 +19,71 @@ export const Spinner = ({
   const r = 14 - strokeWidth
   const c = 2 * r * Math.PI
   return (
-    <ProgressBar aria-label="Loading…" value={30}>
```
lebaudantoine, line 16: is better to use `...` instead of `…` in terms of accessibility?

src/frontend/src/primitives/Spinner.tsx
```diff
@@ -13,44 +19,71 @@ export const Spinner = ({
...
             })}
-            style={{
-              animation: `rotate 1s ease-in-out infinite`,
-              transformOrigin: 'center',
-              transition: 'transform 16ms linear',
-            }}
-          />
-        </svg>
+          >
+            ⏳
```
lebaudantoine: hey, I would use a remix icon hourglass, the icon contrast is not great on the recording toaster : 

Image

Ovgodd: Ok ! no problem and I agree

PR #944, opened 2026-02-02
🥅(summary) catch unexpected file-related exceptions when handling recording objects

lebaudantoine:
Previously, if a recording file was not found in the bucket, the code would crash. This adds proper error handling to avoid unhandled failures.


lebaudantoine:
@coderabbitai full review

PR #957, opened 2026-02-06
🚀(paas) add PaaS deployment scripts, tested on Scalingo

zeylos:
Replaces https://github.com/suitenumerique/meet/pull/668

- fixed https://github.com/suitenumerique/meet/pull/668#discussion_r2282228043
- fixed https://github.com/suitenumerique/meet/pull/668#discussion_r2281960226
- fixed https://github.com/suitenumerique/meet/pull/668#discussion_r2281976789 by migrating to a `deploy/paas` directory, making room for a deploy/docker directory later
- about `Also, just a note: /media and /media-auth are required in Meet for handling recordings!` for now this would require a lot of setup to make it work with PaaS because it needs an external S3 (that would require setting up a custom domain and CORS) and if I'm not wrong for now this part depends on minio anyway so I didn't bother

Tested successfuly on Scalingo

### DIFF
--- a/docs/installation/scalingo.md
+++ b/docs/installation/scalingo.md
@@ -1,0 +1,185 @@
+# Deployment on Scalingo
+
+This guide explains how to deploy La Suite Meet on [Scalingo](https://scalingo.com/) using the [Suite Numérique buildpack](https://github.com/suitenumerique/buildpack).
+
+## Overview
+
+Scalingo is a Platform-as-a-Service (PaaS) that simplifies application deployment. This setup uses a custom buildpack to handle both the frontend (Vite) and backend (Django) builds, serving them through Nginx.
+
+## Prerequisites
+
+- A Scalingo account
+- Scalingo CLI installed (optional but recommended)
+- A PostgreSQL database addon
+- A Redis addon (for caching and sessions)
+
+## Step 1: Create Your App
+
+Create a new app on Scalingo using `scalingo` cli or using the [Scalingo dashboard](https://dashboard.scalingo.com/).
+
+## Step 2: Provision Addons
+
+Add the required PostgreSQL and Redis services.
+
+This will set the following environment variables automatically:
+- `SCALINGO_POSTGRESQL_URL` - Database connection string
+- `SCALINGO_REDIS_URL` - Redis connection string
+
+## Step 3: Configure Environment Variables
+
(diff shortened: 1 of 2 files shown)


src/backend/meet/settings.py
```diff
@@ -92,7 +93,9 @@ class Base(Configuration):
 
     # Database
     DATABASES = {
-        "default": {
+        "default": dj_database_url.config()
+        if environ.get("DATABASE_URL")
```
lebaudantoine: Could you follow the same approach used for other environment variables, using `values.Value` with a default of None?

zeylos: see https://github.com/suitenumerique/meet/pull/957/changes/8b3fbbc7a2da1bfb8ffa0ea9d7a93e575fa04d90, reading https://django-configurations.readthedocs.io/en/stable/values.html#configurations.values.DatabaseURLValue it looks like the right way is to use DatabaseURLValue ?

zeylos: also added https://github.com/suitenumerique/meet/pull/957/changes/314e4fe0515974f85481b846e3ddf1fc68fd76be with the documentation, does that match with what you expected ?

zeylos:
@lebaudantoine Could you review [this change](https://github.com/suitenumerique/meet/pull/957/changes/e82fc8c5530ac803cb3b11b055cfb0aae85cd424) please ? sylvinus told me the assets part will probably be reworked later, but we need the custom logo now on LST and we can't just push files on Scalingo deployments

docs/installation/scalingo.md
```diff
@@ -0,0 +1,185 @@
+# Deployment on Scalingo
+
+This guide explains how to deploy La Suite Meet on [Scalingo](https://scalingo.com/) using the [Suite Numérique buildpack](https://github.com/suitenumerique/buildpack).
+
+## Overview
+
+Scalingo is a Platform-as-a-Service (PaaS) that simplifies application deployment. This setup uses a custom buildpack to handle both the frontend (Next.js) and backend (Django) builds, serving them through Nginx.
```
lebaudantoine: nit: we're not using Next.js! but Vite.
You can fix it before merging, or in a dedicated PR.

lebaudantoine (approved):

PR #963, opened 2026-02-09
🔒️(backend) add application validation when consuming external API JWT tokens

lebaudantoine:
Token generation already verifies that the application is active, but this guarantee was not enforced when the token was used. This change adds a runtime check to ensure the client_id claim matches an existing and active application when evaluating permissions.

This also introduces an emergency revocation mechanism, allowing all previously issued tokens for a given application to be invalidated if the application is disabled.


lebaudantoine:
@coderabbitai full review

lebaudantoine:
@coderabbitai full review

PR #1005, opened 2026-02-23
♻️(frontend) use react aria toolbar for arrow key navigation

Ovgodd:
## Purpose

Follow WAI-ARIA toolbar pattern for emoji reactions: arrow keys to navigate, Tab to dismiss.

## Proposal

- [x] Replace `div[role=toolbar]` + `FocusScope contain` with React Aria `Toolbar` (roving tabindex)
- [x] Tab closes the toolbar and restores focus to the trigger instead of trapping
- [x] ctrl shift E refocus on first emoji when the toolbar is already open


CHANGELOG.md
```diff
@@ -18,8 +18,7 @@ and this project adheres to
 - 🔒️(frontend) uninstall curl from the frontend production image #987
 - 💄(frontend) add focus ring to reaction emoji buttons
 - ✨(frontend) introduce a shortcut settings tab #975
-- 🚚(frontend) rename "wellknown" directory to "well-known" #1009
-- 🌐(frontend) localize SR modifier labels #1010
+- ♻️(frontend) use react aria toolbar for arrow key navigation #1005
```
lebaudantoine, line 25: something weird there.

lebaudantoine:
Addressed in #1216

PR #1008, opened 2026-02-23
wip introduce add-in endpoints and views

lebaudantoine:
### How it works?

```mermaid
sequenceDiagram
    actor Client
    participant Addon as meet.com/addons
    participant TES as Token Exchange Service
    participant Cache
    participant OIDC as OIDC Provider

    Client->>Addon: POST /addons/sessions
    Addon->>TES: Generate session ID
    TES->>Cache: Init cache[session_id] = null
    TES-->>Addon: session_id
    Addon-->>Client: { session_id }
    Client->>Client: Store session_id

    loop Long Polling
        Client->>Addon: GET /addons/sessions/:session_id
        Addon->>Cache: Read cache[session_id]
        Cache-->>Addon: null (not ready)
        Addon-->>Client: 202 Pending
    end

    Client->>Client: Open dialog → meet.com/addons/transit/?session_id=xx

    Client->>Addon: GET /addons/transit/?session_id=xx
    Addon->>Addon: Store session_id in session/state
    Addon-->>Client: Redirect to OIDC /authorize?redirect_uri=/addons/redirect

   Client->>OIDC: Authenticate
    OIDC-->>Client: Redirect to meet.com/callback?code=xx

    Client->>Addon: GET /callback?code=xx
    Addon->>OIDC: Exchange code for tokens
    OIDC-->>Addon: id_token, access_token
    Addon-->>Client: Redirect to /addons/redirect

    Client->>Addon: GET /addons/redirect
    Addon->>Addon: Forge JWT (access_token)
    Addon->>Cache: Set cache[session_id] = access_token
    Addon-->>Client: Redirect to success page

    Note over Client,Cache: Cache now populated — long poll resolves

    Client->>Addon: GET /addons/sessions/:session_id
    Addon->>Cache: Read cache[session_id]
    Cache-->>Addon: access_token
    Addon-->>Client: 200 { access_token }
    Client->>Client: Store access_token

    Client->>Addon: API Request (Authorization: Bearer access_token)
    Addon-->>Client: Protected resource response
```


src/frontend/default.conf
```diff
@@ -11,6 +11,12 @@ server {
       add_header Content-Disposition "attachment; filename=windows-app-web-link";
   }
 
+  location ~ ^/outlook-addin(/.*)?$ {
+      alias /usr/share/nginx/html/outlook-addin$1;
+      add_header Access-Control-Allow-Origin "*";
+      add_header Cache-Control "no-cache, no-store, must-revalidate";
+  }
```
lebaudantoine, line 47: wip to be added conditionally

src/backend/core/addons/service.py
```diff
@@ -0,0 +1,124 @@
...
+        internal_fields = {"expires_at"}
+        return {k: v for k, v in data.items() if k not in internal_fields}
+
+    def clear_session(self, session_id: str) -> None:
+        """Remove session data from cache."""
+
+        cache_key = self._get_cache_key(session_id)
+        cache.delete(cache_key)
+
+    def set_access_token(self, user: User, session_id: str):
```
lebaudantoine, line 144: rethinking it,  should return a short-lived access token and a long live refresh token, and support a refresh mechanism.

src/backend/core/addons/viewsets.py
```diff
@@ -0,0 +1,47 @@
...
+
+    authentication_classes = []
+    permission_classes = []
+    throttle_classes = []
+
+    def create(self, request):
+        """Create a new pending authentication session."""
+        session_id = TokenExchangeService().init_session()
+        return drf_response.Response(
+            {"session_id": session_id}, status=drf_status.HTTP_201_CREATED
```
lebaudantoine: I was thinking about introducing a more complex mechanism for security purposes.
The idea would be to return a session ID in the response payload, but also return a second identifier through an HTTP-only secure cookie.

Access to the authentication result would require both identifiers. This way, even if the session ID in the payload were stolen, only the original client—who has the secure cookie—would be able to read the result from the cache.

That said, it might be overkill.

src/backend/core/addons/service.py
lebaudantoine, line 144: and actually, I don't know to what extent it's feasible, but it would be amazing to use a secure http-only cookie for the refresh token

src/backend/core/external_api/authentication.py
lebaudantoine, line 91: exception when token is expired is not properly caught

Issue #764, opened 2025-10-24
Noise reduction on stereo mic input produces a stream with only left channel


K900:
## Bug Report

**Problematic behavior**
When connecting to a call with a stereo mic, like on some modern laptops, and enabling the noise suppression feature, the output stream is still stereo, but only the left channel has sound - the right channel is completely empty. The stream should probably be either relabeled to mono, or each channel should be processed and streamed separately.

**Expected behavior/code**
Either mono sound or stereo sound with both channels.

**Steps to Reproduce**
1. Join a call with a stereo mic
2. Enable noise suppression
4. And then the bug happens!
**Environment**
- Meet version: 0.1.41
- Platform: NixOS / Firefox Nightly

**Possible Solution**
Need to look at the implementation, but rnnoise works on a single audio stream so you probably need to run two copies, or mix the stream down to mono first?

**Additional context/Screenshots**
Originally discovered with a Lenovo Thinkbook 14 G7 IML.

lebaudantoine:
Thank you for raising this issue, @K900 you’re absolutely right, a few users have also reported this bug. I haven’t had time yet to review the implementation, but @arnaud-robin led the groundwork on this feature.

Your help would be very welcome, even just identifying the root cause or a potential solution would be great. I can handle the code changes afterward.

I’ll be off for the next two weeks, but I’ll be happy to fix it once I’m back.

Best,

RaitoBezarius:
@lebaudantoine Friendly ping boss :-)

lebaudantoine:
A contributor is currently developing a new implementation that could potentially resolve this issue. I’ll keep you updated.

RaitoBezarius:
@lebaudantoine bro i am suffering

lebaudantoine:
PR #1015 is under review 🫡

FloChehab:
I confirmed https://github.com/suitenumerique/meet/pull/1015 is solving this issue, I'll put energy towards merging it ! Sorry for the delay.

PR #1015, opened 2026-02-24
Advanced Voice Optimization (BBBA)

trummerschlunk:
## Purpose
Better Sound for LaSuite Meet.

## Proposal

This pull request introduces [BBBA](https://github.com/trummerschlunk/BigBlueBetterAudio), a powerful voice optimization audio processor for LaSuite Meet. The solution runs entirely on the client, utilizing a WASM AudioWorklet inserted into the LaSuite Meet client's Livekit audio chain.

Features:
- Voice Isolation: Leverages  [RNNoise](https://github.com/xiph/rnnoise), a machine learning model for effective real-time background noise reduction.
- Voice Optimization: Utilizes a custom-designed vocal enhancement process, crafted by [an experienced mix and mastering engineer](https://4ohm.de) with over 25 years of expertise in creating great vocal sound. This algorithm shapes the audio to make speech more natural, pleasant, and intelligible, while reducing distractions.

The result is significantly improved speech clarity and listening experience for all participants.

## Demo Video

https://github.com/user-attachments/assets/3bb74a3d-317a-4de1-9caa-06f866f3b1f3

### DIFF
--- a/src/frontend/package.json
+++ b/src/frontend/package.json
@@ -17,6 +17,7 @@
   "dependencies": {
     "@fontsource-variable/atkinson-hyperlegible-next": "5.2.6",
     "@fontsource-variable/lexend": "5.2.11",
+    "@libreaudio/la-call": "0.1.4",
     "@fontsource/opendyslexic": "5.2.5",
     "@livekit/components-react": "2.9.21",
     "@livekit/components-styles": "1.2.0",
--- a/src/frontend/src/features/rooms/components/Join.tsx
+++ b/src/frontend/src/features/rooms/components/Join.tsx
@@ -216,6 +216,14 @@ export const Join = ({
       try {
         const track = await createLocalAudioTrack({
           deviceId: { exact: audioDeviceId },
+          noiseSuppression: true,
+          echoCancellation: true,
+          autoGainControl: true,
+          voiceIsolation: false,
+          // Audio quality optimized for voice
+          sampleRate: 48000, // High quality sample rate
+          channelCount: 1, // Mono for voice calls (saves bandwidth)
+          sampleSize: 16, // 16-bit audio
         })
         setDynamicAudioTrack(track)
       } catch (error) {
--- a/src/frontend/src/features/rooms/livekit/processors/RnnNoiseProcessor.ts
+++ b/src/frontend/src/features/rooms/livekit/processors/RnnNoiseProcessor.ts
@@ -1,9 +1,6 @@
-import type { Track, TrackProcessor, ProcessorOptions } from 'livekit-client'
-import { NoiseSuppressorWorklet_Name } from '@timephy/rnnoise-wasm'
 
-// This is an example how to get the script path using Vite, may be different when using other build tools
-// NOTE: `?worker&url` is important (`worker` to generate a working script, `url` to get its url to load it)
-import NoiseSuppressorWorklet from '@timephy/rnnoise-wasm/NoiseSuppressorWorklet?worker&url'
+import type { Track, TrackProcessor, ProcessorOptions } from 'livekit-client'
+import { createWasmProcessor } from '@libreaudio/la-call'
 
 // Use Jitsi's approach: maintain a global AudioContext variable
 // and suspend/resume it as needed to manage audio state
@@ -20,7 +17,7 @@ export class RnnNoiseProcessor implements AudioProcessorInterface {
   private source?: MediaStreamTrack
   private sourceNode?: MediaStreamAudioSourceNode
   private destinationNode?: MediaStreamAudioDestinationNode
-  private noiseSuppressionNode?: AudioWorkletNode
+  private noiseSuppressionNode?: AudioNode
 
   async init(opts: ProcessorOptions<Track.Kind.Audio>) {
     if (!opts.track) {
@@ -35,25 +32,16 @@ export class RnnNoiseProcessor implements AudioProcessorInterface {
       await audioContext.resume()
     }
 
-    await audioContext.audioWorklet.addModule(NoiseSuppressorWorklet)
-
     this.sourceNode = audioContext.createMediaStreamSource(
       new MediaStream([this.source])
     )
 
-    this.noiseSuppressionNode = new AudioWorkletNode(
-      audioContext,
-      NoiseSuppressorWorklet_Name,
-      {
-        // RNNoise is a mono algorithm. Its worklet only denoises and writes
-        // channel 0. Force any (possibly stereo) input to down-mix to a single
-        // channel and emit a single channel, so we don't produce a stereo frame
-        // whose right channel is left silent.
-        channelCount: 1,
-        channelCountMode: 'explicit',
-        outputChannelCount: [1],
-      }
-    )
+    this.noiseSuppressionNode = await createWasmProcessor(audioContext, {
+      intensity: 90,
+    })
+    if (!this.noiseSuppressionNode) {
+      throw new Error('Failed to create Wasm processor')
+    }
 
     this.destinationNode = audioContext.createMediaStreamDestination()
 
--- a/src/frontend/src/features/rooms/livekit/processors/RnnNoiseProcessor.ts
+++ b/src/frontend/src/features/rooms/livekit/processors/RnnNoiseProcessor.ts
@@ -1,10 +1,29 @@
-
 import type { Track, TrackProcessor, ProcessorOptions } from 'livekit-client'
-import { createWasmProcessor } from '@libreaudio/la-call'
 
 // Use Jitsi's approach: maintain a global AudioContext variable
 // and suspend/resume it as needed to manage audio state
 let audioContext: AudioContext
+
+/**
+ * Lazily load the WASM processor factory.
+ *
+ * '@libreaudio/la-call' pulls in a WebAssembly payload, so we defer importing
+ * it until a processor is actually initialized rather than at module load.
+ * The import promise is cached so repeated init() calls reuse the same module.
+ */
+type CreateWasmProcessor =
+  (typeof import('@libreaudio/la-call'))['createWasmProcessor']
+
+let wasmProcessorPromise: Promise<CreateWasmProcessor> | undefined
+
+function loadWasmProcessor(): Promise<CreateWasmProcessor> {
+  if (!wasmProcessorPromise) {
+    wasmProcessorPromise = import('@libreaudio/la-call').then(
+      (mod) => mod.createWasmProcessor
+    )
+  }
+  return wasmProcessorPromise
+}
 
 export interface AudioProcessorInterface extends TrackProcessor<Track.Kind.Audio> {
   name: string
@@ -36,6 +55,7 @@ export class RnnNoiseProcessor implements AudioProcessorInterface {
       new MediaStream([this.source])
     )
 
+    const createWasmProcessor = await loadWasmProcessor()
     this.noiseSuppressionNode = await createWasmProcessor(audioContext, {
       intensity: 90,
     })
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,10 @@ and this project adheres to
 
 ## [Unreleased]
 
+### Changed
+
+- ✨(frontend) enhance noise reduction with BBBA audio processing pipeline
+
 ## [1.20.0] - 2026-06-12
 
 ### Changed
--- a/src/frontend/package-lock.json
+++ b/src/frontend/package-lock.json
@@ -11,6 +11,7 @@
         "@fontsource-variable/atkinson-hyperlegible-next": "5.2.6",
         "@fontsource-variable/lexend": "5.2.11",
         "@fontsource/opendyslexic": "5.2.5",
+        "@libreaudio/la-call": "0.1.4",
         "@livekit/components-react": "2.9.21",
         "@livekit/components-styles": "1.2.0",
         "@livekit/track-processors": "0.7.2",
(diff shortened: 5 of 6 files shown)


trummerschlunk:
With our latest commit to this PR, we introduce different handling of audio processing on android devices. The situation is as follows:
- audio worklets (the new and proper way to do web audio things) run at a fixed 128 quantum size
- mobile phones and low-power devices are not able to do any non-trivial processing with such small sizes
- there are WIP specs for allowing to increase quantum size on audio worklets, but no browser has implemented them yet
- to prevent regressions (audio being terrible on mobile and low-power devices) we introduced support for the old web audio way (script processor) as a temporary measure
- once browsers implement audio worklet quantum "render hint" we should reevaluate the situation

lebaudantoine:
@coderabbitai full review

FloChehab:
Hi @trummerschlunk, thanks a lot for your PR, I am sorry we have been very busy lately and hadn't time to put the effort required into it.

I have tested it today and it does work pretty well from my point of view : 
* It fixes https://github.com/suitenumerique/meet/issues/764
* It performs better than the current "beta" solution in filtering out noises, and the voice does feel nicier.

On the technical side, I haven't dived too much, but would it possible for you to package your processing pipeline / wasm / etc. directly on npm so that we don't add those binary files directly on our repo?

I am also seing some mention to specific license in the files, which I am not sure we can pull in like this in this repo.

In all cases, thanks a lot for your work so far !

trummerschlunk:
> I have tested it today and it does work pretty well from my point of view

great, happy to hear :)

> On the technical side, I haven't dived too much, but would it possible for you to package your processing pipeline / wasm / etc. directly on npm so that we don't add those binary files directly on our repo?

Yes, we were planning to do this and are looking into it. Might take a week or two though...

> I am also seing some mention to specific license in the files, which I am not sure we can pull in like this in this repo.

I'll check.

Cheers :)

FloChehab:
Awesome, thanks @trummerschlunk !

trummerschlunk:
What are your license requirements for BBBA?

FloChehab:
Sorry i forgot to reply, I think the easiest for us is to go for MIT @trummerschlunk.

trummerschlunk:
Would GPL3 work for the npm package?

FloChehab:
> Would GPL3 work for the npm package?

We have forwarded your question to the relevant people, we'll get back to you asap on that matter.

lebaudantoine:
I got some feedback from the OSPO team: GPL would be acceptable if we can distribute an image without your code/dependency included, and provide a build flag for self-hosters or anyone wanting to modify LaSuite Meet without being required to republish it under AGPL.

MIT would obviously make things much simpler from an integration and distribution standpoint, but I completely understand the rationale behind choosing GPL in your case.

src/frontend/src/features/rooms/livekit/processors/RnnNoiseProcessor.ts
```diff
@@ -35,16 +31,14 @@ export class RnnNoiseProcessor implements AudioProcessorInterface {
       await audioContext.resume()
     }
 
-    await audioContext.audioWorklet.addModule(NoiseSuppressorWorklet)
-
```
lebaudantoine, line 39: Toggling several times in a row the noise reduction I got this error on Chrome and MacOS 

```
fe58ee2e-7945-4255-b26f-e4a3a2deae73:177 Uncaught NotSupportedError: Failed to execute 'registerProcessor' on 'AudioWorkletGlobalScope': An AudioWorkletProcessor with name:"mapi-proc" is already registered.
    at fe58ee2e-7945-4255-b26f-e4a3a2deae73:177:1
```

src/frontend/package.json
```diff
@@ -18,6 +18,7 @@
     "@fontsource-variable/material-symbols-outlined": "5.2.34",
     "@fontsource/material-icons-outlined": "5.2.6",
     "@fontsource/opendyslexic": "5.2.5",
+    "@libreaudio/la-call": "0.1.0",
```
lebaudantoine: <3

src/frontend/src/features/rooms/livekit/processors/RnnNoiseProcessor.ts
falkTX: can confirm, fixing for next build...

falkTX: should be fixed now?

falkTX:
> I got some feedback from the OSPO team: GPL would be acceptable if we can distribute an image without your code/dependency included, and provide a build flag for self-hosters or anyone wanting to modify LaSuite Meet without being required to republish it under AGPL.

just a note: GPL and AGPL are quite different here. the AGPL would apply to deployments in servers but GPL is mostly about locally-ran programs we have access to (in super simplified terms).

camillem:
>  GPL is mostly about locally-ran programs we have access to

"Modern" javascript is combined and distributed in non source form via the browser, so it's not as clear-cut as that

lebaudantoine:
pulling the branch, trying to run the frontend development image, I got the error : 

```
> meet@1.15.0 dev
> panda codegen && vite

✔️ `src/styled-system/css`: the css function to author styles
✔️ `src/styled-system/tokens`: the css variables and js function to query your tokens
✔️ `src/styled-system/patterns`: functions to implement and apply common layout patterns
✔️ `src/styled-system/jsx`: styled jsx elements for react


1:36:17 PM [vite] (client) Re-optimizing dependencies because lockfile has changed

  VITE v7.3.2  ready in 1068 ms

  ➜  Local:   http://localhost:3000/
  ➜  press h + enter to show help
✘ [ERROR] Cannot assign to "previouslyLoaded" because it is a constant

    node_modules/@libreaudio/la-call/dist/index.js:38:16:
      38 │                 previouslyLoaded = true;
         ╵                 ~~~~~~~~~~~~~~~~                                                                                                  

  The symbol "previouslyLoaded" was declared a constant here:

    node_modules/@libreaudio/la-call/dist/index.js:10:6:
      10 │ const previouslyLoaded = false;
         ╵       ~~~~~~~~~~~~~~~~                                                                                                            

🐼 info [hrtime] Extracted in (279.03ms)
🐼 error [❌] Error: Error during dependency optimization:
                                                                                                                                             
✘ [ERROR] Cannot assign to "previouslyLoaded" because it is a constant                                                                       

    node_modules/@libreaudio/la-call/dist/index.js:38:16:
      38 │                 previouslyLoaded = true;
         ╵                 ~~~~~~~~~~~~~~~~                                                                                                  

  The symbol "previouslyLoaded" was declared a constant here:

    node_modules/@libreaudio/la-call/dist/index.js:10:6:
      10 │ const previouslyLoaded = false;
         ╵       ~~~~~~~~~~~~~~~~                                                                                                            


    at failureErrorWithLog (/Users/antoine/Developer/meet/src/frontend/node_modules/vite/node_modules/esbuild/lib/main.js:1748:15)
    at /Users/antoine/Developer/meet/src/frontend/node_modules/vite/node_modules/esbuild/lib/main.js:1207:25
    at /Users/antoine/Developer/meet/src/frontend/node_modules/vite/node_modules/esbuild/lib/main.js:1626:9
    at process.processTicksAndRejections (node:internal/process/task_queues:105:5)
    ```

File: src/frontend/src/features/rooms/livekit/processors/RnnNoiseProcessor.ts [hunk 3]
```diff
@@ -35,16 +31,14 @@ export class RnnNoiseProcessor implements AudioProcessorInterface {
...
     this.sourceNode = audioContext.createMediaStreamSource(
       new MediaStream([this.source])
     )
 
-    this.noiseSuppressionNode = new AudioWorkletNode(
-      audioContext,
-      NoiseSuppressorWorklet_Name
-    )
+    this.noiseSuppressionNode = await createWasmProcessor(audioContext, { intensity: 90 });
```
cameledev: Perhaps the other main parameters (`leveler_target`, `post_gain`...) could be passed as well ?

cameledev: I guess these parameter values can be set as env variables (or even user parameters but not sure it is worth the information clutter).

cameledev (commented):
Hi, The demo audio sounds great ! 

Do you have any insight on minimal requirements to run this ? (my understanding is RNN has a fallback, do other modules also have one) ? Could be useful if this were ever to be applied by default.

Also, (this is not the scope of this PR, but mentioning it for memory)  maybe it could be nice to have a playback functionality to test out the voice optimization.

src/frontend/src/features/rooms/livekit/processors/RnnNoiseProcessor.ts
trummerschlunk: These parameters were exposed mainly for development. Technically, they could be passed as well, but we believe to have found quite the sweet-spot (especially in combination with browser constraints which are applied before passing the audio to our wasm). These settings are complex and interact with each-other so there is more to loose than to gain.

The 'intensity' parameter only affects the voice isolation stage.
**100** suppresses all non-voice elements and can sound a bit drastic or unnatural.
**90** sounds natural but still eliminates most background noises.
**0** might be good for special cases like online music classes (where it would be counterproductive to suppress music).

We will provide more documentation about the npm package soon :)

src/frontend/src/features/rooms/components/Join.tsx
```diff
@@ -213,6 +213,14 @@ export const Join = ({
...
         const track = await createLocalAudioTrack({
           deviceId: { exact: audioDeviceId },
+          noiseSuppression: true,
+          echoCancellation: true,
+          autoGainControl: true,
+          voiceIsolation: false,
+          // Audio quality optimized for voice
+          sampleRate: 48000, // High quality sample rate
+          channelCount: 1, // Mono for voice calls (saves bandwidth)
+          sampleSize: 16 // 16-bit audio
```
lebaudantoine: @FloChehab let's make sure these settings don't introduce regression with the STT pipeline.

src/frontend/src/features/rooms/livekit/processors/RnnNoiseProcessor.ts
lebaudantoine: Passing environment variables directly to the frontend isn’t supported yet. We could expose the value through the backend configuration as a workaround, but it would be better to address this properly in a dedicated PR on our side.

lebaudantoine (approved):
I just want to give a massive thank you to @trummerschlunk and @falkTX for the amazing contribution to this project ❤️

The amount of time, effort, and care you put into your work really shows, and I’m grateful for it. 

Thank you again for your time, your skills, and your support, it truly means a lot 🙏 @camillem let's find a way to pursue our collaboration.

src/frontend/src/features/rooms/livekit/processors/RnnNoiseProcessor.ts
```diff
@@ -1,9 +1,5 @@
 import { Track, TrackProcessor, ProcessorOptions } from 'livekit-client'
-import { NoiseSuppressorWorklet_Name } from '@timephy/rnnoise-wasm'
-
-// This is an example how to get the script path using Vite, may be different when using other build tools
-// NOTE: `?worker&url` is important (`worker` to generate a working script, `url` to get its url to load it)
-import NoiseSuppressorWorklet from '@timephy/rnnoise-wasm/NoiseSuppressorWorklet?worker&url'
+import { createWasmProcessor } from '@libreaudio/la-call'
```
lebaudantoine: I think I spotted a significant bundle size regression caused by the library swap.

On `main`, `index.js` (raw) is ~1.98 MB. On this branch, it jumps to ~25.95 MB.

The previous library was loaded via Vite’s `?worker&url` syntax, which emitted the worklet as a separate asset. On `main`, you can see this as `NoiseSuppressorWorklet-*.js` (~1.9 MB), loaded lazily through `audioWorklet.addModule(url)`. The WASM therefore stayed outside the main bundle.

The new `import { createWasmProcessor } from '@libreaudio/la-call'` is a static import, so Rollup inlines the entire package, including the WASM, into the entry chunk. 

Gzip only compresses it by ~3.3× (instead of the ~4–5× we typically see for JS), which suggests the WASM is being base64-encoded into the JS bundle. That also means we lose streaming/compilation benefits that we’d get from serving a real `.wasm` asset.

**Impact:** every visitor, including users who never join a call, now downloads ~8 MB gzip before the app becomes interactive. On 4G or constrained mobile connections, that likely adds several seconds to startup time, which is a real accessibility concern for a public service app.

There’s also a build-time impact: CI currently OOMs with the default 4 GB Node heap because Rollup has to render this chunk and its ~70 MB sourcemap in a single pass.

lebaudantoine: https://github.com/jitsi/rnnoise-wasm/blob/master/README.md

falkTX: I did most of the code changes here and I am going to be quite direct here - I don't know how to do the changes you are requesting.
Even when looking at the previous project readme, I still don't quite understand it.

My background is not with web technologies, specially in terms of node and modules, I am more someone that deals with C, C++ and related languages that run directly on a machine.

When initially setting up the wasm so it would integrate the loading of the files internally there were many errors related to packaging (it doesnt help that these tools keep changing all the time, the "vite" one was giving errors related to not fully supporting local imports just yet). As a way to get things moving I put all the necessary files (worklet, wasm blob and wasm js helper) in a single JS file, to then not need to do any loading. This is not ideal, but already allows to test the feature.

The best path forward now is for you (or anyone reading this PR) to help dealing with the NPM package.
The startup script that builds the wasm and npm package is https://github.com/trummerschlunk/BigBlueBetterAudio/blob/npm/build-for-wasm.sh (only tested under Linux so far)
The final files will be stored in npm/dist.

The main javascript code that handles the audio processor is at https://github.com/trummerschlunk/BigBlueBetterAudio/blob/npm/npm/src/index.js; this is the file that needs changing so it can support loading files from somewhere else that is not directly in the source-code.

PRs are very welcome. Just make sure to target the `npm` branch for the BigBlueBetterAudio project.

lebaudantoine (approved):

PR #1019, opened 2026-02-25
♿️(frontend) add skip link component for keyboard navigation

Ovgodd:
## Purpose

Add a skip link at the top of the page so keyboard and screen reader users can bypass the header and go straight to the main content, improving accessibility.

https://github.com/user-attachments/assets/78a59956-454b-491f-b1ec-d322041309ba

## Proposal

- [x] Skip link shown only when focused (first element in tab order)
- [x] Styled like the secondary buttons (white, blue border)
- [x] Focus moves to the first H1 in the main content (or main if none)
- [x] Focus ring visible on the heading after skip link is activated
- [x] Translated for FR, EN, DE, NL

### DIFF
--- a/src/frontend/src/layout/Layout.tsx
+++ b/src/frontend/src/layout/Layout.tsx
@@ -5,6 +5,7 @@ import { layoutStore } from '@/stores/layout'
 import { useSnapshot } from 'valtio'
 import { Footer } from '@/layout/Footer'
 import { ScreenReaderAnnouncer } from '@/primitives'
+import { SkipLink, MAIN_CONTENT_ID } from './SkipLink'
 
 export type Layout = 'fullpage' | 'centered'
 
@@ -21,6 +22,7 @@ export const Layout = ({ children }: { children: ReactNode }) => {
 
   return (
     <>
+      {showHeader && <SkipLink />}
       <div
         className={css({
           display: 'flex',
@@ -35,6 +37,7 @@ export const Layout = ({ children }: { children: ReactNode }) => {
       >
         {showHeader && <Header />}
         <main
+          id={MAIN_CONTENT_ID}
           className={css({
             flexGrow: 1,
             overflow: 'auto',
--- a/src/frontend/src/layout/SkipLink.tsx
+++ b/src/frontend/src/layout/SkipLink.tsx
@@ -1,0 +1,69 @@
+import { type MouseEvent } from 'react'
+import { useTranslation } from 'react-i18next'
+import { styled } from '@/styled-system/jsx'
+
+export const MAIN_CONTENT_ID = 'main-content'
+
+// Visually hidden until focus (not sr-only). Must become visible on focus for keyboard users.
+const StyledSkipLink = styled('a', {
+  base: {
+    position: 'absolute',
+    width: '1px',
+    height: '1px',
+    margin: '-1px',
+    padding: 0,
+    overflow: 'hidden',
+    clip: 'rect(0, 0, 0, 0)',
+    whiteSpace: 'nowrap',
+    border: 0,
+    textDecoration: 'none',
+    _focusVisible: {
+      position: 'fixed',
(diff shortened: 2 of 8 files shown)


src/frontend/src/layout/Layout.tsx
```diff
@@ -21,6 +22,7 @@ export const Layout = ({ children }: { children: ReactNode }) => {
 
   return (
     <>
+      <SkipLink />
```
lebaudantoine: Should we display only when the head is displayed? It doesn't make sense to add in a meeting I guess.

src/frontend/src/layout/SkipLink.tsx
```diff
@@ -0,0 +1,71 @@
...
+  textDecoration: 'none',
+  border: '1px solid var(--colors-primary-800)',
+  borderRadius: 4,
+  outline: '2px solid var(--colors-focus-ring)',
+  outlineOffset: 2,
+}
+
+export const SkipLink = () => {
+  const { t } = useTranslation()
+  const [isFocused, setIsFocused] = useState(false)
```
lebaudantoine: could we use a pure css / style approach instead of a useState for visibility? (no JS state, no re-render on focus/blur). ? 

something like this (warning it's vibe coded and not tested) : 

```
import styled from 'styled-components'

const StyledSkipLink = styled.a`
  /* hidden by default */
  position: absolute;
  width: 1px;
  height: 1px;
  margin: -1px;
  padding: 0;
  overflow: hidden;
  clip: rect(0, 0, 0, 0);
  white-space: nowrap;
  border: 0;

  &:focus-visible {
    position: fixed;
    top: 0.5rem;
    left: 50%;
    transform: translateX(-50%);
    width: auto;
    height: auto;
    margin: 0;
    padding: 0.625rem 1rem;
    overflow: visible;
    clip: auto;
    white-space: normal;
    z-index: 9999;
    background-color: white;
    color: var(--colors-primary-800);
    font-weight: 500;
    font-size: 0.875rem;
    text-decoration: none;
    border: 1px solid var(--colors-primary-800);
    border-radius: 4px;
    outline: 2px solid var(--colors-focus-ring);
    outline-offset: 2px;
  }
`

export const SkipLink = () => {
  const { t } = useTranslation()

  return (
    <StyledSkipLink href={`#${MAIN_CONTENT_ID}`} onClick={handleClick}>
      {t('skipLink')}
    </StyledSkipLink>
  )
}
```

src/frontend/src/layout/SkipLink.tsx
```diff
@@ -0,0 +1,71 @@
...
+  return (
+    <a
+      href={`#${MAIN_CONTENT_ID}`}
+      style={isFocused ? visibleStyle : hiddenStyle}
+      onFocus={() => setIsFocused(true)}
+      onBlur={() => setIsFocused(false)}
+      onClick={handleClick}
+    >
+      {t('skipLink')}
+    </a>
```
lebaudantoine: nitpick, should we use `Link` component from React aria to be consistent and get accessibility primitives for free? 

The onPress function will do the heavy lifting of handling a `PressEvent`, which will allow you to remove `e.preventDefault()` I guess.

src/frontend/src/layout/Layout.tsx
Ovgodd: yes you're right !

src/frontend/src/layout/SkipLink.tsx
Ovgodd: Yes totally ! I like it

src/frontend/src/layout/SkipLink.tsx
Ovgodd: I tried react-aria-components’ Link, but with href="#..." + our focus logic it causes a small scroll jump and the heading focus can be overridden.  So I kept a `<a>` for a stable skip link behavior if it is ok for you ?

PR #1020, opened 2026-02-25
♻️(summary) clean up code and unify logging in preparation for testing

cameledev:
## Purpose

Refactor the summary service to better separate concerns, making components easier to isolate and test. Unify logging logic to ensure consistent behavior and reduce duplication across the service layer. These changes set up the codebase for granular testing.

## Proposal

- [x] factorize currently monolithic `process_audio_transcribe_summarize_v2`
- [x] unify logging (use `logger = get_logger` as global, rather than `self._logger`)
- [x] create new `webhook_service.py` to handle data transfer

## TODO:

- [x] Test run

### DIFF
--- a/src/summary/summary/core/celery_worker.py
+++ b/src/summary/summary/core/celery_worker.py
@@ -10,9 +10,7 @@ import openai
 import sentry_sdk
 from celery import Celery, signals
 from celery.utils.log import get_task_logger
-from requests import Session, exceptions
-from requests.adapters import HTTPAdapter
-from urllib3.util import Retry
+from requests import exceptions
 
 from summary.core.analytics import MetadataManager, get_analytics
 from summary.core.config import get_settings
@@ -30,6 +28,7 @@ from summary.core.prompt import (
     PROMPT_USER_PART,
 )
 from summary.core.transcript_formatter import TranscriptFormatter
+from summary.core.webhook_service import submit_content
 
 settings = get_settings()
 analytics = get_analytics()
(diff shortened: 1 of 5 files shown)


src/summary/summary/core/celery_worker.py
```diff
@@ -56,103 +55,17 @@ def init_sentry(**_kwargs):
...
-    """Create an HTTP session configured with retry logic."""
-    session = Session()
-    retries = Retry(
-        total=settings.webhook_max_retries,
-        backoff_factor=settings.webhook_backoff_factor,
-        status_forcelist=settings.webhook_status_forcelist,
-        allowed_methods={"POST"},
-    )
-    session.mount("https://", HTTPAdapter(max_retries=retries))
-    return session
```
cameledev, line 72: now in http_client.py

src/summary/summary/core/celery_worker.py
```diff
@@ -56,103 +55,17 @@ def init_sentry(**_kwargs):
...
-        assignees = ", ".join(action.get("assignees", [])) or "-"
-        due_date = action.get("due_date") or "-"
-        line = f"- [ ] {title} Assignée à : {assignees}, Échéance : {due_date}"
-        lines.append(line)
-    if lines:
-        return "### Prochaines étapes\n\n" + "\n".join(lines)
-    return ""
-
-
-def post_with_retries(url, data):
```
cameledev, line 93: now in http_client.py

src/summary/summary/core/celery_worker.py
```diff
@@ -56,103 +55,17 @@ def init_sentry(**_kwargs):
...
-    """
-    lines = []
-    for action in llm_output.get("actions", []):
-        title = action.get("title", "").strip()
-        assignees = ", ".join(action.get("assignees", [])) or "-"
-        due_date = action.get("due_date") or "-"
-        line = f"- [ ] {title} Assignée à : {assignees}, Échéance : {due_date}"
-        lines.append(line)
-    if lines:
-        return "### Prochaines étapes\n\n" + "\n".join(lines)
```
cameledev, line 89: moved below

cameledev:
Changelog modification is not needed, I believe.

lebaudantoine:
@coderabbitai full review

lebaudantoine:
when changelog is not needed, you can add the label `noChangeLog` which will make the CI job passes.

src/summary/summary/core/llm_service.py
```diff
@@ -3,12 +3,15 @@
...
 import openai
+from celery.utils.log import get_task_logger
 from langfuse import Langfuse
 
 from summary.core.config import get_settings
 
 settings = get_settings()
 
+logger = get_task_logger(__name__)
+
```
lebaudantoine: Using `get_task_logger` at module level in a service that's not Celery-specific introduces unnecessary coupling; this service could be used outside of a task context and shouldn't depend on Celery's logging infrastructure.

Standard `logging.getLogger(__name__)` works perfectly fine inside Celery tasks: logs will be emitted correctly through the same handlers with the same log level config. The only difference is that get_task_logger adds the task name to log records, but that's cosmetic context that belongs at the task layer, not inside a generic service.

The logger injection pattern from the previous version was awkward, I must admit and unusual for Python.

src/summary/summary/core/transcript_formatter.py
```diff
@@ -1,14 +1,15 @@
...
 
+from celery.utils.log import get_task_logger
+
 from summary.core.config import get_settings
 from summary.core.locales import LocaleStrings
 
 settings = get_settings()
 
-logger = logging.getLogger(__name__)
+logger = get_task_logger(__name__)
```
lebaudantoine: same comment.

src/summary/summary/core/file_service.py
```diff
@@ -7,13 +7,16 @@
...
 import mutagen
+from celery.utils.log import get_task_logger
 from minio import Minio
 from minio.error import MinioException, S3Error
 
 from summary.core.config import get_settings
 
 settings = get_settings()
 
+logger = get_task_logger(__name__)
```
lebaudantoine: same

src/summary/summary/core/celery_worker.py
```diff
@@ -195,23 +106,45 @@ def process_audio_transcribe_summarize_v2(
...
+    return formatter.format(
         transcription,
         room=room,
         recording_date=recording_date,
         recording_time=recording_time,
         download_link=download_link,
     )
 
+
+def submit_transcript_webhook(content, title, email, sub):
```
lebaudantoine: I'd suggest extracting a dedicated service responsible for pushing content to an external destination, rather than having submit_transcript_webhook, create_retry_session, and post_with_retries spread across the task.

This service could then be reused by the summary Celery worker, which also pushes content to docs. The motivation isn't just DRY, it's more about having a single, well-defined ownership boundary for "how we push content outward". When retry logic, error handling, or auth needs to change, there's one place to update instead of hunting across workers.

Looking further ahead, if we introduce new destinations (webhooks, another service...), having this abstraction already in place makes that a much smaller lift.

src/summary/summary/core/celery_worker.py
```diff
@@ -195,23 +106,45 @@ def process_audio_transcribe_summarize_v2(
...
-        return
+        return None
 
     metadata_manager.track_transcription_metadata(task_id, transcription)
+    return transcription
+
 
-    # For locale of context, use in decreasing priority context_language,
-    # language (of meeting), default context language
+def format_transcript(
```
lebaudantoine, line 115: I'd question whether `format_transcript` pulls its weight as a separate function; it's two lines of setup followed by a single method call, isn't reused elsewhere, and doesn't add any error handling.

Inlining it would make the task flow easier to read without losing anything. A helper function should either reduce complexity or enable reuse.

src/summary/summary/core/celery_worker.py
cameledev: the current proposal creates `http_client.py` with `post_with_retries` which is used in this function. This function just wraps it for business specific logic. Isn't this what you meant ?

src/summary/summary/core/llm_service.py
cameledev: understood. will change.

src/summary/summary/core/celery_worker.py
cameledev: I think it is cleaner to separate the functions, from the standpoint of `process_audio_transcribe_summarize_v2` which is coordinating multiple tasks. Separating tasks in functions makes that clear. I would keep the change.

lebaudantoine: disucssed IRL: let's keep it that way.

src/summary/summary/core/celery_worker.py
cameledev: https://github.com/suitenumerique/meet/pull/1020/changes/92daa3f836bd5f745195ab6c89f859bfd58a188e

I didn't see the advantage of putting this in a class so I left it as functions.

I believe this addresses your point (including logging and error handling logic in a new webhook_service). Is that correct ?

src/summary/summary/core/llm_service.py
cameledev: https://github.com/suitenumerique/meet/pull/1020/changes/e82ed2067e17c08a5febd6303b8518d180f862e8

src/summary/summary/core/celery_worker.py
lebaudantoine: Discussed IRL, let's go with your suggestion.

PR #1021, opened 2026-02-25
🩹(frontend) fix German language preference update

lebaudantoine:
German was missing from the frontend/backend language list in the sync hook, causing user preference updates to be ignored.

Add the language to ensure preference changes are properly applied.


cameledev:
sieht gut für mich aus

PR #1026, opened 2026-02-26
♿️(frontend) Improve accessibility of the IntroSlider carousel

Ovgodd:
## Purpose

Improve accessibility of the IntroSlider carousel for screen reader and keyboard users. 
Navigation buttons were announced with unclear labels (e.g. "previous", "next") and focus could vanish or be lost on the first and last slides.

## Proposal

Align the carousel with accessibility best practices and keep focus stable and visible.

- [x] Accessible labels: descriptive aria-labels including slide position ("Go to next slide (2 of 3)")
- [x] Carousel semantics: `role="region"`, `aria-roledescription="carousel"`, and `aria-label` on the container
- [x] Live region: announce current slide position for screen readers
- [x] NVDA focus fallback: CSS `:focus` rule for nav buttons when `:focus-visible` detection fails
- [x] Always focusable buttons: use `aria-disabled` instead of `disabled` to avoid focus loss on the first and last slides

### DIFF
--- a/src/frontend/package-lock.json
+++ b/src/frontend/package-lock.json
@@ -31,7 +31,7 @@
         "livekit-client": "2.17.1",
         "posthog-js": "1.342.1",
         "react": "18.3.1",
-        "react-aria-components": "1.10.1",
+        "react-aria-components": "1.14.0",
         "react-dom": "18.3.1",
         "react-i18next": "15.1.1",
         "use-sound": "5.0.0",
@@ -2129,24 +2129,24 @@
       "license": "BSD-3-Clause"
     },
     "node_modules/@react-aria/autocomplete": {
-      "version": "3.0.0-beta.5",
-      "resolved": "https://registry.npmjs.org/@react-aria/autocomplete/-/autocomplete-3.0.0-beta.5.tgz",
-      "integrity": "sha512-zYiVeKGYHStpBXS0mf51k14xkVunU/dFqxumfYXDiiyknxIDE4L1kN7XKo16nus3TkTmJtqBHJrWmzCfNkRd9g==",
+      "version": "3.0.0-rc.4",
+      "resolved": "https://registry.npmjs.org/@react-aria/autocomplete/-/autocomplete-3.0.0-rc.4.tgz",
+      "integrity": "sha512-4bMMVNaCuYDZX9HM4ZNSAImZMcL/orwhLLe818+lyzmSrvGmW9h433PZxTolb0d+FnJVfn1MDY0zEWLiyI86GA==",
       "license": "Apache-2.0",
       "dependencies": {
-        "@react-aria/combobox": "^3.12.5",
-        "@react-aria/focus": "^3.20.5",
-        "@react-aria/i18n": "^3.12.10",
-        "@react-aria/interactions": "^3.25.3",
-        "@react-aria/listbox": "^3.14.6",
-        "@react-aria/searchfield": "^3.8.6",
-        "@react-aria/textfield": "^3.17.5",
-        "@react-aria/utils": "^3.29.1",
-        "@react-stately/autocomplete": "3.0.0-beta.2",
-        "@react-stately/combobox": "^3.10.6",
-        "@react-types/autocomplete": "3.0.0-alpha.32",
-        "@react-types/button": "^3.12.2",
-        "@react-types/shared": "^3.30.0",
+        "@react-aria/combobox": "^3.14.1",
+        "@react-aria/focus": "^3.21.3",
+        "@react-aria/i18n": "^3.12.14",
+        "@react-aria/interactions": "^3.26.0",
+        "@react-aria/listbox": "^3.15.1",
+        "@react-aria/searchfield": "^3.8.10",
+        "@react-aria/textfield": "^3.18.3",
+        "@react-aria/utils": "^3.32.0",
+        "@react-stately/autocomplete": "3.0.0-beta.4",
+        "@react-stately/combobox": "^3.12.1",
+        "@react-types/autocomplete": "3.0.0-alpha.36",
+        "@react-types/button": "^3.14.1",
+        "@react-types/shared": "^3.32.1",
         "@swc/helpers": "^0.5.0"
       },
       "peerDependencies": {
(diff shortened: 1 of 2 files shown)


lebaudantoine:
@coderabbitai full review

src/frontend/src/features/home/components/IntroSlider.tsx
```diff
@@ -180,18 +212,25 @@ export const IntroSlider = () => {
             <Button
               variant="secondaryText"
               square
-              aria-label={t('previous.label')}
-              tooltip={t('previous.tooltip')}
-              onPress={() => setSlideIndex(slideIndex - 1)}
-              isDisabled={slideIndex == 0}
```
lebaudantoine, line 186: Why are actions not disabled anymore when you're on the first or the last slide?

src/frontend/src/features/home/components/IntroSlider.tsx
```diff
@@ -163,11 +164,42 @@ const SLIDES: Slide[] = [
...
+  const prevSlideIndex = slideIndex - 1
+  const nextSlideIndex = slideIndex + 1
+
+  const previousAriaLabel =
+    slideIndex > 0
+      ? t('previous.withPosition', {
+          target: prevSlideIndex + 1,
+          total: NUMBER_SLIDES,
+        })
+      : t('previous.label')
```
lebaudantoine: It doesn’t really make sense to show a label like “Display previous slide” when there isn’t actually a previous slide available.

For the sake of simplicity and consistency, could we use a single position-based label in all cases, and simply disable the button when there’s no previous slide to navigate to?

CHANGELOG.md
```diff
@@ -22,7 +22,9 @@ and this project adheres to
 - 🚚(frontend) rename "wellknown" directory to "well-known" #1009
 - 🌐(frontend) localize SR modifier labels #1010
 - ⬆️(backend) update python dependencies #1011
-- ♿️(a11y) fix focus ring on tab container components
+- ♿️(frontend) fix focus ring on tab container components
+- ♿️(frontend) Improve accessibility of the IntroSlider carousel #1026
```
lebaudantoine:
```suggestion
- ♿️(frontend) improve accessibility of the IntroSlider carousel #1026
```

src/frontend/src/styles/index.css
```diff
@@ -30,6 +30,19 @@ body,
...
+  outline: 2px solid var(--colors-focus-ring) !important;
+  outline-offset: 1px;
+}
+
+/* aria-disabled state: keep focusable but styled as disabled */
+.carousel-nav-button[aria-disabled='true'] {
+  color: var(--colors-greyscale-400);
+  cursor: default;
+}
+
```
lebaudantoine: nit: you could style in js, to keep the style closer to these buttons, either using a styled button, either, 

````
const carouselNavButton = css({
  '&:focus': {
    outline: '2px solid var(--colors-focus-ring) !important',
    outlineOffset: '1px',
  },
  '&[aria-disabled="true"]': {
    color: 'greyscale.400',
    cursor: 'default',
  },
})
....

className={carouselNavButton}

```

src/frontend/src/features/home/components/IntroSlider.tsx
Ovgodd: I use aria-disabled instead of disabled to fix focus loss when reaching the first or last slide. The buttons still act disabled (no click action, grey styling, announced as disabled), but remain focusable so the focus ring stays visible for keyboard and SR users.

lebaudantoine: Even with aria-disabled="true", the button is still interactive. A keyboard user can focus it and activate it (e.g., pressing Enter still triggers it). It looks disabled, but it isn’t functionally disabled.

https://github.com/user-attachments/assets/cf150630-6980-4989-8c13-f859e1e0cc8e

src/frontend/src/features/home/components/IntroSlider.tsx
```diff
@@ -162,12 +174,40 @@ const SLIDES: Slide[] = [
...
+    if (isFirstRenderRef.current) {
+      isFirstRenderRef.current = false
+      return
+    }
+    announce(
+      t('slidePosition', { current: slideIndex + 1, total: NUMBER_SLIDES }),
+      'polite',
+      'global'
+    )
+  }, [slideIndex, announce, t, NUMBER_SLIDES])
```
lebaudantoine: performances : 

The `useRef` to skip the first render is a sign that `useEffect` might not be the right tool here. Since the buttons are the only way to navigate, I'd call announce directly in the handlers and drop both the `useEffect` and the `useRef`. What do you think?

src/frontend/src/features/home/components/IntroSlider.tsx
Ovgodd: yes good catch, 

I think we can make a compromise: 
- use aria-disabled so buttons stay in the tab order. With isDisabled they’re skipped, which hurts keyboard and SR users who would d miss the control and its “unavailable” state. 
- add pointer-events: none, blocks mouse clicks, and we early-return in onPress so keyboard activation does nothing.

WDYT?

I am open to any suggestion

src/frontend/src/features/home/components/IntroSlider.tsx
Ovgodd: Yes I agree, this is the right thing to do

src/frontend/src/features/home/components/IntroSlider.tsx
```diff
@@ -144,6 +145,21 @@ type Slide = {
...
+  },
+  '&[aria-disabled="true"]': {
+    color: 'greyscale.400',
+    cursor: 'default',
+    pointerEvents: 'none',
+  },
+  '&[aria-disabled="true"][data-pressed]': {
+    backgroundColor: 'transparent',
+  },
+})
```
lebaudantoine, line 161:
```suggestion
const carouselNavButton = css({
  _focusVisible: {
    outline: '2px solid var(--colors-focus-ring) !important',
    outlineOffset: '1px',
  },
  _disabled: {
    color: 'greyscale.400',
    cursor: 'default',
    pointerEvents: 'none',
    _pressed: {
      backgroundColor: 'red !important',
    },
  },
})
```

PR #1030, opened 2026-02-26
✨(backend) add file upload feature & tests

FloChehab:
For the coming features we will need to store files on the meet side (for instance user backgrounds).

This commits adds a new Model to manage files, and the associated serializers & viewsets. All are tested.

This work was heavily inspired by the work done by our friends at https://github.com/suitenumerique/drive
It build on the same architecture design (upload directly to S3 but download goes through our proxy), but model is much much simplier (no folders, no file sharing, etc.).

Remaining TODOs: 

- [x] Need to tweak setup for local development & urls.
- [x] Safer filename handling
- [x] Rebase on fix noop migration PR


FloChehab:
@coderabbitai full review

src/backend/core/api/viewsets.py
```diff
@@ -854,3 +877,362 @@ def media_auth(self, request, *args, **kwargs):
...
+            range_response = s3_client.get_object(
+                Bucket=default_storage.bucket_name,
+                Key=file.file_key,
+                Range="bytes=0-2047",
+            )
+            file_head = range_response["Body"].read()
+        else:
+            file_head = s3_client.get_object(
+                Bucket=default_storage.bucket_name, Key=file.file_key
+            )["Body"].read()
```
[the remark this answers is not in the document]
FloChehab: Shouldn't happen in normal situation, I am leaving those as is.

src/backend/core/utils.py
```diff
@@ -344,3 +350,106 @@ def generate_room_slug():
...
+    # Method 1: Detect from file content (magic bytes) - most reliable
+    mimetype_from_content = mime_detector.from_buffer(file_buffer)
+
+    # If we have a filename, try extension-based detection as well
+    mimetype_from_extension = None
+    if filename:
+        # Use mimetypes module to guess from extension
+        # Use guess_file_type (Python 3.13+) instead of deprecated guess_type
+        mimetype_from_extension, _ = mimetypes.guess_file_type(filename, strict=False)
+
```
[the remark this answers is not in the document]
FloChehab: Une contre-indication à passer en minimum 3.13 ?

src/backend/core/models.py
```diff
@@ -829,3 +830,158 @@ def save(self, *args, **kwargs):
 
         self.domain = self.domain.lower().strip()
         super().save(*args, **kwargs)
+
+
+class FileUploadStateChoices(models.TextChoices):
+    """Defines the possible states of a file."""
```
lebaudantoine: We generally prefer infinitive verbs to open docstrings; however, this issue appears to be backported from the documentation. I've messaged their developer to clarify which convention they follow.

(Python convention (PEP 257) is to start docstrings with an infinitive verb)

src/backend/core/models.py
```diff
@@ -829,3 +830,158 @@ def save(self, *args, **kwargs):
...
+
+    PENDING = "pending", _("Pending")
+    # Commented for know we might use those the day we implement malware detection logic
```
lebaudantoine: nit: for now* or `Commented out for now, as we may need this when we implement the malware detection logic.`

src/backend/core/models.py
```diff
@@ -829,3 +830,158 @@ def save(self, *args, **kwargs):
...
+
+    @property
+    def key_base(self):
+        """Key base of the location where the file is stored in object storage."""
+        if not self.pk:
+            raise RuntimeError(
+                "The file instance must be saved before requesting a storage key."
+            )
+
+        return f"file/{self.pk!s}"
```
lebaudantoine: Should we add a Django setting to let users configure the root path when self-hosting the project?
This would allow them to override it in their own settings.py without touching the core code.

src/backend/core/models.py
```diff
@@ -829,3 +830,158 @@ def save(self, *args, **kwargs):
...
+        ordering = ("created_at",)
+        indexes = [
+            models.Index(fields=["creator", "-created_at"]),
+        ]
+
+    def __str__(self):
+        return str(self.title)
+
+    def save(self, *args, **kwargs):
+        """Set the upload state to pending if it's the first save and it's a file"""
```
lebaudantoine: nit: period is missing at the end of the sentence.

src/backend/core/models.py
```diff
@@ -829,3 +830,158 @@ def save(self, *args, **kwargs):
...
+    hard_deleted_at = models.DateTimeField(null=True, blank=True)
+
+    filename = models.CharField(max_length=255, null=False, blank=False)
+
+    upload_state = models.CharField(
+        max_length=25,
+        choices=FileUploadStateChoices.choices,
+        null=True,
+        blank=True,
+    )
```
lebaudantoine, line 882: Should it be nullable?
Should we instead set it to a default `pending` value, which is set in the save method?


making it non-nullable could be premature and might conflict with future malware detection logic?

src/backend/core/models.py
```diff
@@ -829,3 +830,158 @@ def save(self, *args, **kwargs):
...
+        null=False,
+        blank=False,
+    )
+    title = models.CharField(_("title"), max_length=255)
+    creator = models.ForeignKey(
+        User,
+        on_delete=models.RESTRICT,
+        related_name="files_created",
+        blank=True,
+        null=True,
```
lebaudantoine, line 872: Why is creator nullable? If a file always has an owner, we could drop it. Also, is RESTRICT the right choice here? `CASCADE` seems more appropriate, if a user is deleted, their uploaded backgrounds should be too. 

It might be great to keep nullability if we anticipate auto/system-generated files.

`RESTRICT` would make account deletion painful without prior cleanup.

src/backend/core/models.py
```diff
@@ -829,3 +830,158 @@ def save(self, *args, **kwargs):
...
+            raise RuntimeError(
+                "The file instance must be saved before requesting a storage key."
+            )
+
+        return f"file/{self.pk!s}"
+
+    @property
+    def file_key(self):
+        """Key used to store the file in object storage."""
+        return f"{self.key_base}/{self.filename}"
```
lebaudantoine: Is there any sanitization on filename before it's used to build the storage key? A crafted filename with ../../ could lead to path traversal in the bucket. Using os.path.basename(self.filename) or a dedicated sanitize function would mitigate this.

src/backend/core/api/serializers.py
```diff
@@ -309,3 +325,138 @@ def validate(self, attrs):
...
+        """Ensure the provided ID does not already exist when creating a new file."""
+        request = self.context.get("request")
+
+        # Only check this on POST (creation)
+        if request:
+            if models.File.objects.filter(id=value).exists():
+                raise serializers.ValidationError(
+                    "A file with this ID already exists. You cannot override it.",
+                    code="file_create_existing_id",
+                )
```
lebaudantoine: nit : `if request and models.File.objects.filter(id=value).exists():`

src/backend/core/tasks/file.py
```diff
@@ -0,0 +1,37 @@
+"""
```
lebaudantoine, line 1: Adding a task system introduces Celery as a new required service (broker + worker) for the backend, which is a breaking change for self-hosters. 

If we want to avoid bumping a major version, we should consider making the file upload feature optional, either behind a Django setting (FILE_UPLOAD_ENABLED) with graceful degradation when Celery isn't configured, or by providing a synchronous fallback for cleanup tasks in setups without a broker. Worth aligning on this.

src/backend/core/api/viewsets.py
```diff
@@ -77,20 +100,20 @@ class NestedGenericViewSet(viewsets.GenericViewSet):
     lookup_fields: list[str] = ["pk"]
     lookup_url_kwargs: list[str] = []
 
-    def __getattribute__(self, item):
+    def __getattribute__(self, file):
```
lebaudantoine, line 103: Should we reconsider naming this model File? Drive uses Item which is more generic and could accommodate other types beyond files in the future (links, embeds, etc.). That said, File is more explicit and self-documenting for the current use case, YAGNI applies.

src/backend/core/models.py
```diff
@@ -829,3 +830,158 @@ def save(self, *args, **kwargs):
...
+        User,
+        on_delete=models.RESTRICT,
+        related_name="files_created",
+        blank=True,
+        null=True,
+    )
+    deleted_at = models.DateTimeField(null=True, blank=True)
+    hard_deleted_at = models.DateTimeField(null=True, blank=True)
+
+    filename = models.CharField(max_length=255, null=False, blank=False)
```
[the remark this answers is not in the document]
lebaudantoine: Since title already holds the original filename for display purposes, do we really need it in the storage key? A UUID-based key (file/{uuid}) would be safer (no path traversal), simpler (no special character handling), and more private. Is filename actually serving any other purpose when storing to S3?

src/backend/core/tasks/file.py
```diff
@@ -0,0 +1,37 @@
...
+Tasks related to items.
```
lebaudantoine: nit: items > files

src/backend/core/models.py
FloChehab: I think it shouldn't be null indeed.
It's null in drive because they support folder and we don't. I'll fix that.

src/backend/core/models.py
FloChehab: I thought about that, I was concerned about the potential "hard deletion" without control (if someone upload bad stuff and everything gets cleaned with an account deletion).

src/backend/core/models.py
FloChehab: > A crafted filename with ../../ could lead to path traversal in the bucket.

I think that would be forbidden by s3 as a file path though.

But I do agree that a bit more sanitization would be welcomed (they don't seem to be doing any in drive)

src/backend/core/tasks/file.py
FloChehab: Yes I agree.

src/backend/core/api/viewsets.py
FloChehab: I guess they went for item in drive because they had to manage files and "folders". I went for more explicit to our use case. I'll let you close if agreed.

src/backend/core/models.py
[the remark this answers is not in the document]
FloChehab: By "experience", filenames can be useful for human debugging and also when downloading the file directly from s3 it's easier to just have a proper filename rather than tweaking response headers and play with content-disposition so that people have a relevant downloaded filename by default.
We could double check with the drive team to see if they have insight on that.

src/backend/core/utils.py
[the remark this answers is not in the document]
FloChehab: cc @lebaudantoine

src/backend/locale/de_DE/LC_MESSAGES/django.po
```diff
@@ -8,7 +8,7 @@ msgid ""
 msgstr ""
```
lebaudantoine, line 8: I guess both django.po and django.mo (updated) should be versioned. 
I'm not an expert in i18n and Django; documentation could help: https://docs.djangoproject.com/fr/2.2/topics/i18n/

src/backend/core/models.py
FloChehab: Summary of the discussion we had: we will stay with `file/...` for now and will consider adding a settings later if people ask for it.

src/backend/core/models.py
FloChehab: (waiting for https://github.com/suitenumerique/drive/pull/573 to be done before incorporating changes from there)

src/backend/core/api/viewsets.py
```diff
@@ -66,6 +78,17 @@
 logger = getLogger(__name__)
 
 
+FILE_FOLDER = "file"
```
lebaudantoine: let's refactor it in a Django setting, reused in the model.

src/backend/core/utils.py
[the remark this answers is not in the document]
FloChehab: Done in https://github.com/suitenumerique/meet/pull/1033

src/backend/core/api/serializers.py
```diff
@@ -309,3 +325,137 @@ def validate(self, attrs):
                 ) from e
 
         return attrs
+
+
+class ListFileSerializer(serializers.ModelSerializer):
```
lebaudantoine, line 398: wip: does it need get_abilities?

src/backend/core/api/serializers.py
```diff
@@ -309,3 +325,137 @@ def validate(self, attrs):
...
+            "size",
+            "url",
+        ]
+
+    def get_url(self, obj):
+        """Return the URL of the item."""
+        if obj.upload_state == models.FileUploadStateChoices.PENDING:
+            return None
+
+        return f"{settings.MEDIA_URL}{quote(obj.file_key)}"
```
lebaudantoine: wip: discussed on tchap, let's introduce MEDIA_BASE_URL

src/backend/core/models.py
```diff
@@ -829,3 +830,158 @@ def save(self, *args, **kwargs):
...
+    READY = "ready", _("Ready")
+
+
+class FileTypeChoices(models.TextChoices):
+    """Defines the possible types of a file."""
+
+    BACKGROUND_IMAGE = "background_image", _("Background image")
+
+
+class File(BaseModel):
```
lebaudantoine, line 857: In a dedicated PR let's add an admin view on these.

src/backend/core/api/serializers.py
```diff
@@ -309,3 +325,137 @@ def validate(self, attrs):
...
+            "hard_deleted_at",
+            "upload_state",
+            "url",
+            "policy",
+        ]
+
+    def get_fields(self):
+        """Force the id field to be writable."""
+        fields = super().get_fields()
+        fields["id"].read_only = False
```
lebaudantoine, line 483: I'd love to understand the reasoning here. For simplicity, could we let the backend handle ID generation and avoid having to validate it on our end?

src/backend/core/api/serializers.py
```diff
@@ -309,3 +325,137 @@ def validate(self, attrs):
...
+    """Serializer used to create a new file"""
+
+    title = serializers.CharField(max_length=255, required=False)
+    policy = serializers.SerializerMethodField()
+
+    class Meta:
+        model = models.File
+        fields = [*ListFileSerializer.Meta.fields, "policy"]
+        read_only_fields = [
+            "created_at",
```
lebaudantoine: If we're adding abilities, let's include it. Should we also add file-related fields like size or mimetype ? or any  file's metadata.

src/backend/core/utils.py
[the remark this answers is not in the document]
lebaudantoine: discussed IRL

src/backend/core/utils.py
```diff
@@ -344,3 +350,106 @@ def generate_room_slug():
...
+    logger.info("detect_mimetype: mimetype_from_content: %s", mimetype_from_content)
+    logger.info("detect_mimetype: mimetype_from_extension: %s", mimetype_from_extension)
```
lebaudantoine: I wonder if a logger info could be noisy in production, but we would adjust it later on.

src/backend/core/utils.py
```diff
@@ -344,3 +350,106 @@ def generate_room_slug():
         for size in sizes
     ]
     return "-".join(parts)
+
+
+def detect_mimetype(file_buffer: bytes, filename: str | None = None) -> str:
```
lebaudantoine, line 355: As on Drive, there is no exception handling on this function. Let's discuss it with @lunika.

src/backend/meet/settings.py
```diff
@@ -154,6 +158,52 @@ class Base(Configuration):
...
+    )
+
+    FILE_MIMETYPE_ALLOWED_BACKGROUND_IMAGE = values.ListValue(
+        [
+            "image/jpeg",
+            "image/png",
+        ],
+        environ_name="FILE_MIMETYPE_ALLOWED_BACKGROUND_IMAGE",
+        environ_prefix=None,
+    )
```
lebaudantoine: Instead of separate lists per file type, consider grouping them into a single dict so adding new categories later is just adding a key, without any refactoring:

```python
FILE_UPLOAD_ALLOWED = values.DictValue(
    {
        "background_image": {
            "extensions": [".jpeg", ".jpg", ".png"],
            "mimetypes": ["image/jpeg", "image/png"],
        },
    },
    environ_name="FILE_UPLOAD_ALLOWED",
    environ_prefix=None,
)
```

Adding a `document` category later would then just be:
```python
"document": {
    "extensions": [".pdf", ".docx"],
    "mimetypes": ["application/pdf", "..."],
},
```

One caveat: `DictValue` requires JSON when overriding via env vars, which can be a bit unwieldy. If that's a concern, flat `ListValue` per category is fine, but it might be worth thinking about the structure now before it grows."

src/backend/core/api/serializers.py
```diff
@@ -309,3 +325,137 @@ def validate(self, attrs):
...
+            )
+
+        return value
+
+    def validate(self, attrs):
+        """Validate extension and fill title."""
+        filename_root, ext = splitext(attrs["filename"])
+
+        if settings.RESTRICT_UPLOAD_FILE_TYPE:
+            if attrs["type"] == models.FileTypeChoices.BACKGROUND_IMAGE:
```
lebaudantoine: With the dict structure (based on my other suggestions on settings.py), if/else per type would disappear.

src/backend/core/api/serializers.py
```diff
@@ -309,3 +325,137 @@ def validate(self, attrs):
...
+                raise serializers.ValidationError(
+                    {"filename": _("This file extension is not allowed.")},
+                    code="item_create_file_extension_not_allowed",
+                )
+
+        # The title will be the filename if not provided
+        if not attrs.get("title", None):
+            attrs["title"] = filename_root
+
+        return super().validate(attrs)
```
lebaudantoine:
```suggestion
    def validate(self, attrs):
        """Validate extension and fill title."""
        
        allowed = settings.FILE_UPLOAD_ALLOWED.get(attrs["type"])
        if allowed is None:
            raise NotImplementedError("File type not supported")
        
        filename_root, ext = splitext(attrs["filename"])
        
        if settings.RESTRICT_UPLOAD_FILE_TYPE and ext.lower() not in allowed["extensions"]:
            logger.info(
                "create_item: file extension not allowed %s for filename %s",
                ext,
                attrs["filename"],
            )
            raise serializers.ValidationError(
                {"filename": _("This file extension is not allowed.")},
                code="item_create_file_extension_not_allowed",
            )

        # The title will be the filename if not provided
        attrs.setdefault("title", filename_root)
        return super().validate(attrs)
```

src/backend/core/api/serializers.py
```diff
@@ -309,3 +325,137 @@ def validate(self, attrs):
...
+            raise serializers.ValidationError(
+                "A file with this ID already exists. You cannot override it.",
+                code="file_create_existing_id",
+            )
+
+        return value
+
+    def validate(self, attrs):
+        """Validate extension and fill title."""
+        filename_root, ext = splitext(attrs["filename"])
```
lebaudantoine: let's run validation on filename? or sanitize it?

src/backend/core/utils.py
```diff
@@ -344,3 +350,106 @@ def generate_room_slug():
...
+        )
+    else:
+        s3_client = default_storage.connection.meta.client
+
+    # Generate the policy
+    policy = s3_client.generate_presigned_url(
+        ClientMethod="put_object",
+        Params={"Bucket": default_storage.bucket_name, "Key": key, "ACL": "private"},
+        ExpiresIn=settings.AWS_S3_UPLOAD_POLICY_EXPIRATION,
+    )
```
lebaudantoine, line 455: should we add a maximum size to this policy?

src/backend/core/utils.py
lebaudantoine: for security reason should we log something if the extension is not matching the mimetype? 
or if we detect a suspicious MIME type, raise a suspicious operation?

those might be implemented somewhere else, I'm halfway into the review

src/backend/core/api/serializers.py
FloChehab: I went for "only creator" can do something for now. We could add abilities later ?

src/backend/core/api/viewsets.py
FloChehab: Ok :nerd_face:

src/backend/core/api/serializers.py
FloChehab: (from drive), we could I think

src/backend/core/utils.py
FloChehab: > for security reason should we log something if the extension is not matching the mimetype?

(based on drive)

There is if the declared mimetype is not aligned with the stored one.

src/backend/meet/settings.py
FloChehab: Thanks for the headsup, let's go for DictValue

src/backend/core/api/serializers.py
FloChehab: yup

src/backend/core/utils.py
FloChehab: I'd have to double check but I think you cannot control that.
I think we can enforce the Content-Length though (which would be welcomed and doable in our usecase as we should "always" know the the size of the files we are uploading and won't be doing "streaming" upload -- to confirm).

src/backend/core/api/viewsets.py
```diff
@@ -854,3 +877,358 @@ def media_auth(self, request, *args, **kwargs):
...
+    ]
+    queryset = models.File.objects.filter(hard_deleted_at__isnull=True)
+    default_serializer_class = serializers.FileSerializer
+    serializer_classes = {
+        "list": serializers.ListFileSerializer,
+        "create": serializers.CreateFileSerializer,
+    }
+
+    def get_queryset(self):
+        """Get queryset performing all annotation and filtering on the file tree structure."""
```
lebaudantoine: nit: not a tree structure anymore.

src/backend/core/migrations/0017_alter_resource_users_alter_user_language.py
```diff
@@ -0,0 +1,24 @@
...
+        migrations.AlterField(
+            model_name='resource',
+            name='users',
+            field=models.ManyToManyField(related_name='resources', through='core.ResourceAccess', through_fields=('resource', 'user'), to=settings.AUTH_USER_MODEL),
+        ),
+        migrations.AlterField(
+            model_name='user',
+            name='language',
+            field=models.CharField(choices=[('en-us', 'English'), ('fr-fr', 'French'), ('nl-nl', 'Dutch'), ('de-de', 'German')], default='en-us', help_text='The language in which the user wants to see the interface.', max_length=10, verbose_name='language'),
+        ),
```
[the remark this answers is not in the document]
FloChehab: Will be resolved in another PR. Leaving this open to not forget to clean this commit.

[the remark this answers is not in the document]
FloChehab: cf #1035

src/backend/core/api/serializers.py
FloChehab: New version pushed in 0cea503e.
I'd like to use pydantic to validate the config in the post_setup hook so that we make sure that everything is properly configured for each type. Waiting for the go to do that.
EDIT: pydantic for later.

src/backend/core/tasks/file.py
FloChehab: Fixed in 880dcdd7 I have added the setting CELERY_ENABLED that defaults to False, and custom task decorator that gently fallback. Note that I didn't manage to properly test the bahavior with mocks, etc. The python code gets evaluated before the settings are tweaked. To test, we can toggle the settings manually and rerun tests. Maybe you have a better solution.

src/backend/core/models.py
FloChehab: Added to the PR description, closing here.

src/backend/core/utils.py
FloChehab: Downgraded to debug

src/backend/meet/settings.py
FloChehab: Updated relevantly, I have put the upload size constraint in the config to avoid having a one size fits all parameter for size limit.

src/backend/core/api/viewsets.py
FloChehab: Fixed in 59cd4642

src/backend/core/api/serializers.py
FloChehab: Fixed in 59cd4642

src/backend/locale/de_DE/LC_MESSAGES/django.po
FloChehab: Fixed

FloChehab: Though, I am not sure .mo files should be commited.

lebaudantoine: Discussed IRL : let's commit it for this PR, and then add it to the CI or the Docker image

src/backend/core/api/serializers.py
lebaudantoine: Let's stay iso with drive.

src/backend/core/api/serializers.py
lebaudantoine: abilities YAGNI
block size/mimetypes while creating.

src/backend/core/api/serializers.py
lebaudantoine: Let's stay ISO with Drive

src/backend/core/models.py
lebaudantoine: Let's stay iso with drive.

src/backend/core/utils.py
lebaudantoine: discussed IRL. : should be in the viewset.

src/backend/core/utils.py
lebaudantoine: Let's reinforce it in a dedicated PR.

src/backend/core/api/viewsets.py
```diff
@@ -854,3 +877,359 @@ def media_auth(self, request, *args, **kwargs):
...
+            **serializer.validated_data,
+            creator=self.request.user,
+        )
+        serializer.instance = obj
+
+    def perform_destroy(self, instance):
+        """Override to implement a soft delete instead of dumping the record in database."""
+        instance.soft_delete()
+
+    def list(self, request, *args, **kwargs):
```
lebaudantoine: Is it the front end's responsibility to filter soft deleted files?
Should we add an option in the API List Filter to exclude those files?

Same goes for the upload_state.

src/backend/core/api/viewsets.py
```diff
@@ -854,3 +877,359 @@ def media_auth(self, request, *args, **kwargs):
...
+        # Not yet implemented
+        # Change the file.upload_state when this will be done
+        # malware_detection.analyse_file(file.file_key, file_id=file.id)
+
+        serializer = self.get_serializer(file)
+
+        return drf_response.Response(serializer.data, status=drf_status.HTTP_200_OK)
+
+    def _complete_file_deletion(self, file):
+        """Completely delete an file."""
```
lebaudantoine: let's use infinitive

src/backend/core/api/viewsets.py
```diff
@@ -854,3 +877,359 @@ def media_auth(self, request, *args, **kwargs):
...
+        file.soft_delete()
+        file.hard_delete()
+        process_file_deletion.delay(file.id)
+
+    @decorators.action(detail=True, methods=["get"], url_path="download")
+    def download(self, request, *args, **kwargs):
+        """
+        Permalink endpoint for downloading an file's file.
```
lebaudantoine: same

src/backend/core/api/viewsets.py
```diff
@@ -854,3 +877,359 @@ def media_auth(self, request, *args, **kwargs):
...
+            f"{settings.MEDIA_BASE_URL}{settings.MEDIA_URL}{quote(file.file_key)}"
+        )
+        return drf_response.Response(
+            status=drf_status.HTTP_302_FOUND,
+            headers={"Location": redirect_url},
+        )
+
+    def _authorize_subrequest(self, request, pattern):
+        """
+        Shared method to authorize access based on the original URL of an Nginx subrequest
```
lebaudantoine: same

src/backend/core/api/viewsets.py
```diff
@@ -854,3 +877,359 @@ def media_auth(self, request, *args, **kwargs):
...
+
+        Example:
+        - Ascending: GET /api/v1.0/files/?ordering=created_at
+
+    ### Filtering:
+        - `is_creator_me=true`: Returns files created by the current user.
+        - `is_creator_me=false`: Returns files created by other users.
+
+        Example:
+        - GET /api/v1.0/files/?is_creator_me=true&is_favorite=true
```
lebaudantoine: is_favorite doesn't exist anymore

src/backend/core/api/viewsets.py
```diff
@@ -854,3 +877,359 @@ def media_auth(self, request, *args, **kwargs):
...
+        "create": serializers.CreateFileSerializer,
+    }
+
+    def get_queryset(self):
+        """Get queryset performing all annotation and filtering on the file tree structure."""
+        user = self.request.user
+        queryset = super().get_queryset().select_related("creator")
+
+        if not user.is_authenticated:
+            return queryset.none()
```
lebaudantoine, line 952: why not early return ? 

```suggestion
        if not user.is_authenticated:
            return queryset.none()
            
        queryset = super().get_queryset().select_related("creator")
```

src/backend/core/api/filters.py
```diff
@@ -0,0 +1,50 @@
...
+        """
+        user = self.request.user
+
+        if not user.is_authenticated:
+            return queryset
+
+        if value:
+            return queryset.filter(creator=user)
+
+        return queryset.exclude(creator=user)
```
lebaudantoine, line 67: is it something we really want, as there is no "share" feature yet in the app. Yet we could expect the API to always return only your files.

lebaudantoine: Actually this would result in an always empty queryset is you query is_creator_me=false, as the base queryset keeps only your files.

src/backend/pyproject.toml
```diff
@@ -33,6 +33,7 @@ dependencies = [
     "django-configurations==2.5.1",
     "django-cors-headers==4.9.0",
     "django-countries==8.2.0",
+    "django-filter==25.2",
```
lebaudantoine, line 36: I think, reading the documentation, we should add it to the installed_apps setting

src/backend/core/api/viewsets.py
```diff
@@ -854,3 +877,359 @@ def media_auth(self, request, *args, **kwargs):
...
+        )
+        serializer.instance = obj
+
+    def perform_destroy(self, instance):
+        """Override to implement a soft delete instead of dumping the record in database."""
+        instance.soft_delete()
+
+    def list(self, request, *args, **kwargs):
+        """List files with pagination and filtering."""
+        # Not calling filter_queryset. We do our own cooking.
```
lebaudantoine: But why? Let's discuss it.

src/backend/core/api/permissions.py
```diff
@@ -106,3 +108,25 @@ def has_object_permission(self, request, view, obj):
         if not request.auth or not hasattr(request.auth, "video"):
             return False
         return request.auth.video.room == str(obj.id)
+
+
+class FilePermission(IsAuthenticated):
+    """
+    Permissions applying to the file API endpoint.
```
lebaudantoine, line 115: should we explicitly say it doesn't protect the list endpoint?

src/backend/core/api/permissions.py
```diff
@@ -106,3 +108,25 @@ def has_object_permission(self, request, view, obj):
...
+    Handling sof deletions specificities
```
lebaudantoine: soft*

src/backend/core/api/permissions.py
```diff
@@ -106,3 +108,25 @@ def has_object_permission(self, request, view, obj):
...
+    """
+
+    def has_object_permission(self, request, view, obj):
+        """
+        Return a 404 on deleted files or if the user is not the owner
+        """
+        if not request.user.is_authenticated:
+            return False
```
lebaudantoine: isn't it redundant with `has_permission`?

src/backend/core/api/viewsets.py
```diff
@@ -854,3 +877,359 @@ def media_auth(self, request, *args, **kwargs):
...
+        return drf_response.Response(serializer.data)
+
+    def perform_create(self, serializer):
+        """Set the current user as creator of the newly created file."""
+
+        obj = models.File.objects.create(
+            **serializer.validated_data,
+            creator=self.request.user,
+        )
+        serializer.instance = obj
```
lebaudantoine: could we use something as `serializer.save(creator=self.request.user)` ?

src/backend/core/api/viewsets.py
```diff
@@ -854,3 +877,359 @@ def media_auth(self, request, *args, **kwargs):
...
+        if request.user != file.creator:
+            logger.debug(
+                "User '%s' lacks permission for file '%s'", request.user.id, pk
+            )
+            raise drf_exceptions.PermissionDenied()
+
+        logger.debug(
+            "Subrequest authorization successful. Extracted parameters: %s", url_params
+        )
+        return url_params, request.user.id, file
```
lebaudantoine, line 1158: `_authorize_subrequest` bypasses `get_object()` entirely, querying directly with `models.File.objects.all() `and doing a manual ownership check. It never goes through `FilePermission`, so soft-deleted and hard-deleted files remain accessible via `media_auth`. That's a gap with the download endpoint.

Also we don't check upload state, but that might be a product requirement.

On Drive, the `media_auth` ability is set to False when a file is deleted. By removing the abilities check, we likely removed that safeguard as well. We should reintroduce get_abilities on the File model to better align with their implementation and ensure deleted files cannot be accessed.

src/backend/core/api/viewsets.py
```diff
@@ -854,3 +877,359 @@ def media_auth(self, request, *args, **kwargs):
...
+
+        Returns:
+        - A dictionary of URL parameters if the request is authorized.
+        Raises:
+        - PermissionDenied if authorization fails.
+        """
+        # Extract the original URL from the request header
+        original_url = request.META.get("HTTP_X_ORIGINAL_URL")
+        if not original_url:
+            logger.debug("Missing HTTP_X_ORIGINAL_URL header in subrequest")
```
lebaudantoine: should be a logger debug or considered suspicious and thus escalate to a warning level?

src/backend/core/api/viewsets.py
```diff
@@ -854,3 +877,359 @@ def media_auth(self, request, *args, **kwargs):
...
+            raise drf_exceptions.PermissionDenied()
+
+        # Fetch the file and check if the user has access
+        queryset = models.File.objects.all()
+        # No suspicious analysis implemented yet
+        # queryset = self._filter_suspicious_files(queryset, request.user)
+        try:
+            file = queryset.get(pk=pk)
+        except models.File.DoesNotExist as exc:
+            logger.debug("File with ID '%s' does not exist", pk)
```
lebaudantoine: Again, I would consider it suspicious, and maybe log a warning.

src/backend/core/api/viewsets.py
```diff
@@ -854,3 +877,359 @@ def media_auth(self, request, *args, **kwargs):
...
+                parsed_url.path,
+                pattern,
+            )
+            raise drf_exceptions.PermissionDenied()
+
+        try:
+            url_params = match.groupdict()
+        except (ValueError, AttributeError) as exc:
+            logger.debug("Failed to extract parameters from subrequest URL: %s", exc)
+            raise drf_exceptions.PermissionDenied() from exc
```
lebaudantoine, line 1131: Should we validate the domain of the original URL point to our server domain using `parsed_url.netloc`?

src/backend/core/api/viewsets.py
```diff
@@ -66,6 +78,17 @@
...
 
 
+FILE_FOLDER = settings.FILE_UPLOAD_PATH
+UUID_REGEX = (
+    r"[a-fA-F0-9]{8}-[a-fA-F0-9]{4}-[a-fA-F0-9]{4}-[a-fA-F0-9]{4}-[a-fA-F0-9]{12}"
+)
+FILE_EXT_REGEX = '[^.\\/:*?&"<>|\r\n]+'
+MEDIA_STORAGE_URL_PATTERN = re.compile(
+    f"{settings.MEDIA_URL:s}(?P<preview>preview/)?"
+    f"(?P<key>{FILE_FOLDER:s}/(?P<pk>{UUID_REGEX:s})/.*{FILE_EXT_REGEX:s})$"
```
lebaudantoine: the `*` before the file ext regex let's me rethink our discussion about path traversal.

src/backend/core/api/viewsets.py
```diff
@@ -854,3 +877,359 @@ def media_auth(self, request, *args, **kwargs):
...
+            )
+            raise drf_exceptions.PermissionDenied()
+
+        try:
+            url_params = match.groupdict()
+        except (ValueError, AttributeError) as exc:
+            logger.debug("Failed to extract parameters from subrequest URL: %s", exc)
+            raise drf_exceptions.PermissionDenied() from exc
+
+        pk = url_params.get("pk")
```
lebaudantoine, line 1133: Why do we capture a `key` in the regex?

src/backend/core/api/viewsets.py
```diff
@@ -854,3 +877,359 @@ def media_auth(self, request, *args, **kwargs):
...
+
+        # Apply ordering only now that everything is filtered and annotated
+        queryset = filters.OrderingFilter().filter_queryset(
+            self.request, queryset, self
+        )
+
+        return self.get_response_for_queryset(queryset)
+
+    @decorators.action(detail=True, methods=["post"], url_path="upload-ended")
+    def upload_ended(self, request, *args, **kwargs):
```
lebaudantoine, line 979: which part does call this endpoint?

src/backend/core/api/viewsets.py
```diff
@@ -854,3 +877,359 @@ def media_auth(self, request, *args, **kwargs):
...
+        """
+        Check the actual uploaded file and mark it as ready.
+        """
+
+        file = self.get_object()
+
+        if file.upload_state != models.FileUploadStateChoices.PENDING:
```
lebaudantoine: Could this check be encapsulated in a property?

src/backend/core/api/viewsets.py
```diff
@@ -854,3 +877,359 @@ def media_auth(self, request, *args, **kwargs):
...
+                Range="bytes=0-2047",
+            )
+            file_head = range_response["Body"].read()
+        else:
+            file_head = s3_client.get_object(
+                Bucket=default_storage.bucket_name, Key=file.file_key
+            )["Body"].read()
+
+        # Use improved MIME type detection combining magic bytes and file extension
+        logger.info("upload_ended: detecting mimetype for file: %s", file.file_key)
```
lebaudantoine, line 1030: As the file key is based on the filename this could leak personal data in the logs, shouldn't we use an id instead?

src/backend/core/api/viewsets.py
```diff
@@ -854,3 +877,359 @@ def media_auth(self, request, *args, **kwargs):
...
+                    "upload_ended: file size (%s) for file %s higher than the allowed max size",
+                    file_size,
+                    file.file_key,
+                )
+                raise drf_exceptions.ValidationError(
+                    detail="The file size is higher than the allowed max size.",
+                    code="file_size_exceeded",
+                )
+
+        if file_size > 2048:
```
lebaudantoine, line 1017: What this hardcoded value stands for?

src/backend/core/api/viewsets.py
```diff
@@ -854,3 +877,359 @@ def media_auth(self, request, *args, **kwargs):
...
+
+        if settings.FILE_UPLOAD_APPLY_RESTRICTIONS:
+            config_for_file_type = settings.FILE_UPLOAD_RESTRICTIONS[file.type]
+            allowed_file_mimetypes = config_for_file_type["allowed_mimetypes"]
+            if mimetype not in allowed_file_mimetypes:
+                self._complete_file_deletion(file)
+                logger.info(
+                    "upload_ended: mimetype not allowed %s for filename %s",
+                    mimetype,
+                    file.filename,
```
lebaudantoine: Same comment here, should e log only the mimetype and the extension?

Plus should it be a warning level?

src/backend/core/api/viewsets.py
```diff
@@ -854,3 +877,359 @@ def media_auth(self, request, *args, **kwargs):
...
+                        "Bucket": default_storage.bucket_name,
+                        "Key": file.file_key,
+                    },
+                    ContentType=mimetype,
+                    Metadata=head_response["Metadata"],
+                    MetadataDirective="REPLACE",
+                )
+            except ClientError as error:
+                # Log an exception but don't stop the action.
+                logger.exception(
```
lebaudantoine: Is it fine?

src/backend/core/api/serializers.py
```diff
@@ -309,3 +325,134 @@ def validate(self, attrs):
...
+                    attrs["filename"],
+                )
+                raise serializers.ValidationError(
+                    {"filename": _("This file extension is not allowed.")},
+                    code="item_create_file_extension_not_allowed",
+                )
+
+        # The title will be the filename if not provided
+        if not attrs.get("title", None):
+            attrs["title"] = filename_root
```
lunika, line 522: You should consider this PR to sanitize the file name https://github.com/suitenumerique/drive/pull/573

src/backend/core/api/viewsets.py
```diff
@@ -854,3 +877,359 @@ def media_auth(self, request, *args, **kwargs):
...
+    SerializerPerActionMixin,
+    mixins.CreateModelMixin,
+    mixins.DestroyModelMixin,
+    mixins.UpdateModelMixin,
+    viewsets.GenericViewSet,
+):
+    """
+    FileViewSet API.
+
+    This view set provides CRUD operations and additional actions for managing files.
```
```suggestion
    This viewset provides CRUD operations and additional actions for managing files.
```

src/backend/core/api/viewsets.py
```diff
@@ -854,3 +877,359 @@ def media_auth(self, request, *args, **kwargs):
...
+
+    def get_response_for_queryset(self, queryset, context=None):
+        """Return paginated response for the queryset if requested."""
+        context = context or self.get_serializer_context()
+        page = self.paginate_queryset(queryset)
+        if page is not None:
+            files = list(page)
+            serializer = self.get_serializer(files, many=True, context=context)
+            result = self.get_paginated_response(serializer.data)
+            return result
```
```suggestion
            serializer = self.get_serializer(page, many=True, context=context)
            return self.get_paginated_response(serializer.data)
```

There is no need to do this cast. I make it in Drive for a particular purpose you don't have here.

src/backend/core/api/viewsets.py
```diff
@@ -854,3 +877,359 @@ def media_auth(self, request, *args, **kwargs):
...
+
+        files = list(queryset)
+        serializer = self.get_serializer(files, many=True, context=context)
+        return drf_response.Response(serializer.data)
```
```suggestion

        serializer = self.get_serializer(queryset, many=True, context=context)
        return drf_response.Response(serializer.data)
```

Same here

src/backend/core/api/viewsets.py
lunika: The cooking made in Drive or Docs is to manage the tree structure and find the top level record the user can access. Here you don't have a tree structure, so maybe you can rely on the DRF list mixin ?

src/backend/core/api/viewsets.py
```diff
@@ -854,3 +877,359 @@ def media_auth(self, request, *args, **kwargs):
...
+        return drf_response.Response(serializer.data, status=drf_status.HTTP_200_OK)
+
+    def _complete_file_deletion(self, file):
+        """Completely delete an file."""
+        file.soft_delete()
+        file.hard_delete()
+        process_file_deletion.delay(file.id)
+
+    @decorators.action(detail=True, methods=["get"], url_path="download")
+    def download(self, request, *args, **kwargs):
```
lunika: If you can avoid this, remove it. This is added to Drive because other projects are saving the `url` property on their own, if the user changes the filename the url saved result in a 404.
In your case there is no "share" feature, so I think you can remove this.

src/backend/core/api/viewsets.py
```diff
@@ -854,3 +877,359 @@ def media_auth(self, request, *args, **kwargs):
...
+        if not original_url:
+            logger.debug("Missing HTTP_X_ORIGINAL_URL header in subrequest")
+            raise drf_exceptions.PermissionDenied()
+
+        parsed_url = urlparse(original_url)
+        match = pattern.search(unquote(parsed_url.path))
+
+        # If the path does not match the pattern, try to extract the parameters from the query
+        if not match:
+            match = pattern.search(unquote(parsed_url.query))
```
```suggestion
```

It is in my todo list to remove it.

src/backend/core/api/viewsets.py
lunika: To find the related record then in the database. This part is a little bit tricky. The `media_atuh` route is not a detailed action.

src/backend/core/models.py
FloChehab: Fixed in c50b833e as discussed, we will store only the extension to avoid leaking PII.

src/backend/core/models.py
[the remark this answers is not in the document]
FloChehab: https://github.com/suitenumerique/meet/pull/1030#discussion_r2877109451

src/backend/core/api/serializers.py
FloChehab: Fixed in c2b087e6

src/backend/meet/settings.py
FloChehab: Closing as we should be good

src/backend/core/api/serializers.py
FloChehab: Closing as we should be good

src/backend/core/utils.py
FloChehab: Logged in https://github.com/orgs/suitenumerique/projects/11/views/1?pane=issue&itemId=161666847

src/backend/core/api/viewsets.py
FloChehab: Yes, it's handled by the filter. I am going to backport some changes I have made on my frontend branch here.

src/backend/core/api/viewsets.py
FloChehab: Drive :see_no_evil:

src/backend/core/api/viewsets.py
FloChehab: Sorry my bad

src/backend/core/api/viewsets.py
FloChehab: I think we can update that now that we have limit the possiblities for filenames

src/backend/core/api/viewsets.py
FloChehab: It will be called from the frontend

src/backend/core/api/viewsets.py
FloChehab: Tweaked recently, closing

src/backend/core/api/viewsets.py
FloChehab: From drive, I'd guess it's a relevant value for file header size. I'll double check and add a comment.

src/backend/core/api/viewsets.py
FloChehab: From drive, I'd be ok to crash, but they must have had good reasons to do this.

src/backend/core/api/serializers.py
FloChehab: Discussed with @lebaudantoine, in the end we will not use filenames for storing on s3, only the extension to avoid leaking Personnal Information in logs, etc.
I can still hard sanitize the filename / title, but there wouldn't be hard technical reasons to do it (the backend / frontend should handle user input as not safe in any case). The main issue was with the storing system from my PoV.

src/backend/core/api/viewsets.py
FloChehab: I found this a bit weird too, removed in fixup to come.

src/backend/core/api/viewsets.py
FloChehab: Fixed in 1d36bdb5

src/backend/core/api/viewsets.py
FloChehab: Endpoint removed

src/backend/core/api/viewsets.py
FloChehab: So queryset is defined

src/backend/core/api/filters.py
FloChehab: For simplicity I have filter on creator = user.request in the viewset, but I left the is_creator_me filter to be similar to drive for this and also incentivize being explicit about it.
Open to removing it.

src/backend/pyproject.toml
FloChehab: It's not mandatory. This is a bit of a bigger change (changing the default filter backend -- would need to dig the impact on the other endpoints).

src/backend/core/api/viewsets.py
FloChehab: Cleaned @lunika, thanks for the heads up.

src/backend/core/api/permissions.py
FloChehab: yup

src/backend/core/api/viewsets.py
FloChehab: Fixed in d73b329c (was inspired by drive, I had to tweak some serializers to avoid this).

src/backend/core/api/permissions.py
FloChehab: It still partially protects the list endpoint.

FloChehab: Could you be more explicit ?

src/backend/core/api/viewsets.py
FloChehab: Comment added in c63e7a41

src/backend/core/api/viewsets.py
FloChehab: I guess if we return a permission denied after that, this should be a warning.

src/backend/core/api/viewsets.py
FloChehab: cf 008df0b6

src/backend/core/api/viewsets.py
FloChehab: I am not sure about that, I think it can be easily "faked" in all cases.

src/backend/core/api/viewsets.py
FloChehab: Fixed in 7e72c0f8

src/backend/core/api/viewsets.py
FloChehab: Fixed in ef56ef3d

src/backend/core/api/viewsets.py
lebaudantoine: agree

src/backend/core/api/viewsets.py
lebaudantoine: so we rely on a frontend's call to pass the file from pending to saved?

src/backend/core/api/viewsets.py
lebaudantoine: LGTM

src/backend/core/api/serializers.py
lebaudantoine: Ok on my side @FloChehab.

On the broader sanitization topic: we'll need to address it properly later, covering XSS, injection, and other vectors as recommended by our security auditor and the [OWASP Input Validation Cheat Sheet](https://cheatsheetseries.owasp.org/cheatsheets/Input_Validation_Cheat_Sheet.html). 

Better to handle it in a dedicated, systematic pass than to do it ad-hoc now.

or https://cheatsheetseries.owasp.org/cheatsheets/Django_REST_Framework_Cheat_Sheet.html#api72019-security-misconfiguration

> DO validate, filter, and sanitize all client-provided data, or other data coming from integrated systems.

src/backend/core/api/viewsets.py
lebaudantoine: I’m curious to understand how it behaves, but I’m fine following the drive approach.

src/backend/core/api/viewsets.py
lebaudantoine: I mean, if someone tries to pass anything other than our expected server domain, that should be considered suspicious.

Once the regex passes, there’s currently no difference between https://evil.com/the-right-pattern and https://right-domain.com/the-right-pattern in terms of code execution, and in my opinion, there should be.

src/backend/core/api/viewsets.py
```diff
@@ -854,3 +878,312 @@ def media_auth(self, request, *args, **kwargs):
...
+            raise drf_exceptions.ValidationError(
+                {"file": "This action is only available for files in PENDING state."},
+                code="file_upload_state_not_pending",
+            )
+
+        s3_client = default_storage.connection.meta.client
+
+        head_response = s3_client.head_object(
+            Bucket=default_storage.bucket_name, Key=file.file_key
+        )
```
lebaudantoine, line 996: We have no certitude the object we get here will be the same we get l.1015 while calling get_object.

FloChehab: I'd need to check if we can upload the same file multiple times with the same s3 upload link

src/backend/core/api/permissions.py
lebaudantoine: When the list endpoint was custom, I guess, it wasn't calling this permission class, but that's not the case anymore.

src/backend/core/api/viewsets.py
FloChehab: Yes, my take is someone can anyway fake being https://right-domain.com/the-right-pattern even if there are evil.

src/backend/core/api/viewsets.py
FloChehab: Yes

src/backend/core/api/viewsets.py
FloChehab: So, I think our design here lacks a bit.
Technically, we could "approve" a file  (malware detection, etc.) and use the same signed url to change the actual content after approval.

src/backend/core/api/filters.py
FloChehab: As discussed IRL, we are leaving it as is for now.

src/backend/core/api/viewsets.py
FloChehab: Closing as we should be good.

src/backend/core/api/viewsets.py
FloChehab: As discussed IRL, some routes should only be callable from nginx. Leaving it as is for now.

FloChehab: Backlogged in https://github.com/orgs/suitenumerique/projects/11/views/1?pane=issue&itemId=161760448
closing here

src/backend/core/api/serializers.py
FloChehab: backlogged in https://github.com/orgs/suitenumerique/projects/11/views/1?pane=issue&itemId=161760931
closing here

src/backend/core/api/viewsets.py
FloChehab: Backlogged in https://github.com/orgs/suitenumerique/projects/11/views/1?pane=issue&itemId=161761607

Closing here.

src/backend/core/api/viewsets.py
FloChehab: As discussed, removed from our code, let it crash.

lebaudantoine:
Next PR/Issue to be opened / studied:

- Add malware detection.
- Allow disabling the upload endpoint via Django settings, without requiring a new release.
- Harden user input validation (e.g., filenames).
- Add a Django admin view.
- Enforce a backend limit on the number of files a user can upload.

FloChehab:
Done as part of https://github.com/suitenumerique/meet/pull/1067 :

>     * Allow disabling the upload endpoint via Django settings, without requiring a new release.
> 
>     * Enforce a backend limit on the number of files a user can upload.

PR #1051, opened 2026-03-02
♿(frontend) Align the Settings dialog heading level

Ovgodd:
/////////////

### DIFF
--- a/src/frontend/src/features/settings/components/SettingsDialogExtended.tsx
+++ b/src/frontend/src/features/settings/components/SettingsDialogExtended.tsx
@@ -21,7 +21,6 @@ import { AudioTab } from './tabs/AudioTab'
 import { VideoTab } from './tabs/VideoTab'
 import { TranscriptionTab } from './tabs/TranscriptionTab'
 import { ShortcutTab } from './tabs/ShortcutTab'
-import { useRef } from 'react'
 import { useMediaQuery } from '@/features/rooms/livekit/hooks/useMediaQuery'
 import { SettingsDialogExtendedKey } from '@/features/settings/type'
 import { useIsAdminOrOwner } from '@/features/rooms/livekit/hooks/useIsAdminOrOwner'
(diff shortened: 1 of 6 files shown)


src/frontend/src/features/settings/components/SettingsDialogExtended.tsx
```diff
@@ -68,6 +68,27 @@ export const SettingsDialogExtended = (props: SettingsDialogExtended) => {
...
+        focusSelectedTab()
+        fallbackId = setTimeout(focusSelectedTab, 50)
+      })
+    })
+    return () => {
+      cancelAnimationFrame(id)
+      if (fallbackId != null) clearTimeout(fallbackId)
+    }
+  }, [props.isOpen])
+
```
lebaudantoine: Using a useEffect here makes focus management implicit and timing-dependent — hence the double rAF + setTimeout(50) to beat the dialog's own lifecycle.


A few thoughts:
- Could Dialog's onOpenChange be a more reliable hook point than racing against render timing?
- Have you checked if React Aria's Tabs exposes any built-in API for this? Fighting the component's internal focus management is a red flag.
- Is autofocusing the tab itself the right UX from an accessibility standpoint? Screen readers expect focus to land on the dialog itself (or its title) first, not jump straight to a child element.
- When opening on a specific tab (e.g. shortcuts), should we focus the first interactive element inside the tab panel rather than the tab trigger itself? The user is already being directed to that tab — landing focus inside the content is likely more useful.

Ovgodd: You're right.

After checking, I'm going to remove this `useEffect` and rely on **React Aria's built-in focus management** instead.

* `ModalOverlay/Dialog` already provide the expected behavior: **move focus into the dialog on open**, **trap focus while open**, and **restore focus** to the trigger on close.

* From a WCAG / ARIA APG perspective, the  requirement is that focus **does not remain behind the modal** and that keyboard users can navigate reliably. It's not necessary to forcibly focus the selected tab trigger on open as long as the dialog is properly announced (title/label) and the keyboard flow is consistent.

Ovgodd:
Closing this After checking, I'm going to remove this useEffect and rely on React Aria's built-in focus management instead.
So this pr is not usefull anymore ( h1 to h2 will be handle in another )

PR #1053, opened 2026-03-02
🔒️(backend) enhance API input validation to strengthen security 

lebaudantoine:


lebaudantoine:
@coderabbitai full review

src/backend/core/api/serializers.py
```diff
@@ -300,12 +325,25 @@ def validate(self, attrs):
...
+            if permission.can_subscribe_metrics:
                 raise serializers.ValidationError(
-                    {"permission": f"Invalid permission: {str(e)}"}
-                ) from e
+                    {
+                        "permission": {
+                            "can_subscribe_metrics": "This permission is not implemented."
+                        }
+                    }
+                )
```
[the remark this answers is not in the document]
lebaudantoine: SuspiciousOperation is already logged by Django, using the security logger.

src/backend/core/api/serializers.py
```diff
@@ -1,10 +1,14 @@
 """Client serializers for the Meet core app."""
 
 # pylint: disable=abstract-method,no-name-in-module
+from typing import List, Optional
```
FloChehab: I think you can directly use `list[...]` and `... | None`, no need to import those anymore.

src/backend/core/api/serializers.py
```diff
@@ -261,6 +265,28 @@ class MuteParticipantSerializer(BaseParticipantsManagementSerializer):
...
+    can_publish: Optional[bool] = None
+    can_publish_data: Optional[bool] = None
+    can_publish_sources: List[int] = Field(
+        default_factory=list
+    )  # TrackSource enum values
+    hidden: Optional[bool] = False
+    recorder: Optional[bool] = False
+    can_update_metadata: Optional[bool] = None
+    agent: Optional[bool] = False
+    can_subscribe_metrics: Optional[bool] = None
```
FloChehab: I am not familiar with the underlying behavior, but I feel it's not fully coherent to have some `Optional[bool] = None` and other `Optional[bool] = False`

I'd rather remove the Optional in the class and add defaults handled by Pydantic explicitely.

`hidden: bool = Field(default=False)` (so it's optional when pydantic evaluates the input data, but is not optional when using the class instance).
Unless None / null is an expected input value of course.

src/backend/core/api/serializers.py
```diff
@@ -300,12 +327,25 @@ def validate(self, attrs):
                 f"{', '.join(update_fields)}."
             )
 
-        if "permission" in attrs:
-            try:
-                ParticipantPermission(**attrs["permission"])
-            except ValueError as e:
+        permission = attrs.get("permission")
```
lunika: You can probably move this in a `validate_permission` method instead of the whole `validate` method ?

src/backend/core/api/serializers.py
lebaudantoine: discussed IRL. All fields should be considered optional with no default value.

src/backend/core/tests/rooms/test_api_rooms_participants_management.py
```diff
@@ -151,6 +148,199 @@ def test_update_participant_success(mock_livekit_client):
...
+        },
+    }
+
+    url = reverse("rooms-update-participant", kwargs={"pk": room.id})
+    response = client.post(url, payload, format="json")
+
+    assert response.status_code == status.HTTP_400_BAD_REQUEST
+    mock_suspicious.assert_called_once_with(
+        "Setting the following participant permissions is not allowed: agent."
+    )
```
cameledev: Unless I am mistaken, these three tests are essentially the same and could/should be parametrized.

Edit: at least the first two tests

src/backend/core/tests/rooms/test_api_rooms_participants_management.py
```diff
@@ -151,6 +148,199 @@ def test_update_participant_success(mock_livekit_client):
...
+            "agent": True,
+            "can_subscribe_metrics": False,
+        },
+    }
+
+    url = reverse("rooms-update-participant", kwargs={"pk": room.id})
+    response = client.post(url, payload, format="json")
+
+    assert response.status_code == status.HTTP_400_BAD_REQUEST
+
```
cameledev, line 265: Perhaps validate the reason for bad requests being called with string as in other tests.

cameledev (commented):
Looks good to me. Two comments on tests.

PR #1056, opened 2026-03-02
💄(frontend) truncate pinned participant name with ellipsis on overflow

Ovgodd:
## Purpose

When a participant is pinned or displayed in a tile, long names can overflow and break the layout by wrapping or extending beyond the available space.

Capture d&#39;écran 2026-03-02 161005

## Proposal

Apply text truncation with ellipsis when the participant name overflows its container.

- [x] Add overflow and ellipsis styles to ParticipantName
- [x] Adjust flex layout so ellipsis works correctly
- [x] Move layout rules into livekit.css

### DIFF
--- a/src/frontend/src/features/rooms/livekit/components/ParticipantName.tsx
+++ b/src/frontend/src/features/rooms/livekit/components/ParticipantName.tsx
@@ -1,7 +1,20 @@
+import type { CSSProperties } from 'react'
 import { Text } from '@/primitives'
 import { useTranslation } from 'react-i18next'
 import { useParticipantInfo } from '@livekit/components-react'
 import { Participant } from 'livekit-client'
+
+const participantNameStyles: CSSProperties = {
+  paddingBottom: '0.1rem',
+  overflow: 'hidden',
+  textOverflow: 'ellipsis',
+  whiteSpace: 'nowrap',
+}
+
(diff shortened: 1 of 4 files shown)


lebaudantoine:
@coderabbitai full review

lebaudantoine:
Visual regression : 

Capture d’écran 2026-03-02 à 16 33 11

Ovgodd:
> Visual regression :
> 
> Capture d’écran 2026-03-02 à 16 33 11

indeed, I fixed it by removing flex 1 on the wapper

src/frontend/src/features/rooms/livekit/components/ParticipantName.tsx
```diff
@@ -21,6 +21,9 @@ export const ParticipantName = ({
         variant="sm"
         style={{
           paddingBottom: '0.1rem',
+          overflow: 'hidden',
+          textOverflow: 'ellipsis',
+          whiteSpace: 'nowrap',
           marginLeft: '0.4rem',
         }}
       >
```
lebaudantoine: Nit (not related to your PR): for better performance and to keep the code DRY, we could extract these styles into a shared object defined at the module level, so they aren’t recreated on every render.

Ovgodd: Of course

PR #1058, opened 2026-03-03
🐛(migrations) use settings in migrations

FloChehab:
Use settings directly in migrations to avoid noop
migrations. This might have undisered side effects if we change the config over time 'invalid' data will be in the database.

It's a simple quick fix.
Keeping some migrations that are no useless to avoid changing too much the migration history for users.

Similar to https://github.com/suitenumerique/people/commit/469014ac415b25be0ceed08b31a87d2d40d743cd


lebaudantoine:
@ccoderabbitai full review

src/backend/core/migrations/0001_initial.py
```diff
@@ -96,7 +96,7 @@ class Migration(migrations.Migration):
         migrations.AddField(
             model_name='resource',
             name='users',
-            field=models.ManyToManyField(related_name='resources', through='core.ResourceAccess', to=settings.AUTH_USER_MODEL),
+            field=models.ManyToManyField(related_name='resources', through='core.ResourceAccess', through_fields=('resource', 'user'), to=settings.AUTH_USER_MODEL),
```
lebaudantoine, line 99: Was it manually added?

FloChehab: I added it myself because it gets generated like this on my side, and based on my test last week it should have been there from the beginning.

lebaudantoine:
it closes #1013

Issue #1047, opened 2026-03-01
Page titles


cyberbaloo:
## Bug Report

**Problematic behavior**
Page titles are not relevant, users cannot distinguish between the two "Visio" tabs, the first page being the Visio presentation site page, the second being the page for the meeting we are connected to.

**Expected behavior/code**
`<title>Visio</title>`

**Possible Solution**
The Visio site homepage should display "Visio - Home"

The meeting connection page must reflect the content of this page: "Visio - Join Meeting"

**Additional context/Screenshots**
<img width="489" height="58" alt="Image" src="https://github.com/user-attachments/assets/6bb3b1d6-49c5-4d81-b315-c58982132b5f" />

PR #1060, opened 2026-03-03
♿(frontend) Dynamic tab title when connected to meeting

Ovgodd:
## Purpose

Improve tab identification when participating in a video meeting. When switching tabs (e.g. to copy text into the chat), users need to easily find their meeting tab among many open tabs.

Document title example : 'uhc-vyaw-vas – 3 mars 2026 | LaSuite Visio'
follow this pattern : {room} – {date} | {app}

## Proposal

- [x] Add `RoomPageTitle` to update tab title when connected
- [x] Show room name (or slug) in tab title
- [x] Use `useTitle` from `hoofd`
- [x] Integrate in `VideoConference`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,6 +29,7 @@ and this project adheres to
 - 🐛(migrations) use settings in migrations #1058
 - 💄(frontend) truncate pinned participant name with ellipsis on overflow #1056
 - ♿(frontend) prevent focus ring clipping on invite dialog #1078
+- ♿(frontend) dynamic tab title when connected to meeting #1060
 
 ## [1.9.0] - 2026-03-02
 
--- a/src/frontend/src/features/rooms/livekit/hooks/useRoomPageTitle.ts
+++ b/src/frontend/src/features/rooms/livekit/hooks/useRoomPageTitle.ts
@@ -1,0 +1,28 @@
+import { useTitle } from 'hoofd'
+import { useRoomData } from './useRoomData'
+import { useMemo } from 'react'
+
+const APP_TITLE = import.meta.env.VITE_APP_TITLE ?? ''
+
+/**
+ * Updates the browser tab title with the room name to help users easily find
+ * the meeting tab among many open tabs. Works on both the join screen and
+ * once connected.
+ */
+export const useRoomPageTitle = () => {
+  const roomData = useRoomData()
+
+  const pageTitle = useMemo(() => {
+    if (!roomData) {
+      return APP_TITLE
+    }
+
+    const roomLabel = roomData.name || roomData.slug || ''
+
+    if (!roomLabel) return APP_TITLE
+
+    return `${APP_TITLE} - ${roomLabel}  `
+  }, [roomData])
+
+  useTitle(pageTitle)
+}
--- a/src/frontend/src/App.tsx
+++ b/src/frontend/src/App.tsx
@@ -4,7 +4,7 @@ import { Suspense } from 'react'
 import { ReactQueryDevtools } from '@tanstack/react-query-devtools'
 import { QueryClientProvider } from '@tanstack/react-query'
 import { useTranslation } from 'react-i18next'
-import { useLang } from 'hoofd'
+import { useLang, useTitle } from 'hoofd'
 import { Switch, Route } from 'wouter'
 import { I18nProvider } from 'react-aria-components'
 import { Layout } from './layout/Layout'
@@ -18,6 +18,7 @@ import { useIsSdkContext } from '@/features/sdk/hooks/useIsSdkContext'
 function App() {
   const { i18n } = useTranslation()
   useLang(i18n.language)
+  useTitle(import.meta.env.VITE_APP_TITLE ?? '')
 
   const isSDKContext = useIsSdkContext()
 
--- a/src/frontend/src/features/rooms/livekit/prefabs/VideoConference.tsx
+++ b/src/frontend/src/features/rooms/livekit/prefabs/VideoConference.tsx
@@ -30,6 +30,7 @@ import { useSidePanel } from '../hooks/useSidePanel'
 import { RecordingProvider } from '@/features/recording'
 import { ScreenShareErrorModal } from '../components/ScreenShareErrorModal'
 import { useConnectionObserver } from '../hooks/useConnectionObserver'
+import { useRoomPageTitle } from '../hooks/useRoomPageTitle'
 import { useNoiseReduction } from '../hooks/useNoiseReduction'
 import { useRegisterKeyboardShortcut } from '@/features/shortcuts/useRegisterKeyboardShortcut'
 import { useSettingsDialog } from '@/features/settings'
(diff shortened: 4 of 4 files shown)


CHANGELOG.md
```diff
@@ -11,11 +11,8 @@ and this project adheres to
 ### Changed
 
 - 🔒️(backend) enhance API input validation to strengthen security #1053
-
-### Fixed
-
-- 🐛(migrations) use settings in migrations #1058
-
+- ♿(frontend) Dynamic tab title when connected to meeting
+  #1060
```
lebaudantoine: formatting issue

src/frontend/src/features/rooms/livekit/components/RoomPageTitle.tsx
```diff
@@ -0,0 +1,40 @@
...
+const APP_TITLE = import.meta.env.VITE_APP_TITLE ?? 'Visio'
+
+/**
+ * Updates the browser tab title when the user is connected to a meeting.
+ * Uses the room name (or slug as fallback) and current date to help users
+ * easily find the meeting tab among many open tabs.
+ */
+export const RoomPageTitle = () => {
+  const room = useRoomContext()
+  const connectionState = useConnectionState(room)
```
lebaudantoine: I'm not sure why this should depend on the connection state. Even if you're not connected yet and are still on the join screen, the page title should indicate which room you're about to enter.

It would also help simplify the component’s logic by removing that conditional dependency.

src/frontend/src/features/rooms/livekit/components/RoomPageTitle.tsx
```diff
@@ -0,0 +1,40 @@
...
+import { useMemo } from 'react'
+
+const APP_TITLE = import.meta.env.VITE_APP_TITLE ?? 'Visio'
+
+/**
+ * Updates the browser tab title when the user is connected to a meeting.
+ * Uses the room name (or slug as fallback) and current date to help users
+ * easily find the meeting tab among many open tabs.
+ */
+export const RoomPageTitle = () => {
```
lebaudantoine: Nice work! Small thought, since this returns `null` and is only used once, it could be a custom hook rather than a component. 

The parent already re-renders on `roomData` changes so there'd be no extra overhead to use a hook instead, and it avoids the slightly confusing `<RoomPageTitle />` in the JSX tree for future readers.

src/frontend/src/features/rooms/livekit/components/RoomPageTitle.tsx
```diff
@@ -0,0 +1,40 @@
...
+    if (!isConnected || !roomData) {
+      return APP_TITLE
+    }
+
+    const roomLabel = roomData.name || roomData.slug || ''
+    const date = new Date().toLocaleDateString(undefined, {
+      day: 'numeric',
+      month: 'short',
+      year: 'numeric',
+    })
```
lebaudantoine: Not sure we want to expose the date, is there a specific reason for it?

src/frontend/src/features/rooms/livekit/hooks/useRoomPageTitle.ts
```diff
@@ -0,0 +1,40 @@
...
+      month: 'short',
+      year: 'numeric',
+    })
+
+    if (!roomLabel) return APP_TITLE
+
+    return `${roomLabel} – ${date} | ${APP_TITLE}`
+  }, [connectionState, roomData])
+
+  useTitle(pageTitle)
```
lebaudantoine, line 27: I guess hoofd is not restoring properly the page title on navigation.

**Steps to reproduce:**
1. Land on the home page — title is correctly `LaSuite Meet` (set via `VITE_APP_TITLE` at build time. The value of this env variable is by default LaSuiteMeet for the common digital project, and Visio for the Dinum version built docker image)
2. Create/join a meeting — title updates to `ddl-ziif-ztt – 4 Mar 2026 | LaSuite Meet` ✅
3. Navigate back — title becomes `meet.127.0.0.1.nip.io` instead of reverting to `LaSuite Meet` ❌

Demo : 

https://github.com/user-attachments/assets/537e4a8d-1978-4fc9-b925-58cb997a5d49

src/frontend/src/features/rooms/livekit/components/RoomPageTitle.tsx
lebaudantoine: could be `LaSuiteMeet - xxx-xxxx-xxx` ?

src/frontend/src/features/rooms/livekit/components/RoomPageTitle.tsx
```diff
@@ -0,0 +1,40 @@
+import { useConnectionState, useRoomContext } from '@livekit/components-react'
+import { ConnectionState } from 'livekit-client'
+import { useTitle } from 'hoofd'
+import { useRoomData } from '../hooks/useRoomData'
+import { useMemo } from 'react'
+
+const APP_TITLE = import.meta.env.VITE_APP_TITLE ?? 'Visio'
```
lebaudantoine: We shouldn’t default to Visio, since this is a shared digital project and we want to avoid hardcoding references to “Visio,” which is specific to the French instance.

Instead, the implementation should remain instance-agnostic and configurable to support all deployments consistently.

src/frontend/src/features/rooms/livekit/components/RoomPageTitle.tsx
Ovgodd: hm I agree, at first it was a clause guard, 

but yes let simplify it

src/frontend/src/features/rooms/livekit/components/RoomPageTitle.tsx
Ovgodd: > could be `LaSuiteMeet - xxx-xxxx-xxx` ?

yes of course

Ovgodd: > Not sure we want to expose the date, is there a specific reason for it?

Initially yes, I thought it would help distinguish tabs, but it doesn’t really add much. We can remove it.

src/frontend/src/features/rooms/livekit/hooks/useRoomPageTitle.ts
Ovgodd: oh nice catch

indeed hoofd had no base title in its stack, so I added a useTitle call in App.tsx. That fixes it.

src/frontend/src/features/rooms/livekit/components/RoomPageTitle.tsx
Ovgodd: Thanks I agree ! I’ve converted it to a useRoomPageTitle hook, keeps the logic in one place and avoids the extra component !

lebaudantoine:
It closes #1047

lebaudantoine:
@coderabbitai full review

CHANGELOG.md
```diff
@@ -18,6 +18,7 @@ and this project adheres to
 
 - 🐛(migrations) use settings in migrations #1058
 - 💄(frontend) truncate pinned participant name with ellipsis on overflow #1056
+- ♿(frontend) Dynamic tab title when connected to meeting #1060
```
lebaudantoine: nit: formatting, lowercase dynamic

lebaudantoine (approved):
LGTM, don't forget to rebase and squash your fixup commits

PR #1077, opened 2026-03-04
Add tests and ci testing to summary

cameledev:
## Purpose

- [x] unit tests
- [x] integration/API tests (NB: currently a lot of mocking, so would not exactly qualify as integration)
- [x] add to CI

### DIFF
--- a/src/summary/summary/core/webhook_service.py
+++ b/src/summary/summary/core/webhook_service.py
@@ -30,7 +30,7 @@ def _post_with_retries(url, data):
     session = _create_retry_session()
     session.headers.update(
         {
-            "Authorization": f"Bearer {get_settings().webhook_api_token.get_secret_value()}"
+            "Authorization": f"Bearer {get_settings().webhook_api_token.get_secret_value()}"  # noqa: E501
         }
     )
     try:


Makefile
FloChehab, line 192: Could make test also run tests for summary?

src/summary/tests/conftest.py
```diff
@@ -0,0 +1,26 @@
...
+os.environ.setdefault("LLM_API_KEY", "test-llm-key")
+os.environ.setdefault("LLM_MODEL", "test-model")
+os.environ.setdefault("WEBHOOK_API_TOKEN", "test-webhook-token")
+os.environ.setdefault("WEBHOOK_URL", "http://localhost:8002/webhook")
+os.environ.setdefault("CELERY_BROKER_URL", "memory://")
+os.environ.setdefault("CELERY_RESULT_BACKEND", "cache+memory://")
+os.environ.setdefault("POSTHOG_ENABLED", "false")
+os.environ.setdefault("SENTRY_IS_ENABLED", "false")
+os.environ.setdefault("LANGFUSE_ENABLED", "false")
+os.environ.setdefault("TASK_TRACKER_REDIS_URL", "redis://localhost:6379/0")
```
FloChehab: I find this a bit risky / odd. I have checked around a bit not sure what would be the best solution here with pydantic settings. Sincere there is a `get_settings` helper, maybe we could a have a TestSettings class that inherits from the main Settings class and does the necessary transformation.

Makefile
cameledev: yes. will change

src/summary/tests/conftest.py
cameledev: the problem is with the celery worker which requires a module-level `celery` variable with which we cannot use `get_settings`. 

One option is changing `get_settings` to something like
```
@lru_cache
def get_settings():
    """Load and cache application settings."""
    if os.environ.get("SUMMARY_ENV") == "test":
        return TestSettings()
    return Settings()
```
(with `TestSettings` now being in `summary.core.config.py`)

    
But not sure this is really better. Maybe we can leave it as is for now since it works, and change it in a future PR as it will require refactoring. What do you think ?

FloChehab: Maybe something like :

```python
@lru_cache
def get_settings():
    """Load and cache application settings."""
    if os.environ.get("SUMMARY_ENV") == "test":
        return Settings({"posthog_enabled": False, ...})
    return Settings()
```

FloChehab: I don't know if such thing is possible with pydantic settings though.
I just "fear" that we could easily run test on non tests resources.

cameledev: Made this proposal following your comment (TestSettings class, and using `get_settings` everywhere it's possible, therefore clarifying how test env is used, but fundamentally still setting test vars with `os.environ.setdefault`):

https://github.com/suitenumerique/meet/pull/1077/changes/0f575481a07d8bc42c30db0ca100b4e9ace2ff67

What do you think ?

FloChehab: lgtm thx

lebaudantoine:
@coderabbitai full review

.github/workflows/meet.yml
```diff
@@ -294,9 +294,33 @@ jobs:
...
+        uses: actions/checkout@v6
+
+      - name: Install Python
+        uses: actions/setup-python@v6
+        with:
+          python-version: "3.13"
+          cache: "pip"
+
+      - name: Install development dependencies
+        run: pip install --user .[dev]
```
lebaudantoine, line 319: why do we need the `--user` arg ?

bin/pytest
```diff
@@ -1,8 +1,9 @@
 #!/usr/bin/env bash
+# NB: this file is used locally only. In CI, it is overwritten by pytest install
 
 source "$(dirname "${BASH_SOURCE[0]}")/_config.sh"
 
 _dc_run \
     -e DJANGO_CONFIGURATION=Test \
     app-dev \
-    pytest "$@"
+    pytest "$@"
```
lebaudantoine, line 9: nit: is this change necessary?

bin/pytest-summary
```diff
@@ -0,0 +1,7 @@
+#!/usr/bin/env bash
+
```
lebaudantoine, line 2:
```suggestion
# NB: this file is used locally only. In CI, it is overwritten by pytest install
```

src/summary/summary/core/celery_worker.py
```diff
@@ -140,7 +140,7 @@ def format_transcript(
     )
 
 
-def format_actions(llm_output: dict) -> str:
+def _format_actions(llm_output: dict) -> str:
```
lebaudantoine, line 143: It is a nice improvement, but it's a behavioural change bundled into a test PR. It should ideally be a separate commit or at least explicitly noted next time.

src/summary/summary/core/file_service.py
```diff
@@ -31,23 +28,24 @@ class FileService:
...
 
         self._minio_client = Minio(
             endpoint,
-            access_key=settings.aws_s3_access_key_id,
-            secret_key=settings.aws_s3_secret_access_key.get_secret_value(),
-            secure=settings.aws_s3_secure_access,
+            access_key=get_settings().aws_s3_access_key_id,
+            secret_key=get_settings().aws_s3_secret_access_key.get_secret_value(),
+            secure=get_settings().aws_s3_secure_access,
         )
```
lebaudantoine, line 42: it feels weird to call the function several times per scope, even if it's a singleton.

src/summary/summary/core/config.py
```diff
@@ -91,9 +92,37 @@ class Settings(BaseSettings):
...
+    sentry_is_enabled: bool = False
+    langfuse_enabled: bool = False
+    task_tracker_redis_url: str = "redis://localhost:6379/0"
+
+
 @lru_cache
 def get_settings():
     """Load and cache application settings."""
+    if os.environ.get("SUMMARY_ENV") == "test":
+        return TestSettings()
```
lebaudantoine, line 125: I'm not huge fan of it, as it's a test-only flag in production code. That's test-induced damage.

src/summary/summary/core/llm_service.py
```diff
@@ -34,28 +31,28 @@ def __init__(
...
-                secret_key=settings.langfuse_secret_key.get_secret_value(),
-                public_key=settings.langfuse_public_key,
-                host=settings.langfuse_host,
-                environment=settings.langfuse_environment,
+                secret_key=get_settings().langfuse_secret_key.get_secret_value(),
+                public_key=get_settings().langfuse_public_key,
+                host=get_settings().langfuse_host,
+                environment=get_settings().langfuse_environment,
                 mask=masking_function,
             )
```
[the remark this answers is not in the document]
lebaudantoine: I agree.

cameledev:
Partially implemented in https://github.com/suitenumerique/meet/pull/1162 , skipping the integration tests included here.

PR #1087, opened 2026-03-04
💄(fix) add consistent bottom spacing to participant name label

Ovgodd:
## Purpose

On certain window sizes, the space below the participant name label disappears, and the label sits flush against the bottom edge of the video tile. A default `gap` on `.lk-participant-tile` was causing inconsistent spacing and a black strip in some layouts.

## Proposal

Override LiveKit's default `gap` on `.lk-participant-tile` so the layout stays consistent at all window sizes.

- [x] Add `gap: 0` to `.lk-participant-tile` in `livekit.css`


lebaudantoine:
Solved in #1094

PR #1104, opened 2026-03-05
Fix/reactions toolbar focus close

UGilfoyle:
## Purpose

With keyboard navigation, when the reactions toolbar is activated, it is possible to continue tabbing and the toolbar remains open. It should close when focus leaves.

Root cause: `FocusScope contain` tried to trap focus (unreliably), and `shouldCloseOnInteractOutside={() => false}` actively prevented the popover from closing.

## Proposal

- Removed `contain` from `FocusScope` — non-modal popovers should not trap focus (WAI-ARIA)
- Removed `shouldCloseOnInteractOutside` override — let React Aria's default dismiss behavior handle closing on outside interaction

- [x] TypeScript build passes
- [x] Changelog entry added

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@ and this project adheres to
 - 💄(frontend) truncate long names with ellipsis in reaction overlay #1099
 - 🐛(backend) add link to "Open" text in recording email
 - 🐛(frontend) fix "Discover the Suite" link info loss below 1024px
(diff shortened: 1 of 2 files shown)


Ovgodd:
Thank you for your contribution.
I’m closing this PR because this topic was already assigned.

Also, this PR mixes several subjects together: some of the changes and files included here are unrelated to this PR and seem to belong to other PRs.

PR #1108, opened 2026-03-06
♿(frontend) improve ui and aria labels for help article links

Ovgodd:
## Purpose

Improve accessibility for help article links. Screen readers currently announce only "Learn more", which is not descriptive enough.

## Proposal

- [x] Keep body text as plain text (link not covering the whole phrase) for clearer UX
- [x] Change visible link label to "Open documentation" (fr: "Ouvrir la documentation") for more explicit purpose
- [x] Add contextual `linkAriaLabel` (fr, en, de, nl) so screen readers announce link purpose and new window
- [x] Add `rel="noopener noreferrer"` on external links
- [x] Add `externalIcon` and `color="note"`
- [x] Fix typo: "Accèder à d'avantage" → "Accéder à davantage" (fr)
- [x] Add `paddingTop: '0.25rem'` in Tools for focus outline

### Technical notes

**rel="noopener noreferrer"** Makes external links safer by blocking access to the opener window and hiding the referrer.

### DIFF
--- a/src/frontend/src/features/recording/components/TranscriptSidePanel.tsx
+++ b/src/frontend/src/features/recording/components/TranscriptSidePanel.tsx
@@ -187,12 +187,16 @@ export const TranscriptSidePanel = () => {
         </H>
         <Text variant="body" fullWidth>
           {recordingMaxDuration
-            ? t('body', {
-                max_duration: recordingMaxDuration,
-              })
+            ? t('body', { max_duration: recordingMaxDuration })
             : t('bodyWithoutMaxDuration')}{' '}
           {data?.support?.help_article_transcript && (
-            <A href={data.support.help_article_transcript} target="_blank">
+            <A
+              href={data.support.help_article_transcript}
+              target="_blank"
+              rel="noopener noreferrer"
+              externalIcon
+              aria-label={t('linkAriaLabel')}
+            >
               {t('linkMore')}
             </A>
           )}
@@ -208,6 +212,7 @@ export const TranscriptSidePanel = () => {
                   href={data.transcription_destination}
                   target="_blank"
                   rel="noopener noreferrer"
+                  externalIcon
                 >
                   {data.transcription_destination.replace('https://', '')}
                 </A>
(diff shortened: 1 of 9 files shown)


src/frontend/src/features/recording/components/TranscriptSidePanel.tsx
lebaudantoine, line 217: this is also an external link

lebaudantoine (approved):
question : 

Should the external-link icon be underlined for visual accessibility, as it appears on official websites?

From what I can see in examples such as French government websites, the underline extends across both the text and the external-link icon, which visually indicates that they are part of the same clickable element.

Image

I believe the correct approach is to place the icon **inside the link element**, so that the **text and icon are part of the same interactive target**. This also ensures the underline applies to both elements and improves clarity for users.

Previously, I implemented the icon **outside of the link**, which was not correct from both a semantic and accessibility standpoint. Including the icon inside the `<a>` element seems to align better with common accessibility practices and the patterns used on official websites.

Would it make sense to adjust the implementation so that the icon is part of the link, allowing the underline to extend across both the label and the icon?

Ovgodd:
> question :
> 
> Should the external-link icon be underlined for visual accessibility, as it appears on official websites?
> 
> From what I can see in examples such as French government websites, the underline extends across both the text and the external-link icon, which visually indicates that they are part of the same clickable element.
> 
> Image
> I believe the correct approach is to place the icon **inside the link element**, so that the **text and icon are part of the same interactive target**. This also ensures the underline applies to both elements and improves clarity for users.
> 
> Previously, I implemented the icon **outside of the link**, which was not correct from both a semantic and accessibility standpoint. Including the icon inside the `<a>` element seems to align better with common accessibility practices and the patterns used on official websites.
> 
> Would it make sense to adjust the implementation so that the icon is part of the link, allowing the underline to extend across both the label and the icon?


As discussed with you, the external icon is already semantically inside the `<a>` element, it's rendered via a CSS `::after` pseudo-element on the link, so it is part of the same interactive target.

But `text-decoration: underline` does not extend to replaced content (`content: url(...)`), which is why the underline stops before the icon. This is a CSS limitation

Fixing this I think would affect the global `A` component,
So we can do it in a separate issue/PR if you like ?

PR #1109, opened 2026-03-06
♿(frontend) improve chat toast a11y for screen readers

Ovgodd:
## Purpose

Improve screen reader support for chat messages so users are notified when new messages arrive, whether the chat panel is open or closed.

## Proposal

- **Chat panel open**: Use `role="log"` and `aria-relevant="additions"` on the message list. `role="log"` defines a live region for sequential updates. `aria-relevant="additions"` tells assistive tech to announce only new items, not the whole list.
- **Chat panel closed**: Keep the existing toast and add a polite `announce()` via `useScreenReaderAnnounce` so the SR announces the sender and message content.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ and this project adheres to
 
 - ♿️(frontend) sync html lang attribute with i18n for screen readers #1111
 - ♿️(frontend) improve MoreLink a11y and UX on home page #1112
+- ♿(frontend) improve chat toast a11y for screen readers #1109
 
 ## [1.10.0] - 2026-03-05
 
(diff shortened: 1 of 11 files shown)


CHANGELOG.md
```diff
@@ -8,6 +8,10 @@ and this project adheres to
 
 ## [Unreleased]
 
+### Changed
+
+- ♿(frontend) improve chatt toast a11y for screen readers #1109
```
lebaudantoine: nit: chat*

src/frontend/src/locales/fr/notifications.json
```diff
@@ -9,6 +9,7 @@
   },
   "muted": "{{name}} a coupé votre micro. Aucun participant ne peut l'entendre.",
   "openChat": "Ouvrir le chat",
+  "chatMessageReceived": "{{name}} dit\u00a0: {{message}}",
```
lebaudantoine: is it working?

```suggestion
  "chatMessageReceived": "{{name}} dit : {{message}}",
```

PR #1110, opened 2026-03-07
✨(entitlements) add Entitlements system with pluggable backends

sylvinus:
This follows implementations in Drive, Messages and Calendars. This system allows Meet to gate some features for users depending on an authorization server. We provide 2 backends: a local one that always allows room creation, mimicking the current behaviour, and a DeployCenter backend, that fetches a "can_create" flag from a remote API. Future deployment contexts might add new backends, or reuse the API format of the DeployCenter one.

Currenty being tested in ANCT's instance.

### DIFF
--- a/src/backend/core/api/permissions.py
+++ b/src/backend/core/api/permissions.py
@@ -2,6 +2,8 @@
 
 from rest_framework import permissions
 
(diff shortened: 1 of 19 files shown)


sylvinus:
@coderabbitai review

PR #1120, opened 2026-03-10
🔨(python-env) migrate meet main app to UV

FloChehab:
Migrate main meet app to use UV for dependancy management.
Also optimized the backend image build sequence for faster rebuilds
when dependencies don't change.
Also removed compiled django translations files are they are done in the build process now.

Changes inspired by drive repo.


FloChehab:
> LGTM, thanks a lot! This is a huge improvement for both dependency management and i18n.
> 
> I checked it out locally and everything seems to be working fine. However, one test keeps failing: `tests/recording/event/test_notification.py::test_notify_user_by_email_success`. Running make bootstrap fails also, error message :
> 
> ```
> The po files under /app/.venv/lib/python3.13/site-packages/django/contrib/flatpages/locale/en/LC_MESSAGES are in a seemingly not writable location. mo files will not be updated/created.
> ```
> 
> I’m guessing it’s because the `.mo` files were removed from the repository. Should I adapt the test, or generate them manually?

I fixed bootstrap failing in [c2a052e](https://github.com/suitenumerique/meet/pull/1120/commits/c2a052ed54dae8b06767b9d2f8d4fa66540ba865)

Compiled translations versions shouldn't need to be updated at any time they are part of the image build process.
I am not sure why the test was failing (it was failing for me locally also, but not after the last commit). I still left the compile step in the bootstrap command to be iso with drive.

FloChehab:
NB @lebaudantoine I sticked to an explicit python version in the CI and a less strict python version in the pyproject.yaml to avoid different python version being used locally and in CI (which raised a pylint error).

PR #1121, opened 2026-03-10
🔨(docker) set file permissions to docker user

FloChehab:
Set the app related files as owned by the docker user the image was build with. This fixes a sync permission error in tilt.

Also improved tilt setup so that correct uid / gid are infered for the build args.


FloChehab:
Closing in favor of removing user related logic in tilt.

FloChehab:
Preferred https://github.com/suitenumerique/meet/pull/1123

PR #1125, opened 2026-03-10
🌐(frontend) improve German translation

ColorfulRhino:
## Purpose

As far as I understand, the German localization was mostly machine-translated so far. I went through all the text by hand and refined lots of words and phrases to raise the German language quality in the Meet app.

This PR also consistently uses the German "du" form of address, it was a mix of "du" and "Sie" before. When machine-translating future lines, the following text added to the AI prompt can help to make them consistent with the existing translation:

```
1) Use the "du" form of address throughout the translation. If a phrase allows for neutral language that doesn't compromise its meaning, prefer using that over direct address.
2) Incorporate inclusive and gender-neutral language.
3) Maintain the original meaning and context of the text.
```

## Proposal

Enhancements to the German localization include:

- [x] Refine phrases to sound more natural
- [x] Ensure consistency in words usage
- [x] Adapt phrases to be inclusive and gender-neutral
- [x] Standardize to the "du" form of address, aligning with modern practices and other open-source projects like GitLab and the Android Open Source Project

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,7 @@ and this project adheres to
 - ♿️(frontend) improve MoreLink a11y and UX on home page #1112
 - ♿(frontend) improve chat toast a11y for screen readers #1109
 - ♿(frontend) improve ui and qria labels for help article links #1108
+- 🌐(frontend) improve German translation #1125
 
 ### Fixed
 
--- a/src/frontend/src/locales/de/accessibility.json
+++ b/src/frontend/src/locales/de/accessibility.json
@@ -8,12 +8,12 @@
     },
     "scope": "Diese Erklärung zur Barrierefreiheit gilt für die Website visio.numerique.gouv.fr",
     "complianceStatus": {
-      "title": "Stand der Vereinbarkeit",
+      "title": "Stand der Gesetzeskonformität",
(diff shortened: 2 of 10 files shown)


ColorfulRhino:
> In the future, we’re planning to introduce Crowdin tool so that the community can help improve existing translations and contribute new languages. The goal is to make it easier for people to add and maintain translations collaboratively.
> 
> What do you think about that approach?

Some kind of community-sourced translation app sounds good and in my opinion is the way to go 👍
I have contributed translations to a couple of projects that were using Hosted Weblate, but only to one project that's using Crowdin. So my experience with Crowdin is limited. At least for the projects I contributed, Weblate was the more popular choice.

The UI in Crowdin is a bit more modern, but afaik it's a proprietary SaaS while Hosted Weblate is open source. I think it's worth to compare both options. I can't say for sure that one is better than the other for Meet or for La Suite apps in general.

---

Edit: One thing I'd keep in mind though, new languages should be added with care. A translation where only half is translated or badly translated is worse than just using the english version in my opinion.

lebaudantoine:
Crowdin is free for open-source projects.

ColorfulRhino:
> Crowdin is free for open-source projects.

You're righ! I still mentioned that Crowdin is proprietary since this may make a difference. I don't know the exact mission of La Suite/suitenumerique, but depending on that it may make a difference if something is FOSS or proprietary :)
In any case, in my opinion it still makes sense to compare and discuss within the La Suite team before settling on one solution, since migrations can be painful.

PR #1129, opened 2026-03-10
Fix throttling on Lobby

lebaudantoine:
Please have a look to commits' body. Fix an issue observed in production where users reach the throttling limit when they are all behind the same NAT/proxy.

### DIFF
--- a/src/backend/core/tests/rooms/test_api_rooms_lobby.py
+++ b/src/backend/core/tests/rooms/test_api_rooms_lobby.py
@@ -631,3 +631,109 @@ def test_list_waiting_participants_empty(settings):
 
     assert response.status_code == 200
     assert response.json() == {"participants": []}
+
+
+@mock.patch.object(utils, "notify_participants", return_value=None)
+@mock.patch.object(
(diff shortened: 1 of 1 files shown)


lebaudantoine:
@coderabbitai full review

src/backend/core/api/throttling.py
```diff
@@ -14,11 +16,46 @@ class MonitoredAnonRateThrottle(MonitoredThrottleMixin, AnonRateThrottle):
...
+        multiple users behind the same NAT/proxy, and is consistent with how
+        LobbyService identifies participants.
+
+        Note: as per DRF documentation, application-level throttling is not a
+        security measure against brute-force or DoS attacks. This throttle exists
+        solely to guard against accidental hammering from buggy clients.
+        """
+        participant_id = request.COOKIES.get(settings.LOBBY_COOKIE_NAME)
+
+        if not participant_id:
```
jmaupetit: Maybe we should be more explicit:

```suggestion
        participant_id = request.COOKIES.get(settings.LOBBY_COOKIE_NAME, None)

        if participant_id is None:
```

PR #1132, opened 2026-03-11
🦺(backend) optionally validate room names using a regex

lebaudantoine:
In production, room names follow a specific pattern that was only enforced on the frontend. This caused minor bounties during the bug bounty.

Move validation to the API to ensure the field is always validated server-side.

Following OWASP recommendations for DRF, treat invalid room names as suspicious input, since such requests cannot originate from the official frontend client.

### DIFF
--- a/src/backend/core/api/serializers.py
+++ b/src/backend/core/api/serializers.py
@@ -2,6 +2,7 @@
 
 # pylint: disable=abstract-method,no-name-in-module
 import logging
+import re
 from os.path import splitext
 from typing import Literal
 from urllib.parse import quote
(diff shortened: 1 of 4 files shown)


lebaudantoine:
@coderabbitai full review

lebaudantoine:
@coderabbitai full review

lebaudantoine:
@coderabbitai full review

src/backend/meet/settings.py
```diff
@@ -752,6 +752,13 @@ class Base(Configuration):
         environ_prefix=None,
     )
 
+    # Room validation
+    ROOM_NAME_REGEX = values.RegexValue(
+        None,
+        environ_name="ROOM_NAME_REGEX",
+        environ_prefix=None,
```
lunika, line 759: why not providing a valid one ?

src/backend/core/api/serializers.py
```diff
@@ -131,6 +132,21 @@ class Meta:
...
+
+        room_name_regex = settings.ROOM_NAME_REGEX
+        if room_name_regex is None:
+            return value
+
+        if not re.fullmatch(room_name_regex, value):
+            raise SuspiciousOperation("Name does not match the expected format.")
+
+        return value
+
```
jmaupetit, line 149: Would it be an interesting idea to validate the `name` field at the model level so that wherever the data comes from it is always validated the same way?

PR #1133, opened 2026-03-11
Align the Application model with the project convention by renaming active to is_active

lebaudantoine:
Align the Application model with the project convention by renaming active to is_active (with migration and updated tests), and expose the field in Django admin so administrators can deactivate rogue or compromised applications.


lebaudantoine:
@coderabbitai full review

PR #1135, opened 2026-03-11
🔐(backend) avoids revealing the inactive status of an application

lebaudantoine:
Authenticate the application secret before checking whether the
application is inactive.

This avoids revealing the inactive status of an application when
an incorrect secret is provided, preventing an authentication
state oracle and client_id enumeration.


FloChehab (approved):
Lgtm 🪂

PR #1136, opened 2026-03-11
🔧(backend) trigger webhook only for recording file uploads

lebaudantoine:
With the introduction of file background uploads, only trigger the webhook for files related to recordings.

Avoid firing the "recording saved" event for other file uploads, preventing unnecessary queries and false triggers.


lebaudantoine:
@coderabbitai full review

PR #1137, opened 2026-03-11
Support marking recordings as failed to stop manually

lebaudantoine:
This PR adds a Django admin action to let support staff manually mark recordings as "failed to stop" or update their status when automatic stopping fails, while keeping the status read-only by default. 

On the frontend, users can download recordings with this status, ensuring they can recover their recordings even if the stop process fails.


lebaudantoine:
@coderabbitai full review

PR #1141, opened 2026-03-12
✨(settings) disable file upload by default & max count

FloChehab:
Add FILE_UPLOAD_ENABLED setting (default to False, to avoid a breaking change).
Also adds a max_count_by_user sub setting to restrict the number of uploaded files per user.
And send more info to the frontend about the file related config.


env.d/development/common.dist
```diff
@@ -28,6 +28,7 @@ AWS_S3_ENDPOINT_URL=http://minio:9000
 AWS_S3_ACCESS_KEY_ID=meet
 AWS_S3_SECRET_ACCESS_KEY=password
 MEDIA_BASE_URL=http://localhost:8083
+FILE_UPLOAD_ENABLED=True
```
lebaudantoine, line 31: let's configure it for tilt.

FloChehab: Already done in another PR

PR #1144, opened 2026-03-13
🔒️(backend) avoid information exposure through exception messages

lebaudantoine:
Sanitize error handling to prevent leaking internal details when invalid or malicious requests are sent to the API.

Return generic error responses to reduce the risk of information disclosure during probing attempts.


lebaudantoine:
@coderabbitai full review

PR #1145, opened 2026-03-13
🔧(ci) explicitly set Docker Hub CI permissions to read-only

lebaudantoine:
Define read-only permissions in the workflow to clarify the expected access level and follow the principle of least privilege.


lebaudantoine:
@coderabbitai full review

PR #1153, opened 2026-03-16
feat: pass session ID to native apps via OIDC redirect URL

mmaudet:
## Summary

- When native apps (iOS, Android, Desktop) use OIDC login with a custom URL scheme as `returnTo` (e.g. `visio://auth-callback`), they cannot access the `sessionid` cookie set by Django because the system browser and the native app run in separate processes
- This adds a custom OIDC callback view that detects non-HTTPS redirect schemes and appends the session ID as a query parameter to the redirect URL
- Standard HTTPS redirects (web browser) are not affected — they continue using cookies as before

## Changes

- `src/backend/core/authentication/views.py` — New `OIDCAuthenticationCallbackView` that overrides `login_success()` to append session ID for custom-scheme redirects
- `src/backend/meet/settings.py` — Point `OIDC_CALLBACK_CLASS` to the new view

## Context

Native mobile/desktop apps (Visio Mobile) need to authenticate via OIDC. On iOS (`ASWebAuthenticationSession`) and Android (`Chrome Custom Tabs`), the system browser handles OIDC but cannot share cookies back to the app. With this change, the redirect URL becomes `visio://auth-callback?meet_sessionid=<session_key>`, allowing the app to extract the session ID directly.

## Test plan

- [ ] Web browser OIDC login still works (returnTo=https://...) — cookie-based, no change
- [ ] Native app OIDC login with returnTo=visio://auth-callback receives session ID in URL
- [ ] Session ID in URL is valid and can be used for API calls

### DIFF
--- a/src/backend/core/authentication/views.py
+++ b/src/backend/core/authentication/views.py
@@ -1,0 +1,47 @@
+"""Custom OIDC authentication views for native app support.
+
+When a native app (iOS, Android, Desktop) initiates OIDC login, it sets
+`returnTo` to a custom URL scheme (e.g. `visio://auth-callback`). After
+the OIDC flow completes, Django sets the session cookie and redirects to
+that URL. However, native apps cannot read browser cookies — they need
+the session ID passed explicitly in the redirect URL.
+
+This module overrides the OIDC callback view to append the session ID as
+a query parameter when the redirect target uses a non-HTTPS scheme.
+"""
+
+from urllib.parse import urlencode, urlparse
+
+from django.conf import settings
+from django.http import HttpResponseRedirect
+
+from lasuite.oidc_login.views import (
+    OIDCAuthenticationCallbackView as BaseCallbackView,
+)
+
+
+class OIDCAuthenticationCallbackView(BaseCallbackView):
+    """Callback view that passes session ID to native app deep links."""
+
+    def login_success(self):
+        """After successful login, append session ID for custom-scheme redirects."""
+        response = super().login_success()
+
+        # Only modify redirects to custom URL schemes (native apps).
+        # Standard HTTPS redirects (web browser) work fine with cookies.
+        if isinstance(response, HttpResponseRedirect):
+            redirect_url = response.url
+            parsed = urlparse(redirect_url)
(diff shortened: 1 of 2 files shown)


sylvinus:
Native apps for Meet will be great!

However I think there are significant security issues with this PR's implementation: no protocol whitelist, session_id in a URL. 

There are many resources online explaining best practices with OAuth2 and native apps, which I think we should follow:
https://auth0.com/blog/oauth-2-best-practices-for-native-apps/
https://curity.io/resources/learn/oauth-for-mobile-apps-best-practices/

mmaudet:
> Native apps for Meet will be great!

We work on it. Have a look here (beta) : https://github.com/mmaudet/visio-mobile

mmaudet:
Thanks @sylvinus for the review.

I've reworked the implementation in #1163 which addresses both security issues following your recommandations:

1. Protocol whitelist
Only schemes listed in NATIVE_APP_REDIRECT_SCHEMES (default: ["visio"]) trigger the mechanism. Redirects to any other scheme are left untouched (standard cookie-based behavior).

2. No more session ID in the URL
The redirect now contains a short-lived exchange code (UUID, 30s TTL, single-use) stored in Redis cache. The native app exchanges this code for the session ID via POST /api/v1.0/auth/session-exchange/, and the code is immediately deleted.

|                                | #1153                    | #1163                                                   |
|-------------------|------------------|-------------------------------------|
| Data in URL           | Raw session_key | Disposable UUID code                       |
| Scheme whitelist  | None                     | NATIVE_APP_REDIRECT_SCHEMES |
| Replay protection | None                     | Code deleted on first use                   |
| Lifetime                 | 12h (session)        | 30 seconds                                          |

8 tests added:
 - Valid code → session ID returned
 - Invalid / expired / too short / missing code → 400
 - Replay (same code used twice) → second call fails
  - Response key matches SESSION_COOKIE_NAME
  - GET method → 405

All 74 existing authentication + user tests pass with no regression. And simple test done with Visio Mobile locally.

This PR supersedes #1153.

mmaudet:
Recreating this PR from origin (suitenumerique/meet) now that I have push access.

PR #1155, opened 2026-03-17
✨(backend) force ProConnect official name in trusted rooms

### LINKED ISSUE #784: Security link : Display the ProConnect offical name of the participants ( for the option "ouvrir aux personnes de confiance"
## Feature Request

Actually, the participants can fill in the name they want, they can fill "mickey mousse" or the name of an other real personne if they want.
But sometimes, the creator of a visio wants to be sure that the participants are the real participants.

So the solution is that the participants may not have the possibility to fill in the name they want.
It will be automatically the name that ProConnect have in his system.
This fonctionnality could be activated when the administrator of the visio choose the option "ouvrir aux personnes de confiance", option in wich the participants have to log with ProConnect.
This option could be re-nammed "Lien sécurisé : Ouvrir aux personnes de confiance"

mmaudet:
## Summary

- When a room has **trusted** access level and the user is authenticated via ProConnect/OIDC, the display name is now forced to the user's official `full_name` from the OIDC claims
- Custom usernames passed as query parameters are ignored for authenticated users in trusted rooms, preventing identity fraud
- Falls back to `str(user)` (email) if `full_name` is not available
- Applied consistently in both the room serializer (direct room access) and the lobby service (entry request flow)

Closes #784

## Test plan

- [ ] Create a room with **trusted** access level
- [ ] Join as an authenticated ProConnect user with a custom username → verify the display name shows the official ProConnect name, not the custom one
- [ ] Join as an authenticated user with no `full_name` set → verify fallback to email
- [ ] Join a **public** room as an authenticated user with a custom username → verify custom username is still used (no override)
- [ ] Run backend tests: `pytest src/backend/core/tests/rooms/test_api_rooms_retrieve.py src/backend/core/tests/services/test_lobby.py`

### DIFF
--- a/src/backend/core/api/serializers.py
+++ b/src/backend/core/api/serializers.py
@@ -172,6 +172,14 @@ class RoomSerializer(serializers.ModelSerializer):
         if should_access_room:
             room_id = f"{instance.id!s}"
             username = request.query_params.get("username", None)
+
+            # Force official identity for authenticated users in trusted rooms
+            if (
+                instance.access_level == models.RoomAccessLevel.TRUSTED
+                and request.user.is_authenticated
+            ):
+                username = request.user.full_name or str(request.user)
+
             output["livekit"] = utils.generate_livekit_config(
                 room_id=room_id,
                 user=request.user,
--- a/src/backend/core/services/lobby.py
+++ b/src/backend/core/services/lobby.py
@@ -146,6 +146,13 @@ class LobbyService:
         room_id = str(room.id)
 
         if self.can_bypass_lobby(room=room, user=request.user):
+            # Force official identity for authenticated users in trusted rooms
+            if (
+                room.access_level == models.RoomAccessLevel.TRUSTED
+                and request.user.is_authenticated
+            ):
+                username = request.user.full_name or str(request.user)
+
             if participant is None:
                 participant = LobbyParticipant(
                     status=LobbyParticipantStatus.ACCEPTED,
@@ -176,6 +183,13 @@ class LobbyService:
             self.refresh_waiting_status(room.id, participant_id)
 
         elif participant.status == LobbyParticipantStatus.ACCEPTED:
+            # Force official identity for authenticated users in trusted rooms
+            if (
+                room.access_level == models.RoomAccessLevel.TRUSTED
+                and request.user.is_authenticated
+            ):
+                username = request.user.full_name or str(request.user)
+
             # wrongly named, contains access token to join a room
             livekit_config = utils.generate_livekit_config(
                 room_id=room_id,
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,10 @@ and this project adheres to
 
 ## [Unreleased]
 
+### Added
+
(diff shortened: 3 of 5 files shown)


lebaudantoine:
@coderabbitai full review

src/backend/core/api/serializers.py
```diff
@@ -172,6 +172,14 @@ def to_representation(self, instance):
         if should_access_room:
             room_id = f"{instance.id!s}"
             username = request.query_params.get("username", None)
+
+            # Force official identity for authenticated users in trusted rooms
+            if (
+                instance.access_level == models.RoomAccessLevel.TRUSTED
+                and request.user.is_authenticated
+            ):
+                username = request.user.full_name or str(request.user)
```
davd-gzl, line 181: These six lines are written three times: here, at [`lobby.py:149-154`](https://github.com/mmaudet/meet/blob/feat/784-force-proconnect-name-trusted-rooms/src/backend/core/services/lobby.py#L149-L154) and at [`lobby.py:186-191`](https://github.com/mmaudet/meet/blob/feat/784-force-proconnect-name-trusted-rooms/src/backend/core/services/lobby.py#L186-L191), and all three fall back to the address. One `resolve_username(room, user, username)` beside `generate_livekit_config` would hold the rule in one place, and keep the typed name when the account carries no name.

an account whose /users/me/ carries no name types Ana into two meetings: the one only signed-in people may enter shows meet@meet.world on the tile and in the participants list, the one anyone with the link may enter shows Ana

src/backend/core/services/lobby.py
```diff
@@ -176,6 +183,13 @@ def request_entry(
             self.refresh_waiting_status(room.id, participant_id)
 
         elif participant.status == LobbyParticipantStatus.ACCEPTED:
+            # Force official identity for authenticated users in trusted rooms
+            if (
+                room.access_level == models.RoomAccessLevel.TRUSTED
+                and request.user.is_authenticated
+            ):
+                username = request.user.full_name or str(request.user)
```
davd-gzl, line 191: [`can_bypass_lobby`](https://github.com/mmaudet/meet/blob/feat/784-force-proconnect-name-trusted-rooms/src/backend/core/services/lobby.py#L119-L122) sends everyone signed in down the branch above, so only unauthenticated visitors reach this one and the condition is false every time. A room at this level still lets them knock, since [`request_entry`](https://github.com/mmaudet/meet/blob/feat/784-force-proconnect-name-trusted-rooms/src/backend/core/api/viewsets.py#L386) is open to anyone, and a moderator can admit them. The name they typed is then the name everyone sees, so the people free to pick a name are the ones nobody can identify.

davd-gzl (commented):
In a trusted room the typed name is discarded for the account's official name, or its email when the account has none. But the prejoin still shows an editable name field, so a participant types a name, never sees it replaced, and has their email published to the room without consent. Prefill the field with the name the meeting will use, the email included, and disable it.

the prejoin of a trusted room, signed in with an account that has no name: the name field holds the typed "Ana" while the account email meet@meet.world sits in the corner chip, and the meeting shows the email

PR #1160, opened 2026-03-17
🔧(build) update openssl and libssl3t64 versions to fix build

cameledev:
Update OpenSSL and libssl3t64 package versions to resolve a [build failure](https://github.com/suitenumerique/meet/actions/runs/23191530819/job/67427263939?pr=1077). Error is due to pinned openssl and libssl versions, which are regressions (python image must have updated versions).


cameledev:
My bad, maybe we should use: `apt-get install -y --only-upgrade openssl libssl3t64`, otherwise we are upgrading all packages

PR #1163, opened 2026-03-18
🔒️(backend) secure native app OIDC login with one-time exchange codes

mmaudet:
## Summary

- Replaces #1153 with a more secure approach addressing the security concerns raised in review
- Instead of passing the session ID directly in the redirect URL, generates a **short-lived (30s), single-use exchange code** stored in Redis
- Native apps exchange this code for the session ID via a dedicated API endpoint
- Adds a **whitelist of allowed custom URL schemes** (`NATIVE_APP_REDIRECT_SCHEMES` setting, defaults to `["visio"]`)

## Changes

- `src/backend/core/authentication/views.py` — Custom OIDC callback view that generates a UUID exchange code for whitelisted custom-scheme redirects, stored in cache with 30s TTL
- `src/backend/core/authentication/api.py` — New `POST /api/v1.0/auth/session-exchange/` endpoint: validates exchange code, returns session ID, deletes code (single-use)
- `src/backend/core/urls.py` — Route registration for the exchange endpoint
- `src/backend/meet/settings.py` — `OIDC_CALLBACK_CLASS` pointed to custom view + new `NATIVE_APP_REDIRECT_SCHEMES` setting (configurable via env var)
- `src/backend/core/tests/authentication/test_session_exchange.py` — 8 tests

## Security improvements over #1153

| Concern | #1153 | This PR |
|---------|-------|---------|
| Session ID in URL | Raw `session_key` in redirect | One-time exchange code (UUID) |
| Scheme whitelist | Any non-HTTP(S) scheme accepted | Only `NATIVE_APP_REDIRECT_SCHEMES` (default: `["visio"]`) |
| Replay protection | None — session ID reusable | Code deleted on first use |
| Time-bound | Session lifetime (12h) | 30 second TTL |

## Tests

8 tests covering:
- ✅ Valid code exchange returns session ID
- ✅ Invalid/unknown code returns 400
- ✅ Expired code (TTL=0) returns 400
- ✅ Code too short rejected by validation
- ✅ Missing code field rejected
- ✅ Replay attack: second use of same code fails
- ✅ Response key matches `SESSION_COOKIE_NAME` setting
- ✅ GET method returns 405 (POST only)

All 74 existing authentication + user tests pass with no regression.

## Context

Native mobile/desktop apps (Visio Mobile) authenticate via OIDC using the system browser. The browser cannot share cookies back to the app. This change provides a secure way to pass authentication back:

1. Server redirects to `visio://auth-callback?code=<uuid>` (instead of `?session_id=<key>`)
2. App POSTs `{"code": "<uuid>"}` to `/api/v1.0/auth/session-exchange/`
3. Server returns `{"meet_sessionid": "<session_key>"}` and deletes the code

Companion changes in [visio-mobile-v2](https://github.com/anthropics/visio-mobile-v2) implement the client-side exchange with fallback to cookie extraction for backward compatibility with servers not yet updated.

## Test plan

- [ ] Web browser OIDC login still works (returnTo=https://...) — no change in behavior
- [ ] Native app OIDC login receives exchange code in callback URL
- [ ] Exchange endpoint returns valid session ID
- [ ] Exchange code cannot be reused
- [ ] Exchange code expires after 30 seconds
- [ ] Non-whitelisted schemes are not intercepted

Supersedes #1153

### DIFF
--- a/src/backend/core/authentication/api.py
+++ b/src/backend/core/authentication/api.py
@@ -1,0 +1,47 @@
+"""API endpoint for exchanging a one-time code for a session ID."""
+
+from django.conf import settings
+from django.core.cache import cache
+
+from rest_framework import serializers, status
+from rest_framework.decorators import api_view, permission_classes
+from rest_framework.permissions import AllowAny
+from rest_framework.response import Response
+
+from .views import EXCHANGE_CODE_PREFIX
+
+
+class SessionExchangeSerializer(serializers.Serializer):
+    """Validates the exchange code request."""
+
+    code = serializers.CharField(max_length=64, min_length=16)
+
+
+@api_view(["POST"])
+@permission_classes([AllowAny])
+def session_exchange(request):
+    """Exchange a one-time code for a session ID.
+
+    The code was generated during the OIDC callback and stored in cache
+    with a short TTL. This endpoint retrieves the session ID, deletes the
+    code (single-use), and returns the session ID to the native app.
+    """
+    serializer = SessionExchangeSerializer(data=request.data)
+    serializer.is_valid(raise_exception=True)
+
+    code = serializer.validated_data["code"]
+    cache_key = f"{EXCHANGE_CODE_PREFIX}{code}"
+
+    session_key = cache.get(cache_key)
+    if session_key is None:
+        return Response(
+            {"detail": "Invalid or expired code."},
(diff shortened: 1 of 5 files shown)


mmaudet:
Recreating from origin now that I have push access.

PR #741, opened 2025-10-09
Metadata agent

lebaudantoine:
/!\ not ready for merge, example of my researches for @GuittenyMartin.

The agent outputs the following data structure : 

```
{
  "events": [
    {
      "participant_id": "da8d39ff-3b1c-4e8d-9a70-c630c9871bcb",
      "type": "participant_connected",
      "timestamp": "2025-10-09T21:52:33.969950"
    },
    {
      "participant_id": "da8d39ff-3b1c-4e8d-9a70-c630c9871bcb",
      "type": "speech_start",
      "timestamp": "2025-10-09T21:52:36.039456"
    },
    {
      "participant_id": "da8d39ff-3b1c-4e8d-9a70-c630c9871bcb",
      "type": "speech_end",
      "timestamp": "2025-10-09T21:52:36.589114"
    },
    {
      "participant_id": "da8d39ff-3b1c-4e8d-9a70-c630c9871bcb",
      "type": "speech_start",
      "timestamp": "2025-10-09T21:52:38.887518"
    },
    {
      "participant_id": "da8d39ff-3b1c-4e8d-9a70-c630c9871bcb",
      "type": "speech_end",
      "timestamp": "2025-10-09T21:52:39.438141"
    },
    {
      "participant_id": "da8d39ff-3b1c-4e8d-9a70-c630c9871bcb",
      "type": "participant_disconnected",
      "timestamp": "2025-10-09T21:52:43.223255"
    }
  ],
  "participants": [
    {
      "participantId": "da8d39ff-3b1c-4e8d-9a70-c630c9871bcb",
      "name": "antoine lebaud"
    }
  ]
}
```

We could extend this data structure to capture even more event or other types of data. 
Ex : when someone mutes its mic., raises hand, send a chat message.

PR #1166, opened 2026-03-18
Metadata agent

cameledev:
WIP. Continuing work started by @lebaudantoine in https://github.com/suitenumerique/meet/pull/741/changes .

## TODO
- [x] Rebase
- [x] ajout de test minimal pour s'assurer que le start/stop du service est bien appelé
- [x] modification des conditions de lancement et arrêt du service (transcribe=true / si metadata_collector_id set alors stop)
- [ ] modification de la kubernetisation
- [x] s'assurer qu'on ne fait pas de betise avec l'object event qui est partager entre plusieurs VADAgent
- [x] regarder si on peut rendre le call à minio async

### DIFF
--- a/src/backend/core/api/viewsets.py
+++ b/src/backend/core/api/viewsets.py
@@ -343,8 +343,7 @@ class RoomViewSet(
             )
 
         if settings.METADATA_COLLECTOR_ENABLED and (
-            recording.mode == models.RecordingModeChoices.TRANSCRIPT
-            or recording.options.get("transcribe", False)
+            recording.options.get("collect_metadata", False)
         ):
             try:
                 MetadataCollectorService().start(recording)
--- a/src/backend/core/services/livekit_events.py
+++ b/src/backend/core/services/livekit_events.py
@@ -178,9 +178,7 @@ class LiveKitEventsService:
         except utils.MetadataUpdateException as e:
             logger.exception("Failed to update room's metadata: %s", e)
 
-        if settings.METADATA_COLLECTOR_ENABLED and recording.options.get(
-            "metadata_collector_dispatch_id"
-        ):
+        if recording.options.get("metadata_collector_dispatch_id", None) is not None:
             try:
                 MetadataCollectorService().stop(recording)
             except MetadataCollectorException:
--- a/src/backend/core/api/serializers.py
+++ b/src/backend/core/api/serializers.py
@@ -232,11 +232,14 @@ class RecordingOptions(BaseModel):
             When `None`, falls back to the application default.
         original_mode: The original recording mode before any override.
             Must be one of the valid RecordingModeChoices values when provided.
+        collect_metadata: Whether to collect additional metadata during recording.
+            When `None`, no metadata are collected.
 
     """
 
     language: str | None = None
     transcribe: bool | None = None
+    collect_metadata: bool | None = None
     original_mode: Literal["screen_recording", "transcript"] | None = None
 
     model_config = {"extra": "forbid"}
--- a/src/frontend/src/features/analytics/enums.ts
+++ b/src/frontend/src/features/analytics/enums.ts
@@ -5,4 +5,5 @@ export enum FeatureFlags {
   noiseReduction = 'noise-reduction',
   subtitles = 'subtitles',
   candidatePolling = 'candidate-polling',
+  metadataCollector = 'metadata-collector',
 }
--- a/src/frontend/src/features/recording/components/TranscriptSidePanel.tsx
+++ b/src/frontend/src/features/recording/components/TranscriptSidePanel.tsx
@@ -32,6 +32,7 @@ import { NoAccessView } from './NoAccessView'
 import { ControlsButton } from './ControlsButton'
 import { RowWrapper } from './RowWrapper'
 import { useMutateRecording } from '../hooks/useMutateRecording'
+import { useIsMetadataCollectorEnabled } from '../hooks/useMetadataCollectorEnabled'
 import { useSidePanel } from '@/features/rooms/livekit/hooks/useSidePanel'
 
 export const TranscriptSidePanel = () => {
@@ -58,6 +59,8 @@ export const TranscriptSidePanel = () => {
     RecordingMode.Transcript,
     FeatureFlags.Transcript
   )
+
+  const isMetadataCollectorEnabled = useIsMetadataCollectorEnabled()
 
   const roomId = useRoomId()
 
@@ -106,6 +109,7 @@ export const TranscriptSidePanel = () => {
             transcribe: true,
             original_mode: RecordingMode.Transcript,
           }),
+          collect_metadata: isMetadataCollectorEnabled,
         }
 
         await startRecording({
--- a/src/frontend/src/features/recording/hooks/useMetadataCollectorEnabled.ts
+++ b/src/frontend/src/features/recording/hooks/useMetadataCollectorEnabled.ts
@@ -1,0 +1,10 @@
+import { useFeatureFlagEnabled } from 'posthog-js/react'
+import { useIsAnalyticsEnabled } from '@/features/analytics/hooks/useIsAnalyticsEnabled'
+import { FeatureFlags } from '@/features/analytics/enums'
+
+export const useIsMetadataCollectorEnabled = () => {
+  const featureEnabled = useFeatureFlagEnabled(FeatureFlags.metadataCollector)
+  const isAnalyticsEnabled = useIsAnalyticsEnabled()
+
+  return (featureEnabled && isAnalyticsEnabled) || true
+}


src/backend/core/api/viewsets.py
```diff
@@ -336,6 +340,15 @@ def start_room_recording(self, request, pk=None):  # pylint: disable=unused-argu
                 status=drf_status.HTTP_500_INTERNAL_SERVER_ERROR,
             )
 
+        if (
+            settings.METADATA_COLLECTOR_ENABLED
+            and recording.mode == models.RecordingModeChoices.TRANSCRIPT
+        ):
+            try:
+                MetadataCollectorService().start(recording)
+            except MetadataCollectorException:
```
lebaudantoine, line 350: wip: we might maybe precise why we pass if the metadata collector start to fails

src/backend/core/api/viewsets.py
```diff
@@ -336,6 +340,15 @@ def start_room_recording(self, request, pk=None):  # pylint: disable=unused-argu
                 status=drf_status.HTTP_500_INTERNAL_SERVER_ERROR,
             )
 
+        if (
+            settings.METADATA_COLLECTOR_ENABLED
+            and recording.mode == models.RecordingModeChoices.TRANSCRIPT
```
lebaudantoine: outdated, as recordings have now options, and in the options we determine if a transcription is needed.

src/backend/core/services/livekit_events.py
```diff
@@ -174,6 +178,15 @@ def _handle_egress_ended(self, data):
         except utils.MetadataUpdateException as e:
             logger.exception("Failed to update room's metadata: %s", e)
 
+        if (
+            settings.METADATA_COLLECTOR_ENABLED
+            and recording.mode == models.RecordingModeChoices.TRANSCRIPT
```
lebaudantoine: isn't it better to check if a dispatch id was set / saved / persisted, and if it's set, to call the stop function.

src/helm/env.d/dev-keycloak/values.meet.yaml.gotmpl
```diff
@@ -282,12 +283,23 @@ agents:
...
   image:
     repository: localhost:5001/meet-agents
     pullPolicy: Always
     tag: "latest"
 
+  command:
+    - "python"
+    - "metadata_collector.py"
+    - "start"
+
```
lebaudantoine, line 284: we should add another deployment to keep the two agent in the stack (one for the subtitle, one for the metadata collector)

src/agents/metadata_collector.py
```diff
@@ -0,0 +1,383 @@
...
+        """Initialize metadata agent."""
+        self.minio_client = Minio(
+            endpoint=os.getenv("AWS_S3_ENDPOINT_URL"),
+            access_key=os.getenv("AWS_S3_ACCESS_KEY_ID"),
+            secret_key=os.getenv("AWS_S3_SECRET_ACCESS_KEY"),
+            secure=os.getenv("AWS_S3_SECURE_ACCESS", "False").lower() == "true",
+        )
+
+        if os.getenv("AWS_STORAGE_BUCKET_NAME") is not None:
+            self.bucket_name = os.getenv("AWS_STORAGE_BUCKET_NAME")
```
lebaudantoine: double call 

```suggestion
        aws_.... = os.getenv("AWS_STORAGE_BUCKET_NAME")
        if aws_....  is not None:
            self.bucket_name = aws_.... 
```

src/agents/exceptions.py
```diff
@@ -0,0 +1,5 @@
+"""Storage parsers specific exceptions."""
+
+
+class MissingConfigError(Exception):
+    """Raised when a variable is not set in configuration."""
```
lebaudantoine, line 5: should we add a message to this exception

src/agents/metadata_collector.py
```diff
@@ -0,0 +1,383 @@
...
+        stream = BytesIO(data)
+
+        try:
+            self.minio_client.put_object(
+                self.bucket_name,
+                self.output_filename,
+                stream,
+                length=len(data),
+                content_type="application/json",
+            )
```
lebaudantoine, line 203: I guess, this blocking call could be problematic and block the event loop.
synchronous save() method is called from aclose() which is an asynchronous method.

src/agents/metadata_collector.py
```diff
@@ -0,0 +1,383 @@
...
+        def on_user_state(event):
+            timestamp = datetime.now(timezone.utc)
+
+            if event.new_state == "speaking":
+                event = MetadataEvent(
+                    participant_id=self.participant_identity,
+                    type="speech_start",
+                    timestamp=timestamp,
+                )
+                self.events.append(event)
```
lebaudantoine, line 98: These operations are not thread-safe. let's discuss it.

src/backend/core/recording/services/metadata_collector.py
```diff
@@ -0,0 +1,91 @@
+"""Meeting metadata collection service."""
```
lebaudantoine, line 1: write some unit tests / and enhance the test suite for the recording events.

src/backend/core/api/viewsets.py
cameledev: added warning

src/backend/core/api/viewsets.py
cameledev: fixed, now using options.transcribe

src/backend/core/services/livekit_events.py
cameledev: fixed. checks if the Id was set

src/agents/metadata_collector.py
cameledev: changed for walrus operator: 

```
if (bucket_name := os.getenv("AWS_STORAGE_BUCKET_NAME")) is not None:
    self.bucket_name = bucket_name
else:
    raise MissingConfigError
```

src/agents/metadata_collector.py
cameledev: this should be okay as `self.events.append` as they are run in a single asyncio event loop

src/agents/metadata_collector.py
cameledev: Each entrypoint has its own JobContext , so since save is called at the end of the process this should not be a problem

src/backend/core/api/viewsets.py
```diff
@@ -336,6 +340,14 @@ def start_room_recording(self, request, pk=None):  # pylint: disable=unused-argu
                 status=drf_status.HTTP_500_INTERNAL_SERVER_ERROR,
             )
 
+        if settings.METADATA_COLLECTOR_ENABLED and recording.options.get(
+            "transcribe", False
+        ):
```
cameledev: This does not work. Check why. (replacing with "if True", we actually see the output file)

src/backend/core/recording/services/metadata_collector.py
cameledev: tests testing if start and stop are called were added

src/backend/core/api/viewsets.py
cameledev: `logger.warning(options)`

returns 

`language='fr' transcribe=None original_mode=None`

even when calling transcribe in frontend

src/agents/metadata_collector.py
```diff
@@ -0,0 +1,383 @@
...
+
+
+@dataclass
+class MetadataEvent:
+    """A single timestamped event recorded during a meeting."""
+
+    participant_id: str
+    type: str
+    timestamp: datetime
+    data: Optional[str] = None
```
cameledev, line 65: Shouldn't this be:

`data: Optional[dict]`

to handle a larger variety of events (admin muted another user, user sent file on the chat...) ?

src/agents/metadata_collector.py
```diff
@@ -0,0 +1,383 @@
...
+        except Exception:
+            logger.exception("Error treating the job for '%s'", room_name)
+            await job_req.reject()
+
+
+@server.rtc_session(agent_name=AGENT_NAME, on_request=handle_job_request)
+async def entrypoint(ctx: JobContext):
+    """Initialize and run the metadata collector."""
+    logger.info("Starting metadata agent in room: %s", ctx.room.name)
+    recording_id = ctx.job.metadata
```
cameledev, line 366: should we use `attributes here` ?
```suggestion
    recording_id = ctx.job.attributes["recording_id"]
```

cameledev: my bad, attributes are associated to individual participants, not room

src/backend/core/services/livekit_events.py
```diff
@@ -158,7 +162,7 @@ def _handle_egress_ended(self, data):
         """Handle 'egress_ended' event."""
 
         try:
-            recording = models.Recording.objects.get(
+            recording = models.Recording.objects.select_related("room").get(
```
cameledev, line 165: why are we changing this ?

src/helm/env.d/dev-keycloak/values.meet.yaml.gotmpl
```diff
@@ -282,12 +283,23 @@ agents:
     {{- end }}
     {{- end }}
     ENABLE_SILERO_VAD: "false"
+    AWS_S3_ENDPOINT_URL: minio.meet.svc.cluster.local:9000
+    AWS_S3_ACCESS_KEY_ID: meet
+    AWS_S3_SECRET_ACCESS_KEY: password
+    AWS_S3_SECURE_ACCESS: False
+    AWS_STORAGE_BUCKET_NAME: meet-media-storage
+    AWS_S3_OUTPUT_FOLDER: recordings
```
cameledev: shouldn't this be
```suggestion
    AWS_S3_OUTPUT_FOLDER: metadata
```

src/backend/core/services/livekit_events.py
cameledev: performance optimization

src/backend/core/api/viewsets.py
cameledev: it seems like `transcribe` is only being set when mode is `RecordingModeChoices.SCREEN_RECORDING`

cameledev: solved by latest commit to : src/frontend/src/features/recording/components/TranscriptSidePanel.tsx

cameledev: reverted change, and fixed condition for use of "transcribe"

src/backend/core/recording/services/metadata_collector.py
```diff
@@ -0,0 +1,91 @@
...
+        dispatch_id = getattr(response, "id", None)
+
+        if not dispatch_id:
+            logger.error("LiveKit response missing dispatch ID for room %s", room_id)
+            raise MetadataCollectorException(
+                f"LiveKit did not return a dispatch_id for room {room_id}"
+            )
+
+        recording.options["metadata_collector_dispatch_id"] = dispatch_id
+        await sync_to_async(recording.save)(update_fields=["options"])
```
lebaudantoine, line 59: let's discuss it

src/agents/metadata_collector.py
```diff
@@ -0,0 +1,384 @@
...
+    Agent,
+    AgentServer,
+    AgentSession,
+    AutoSubscribe,
+    JobContext,
+    JobProcess,
+    JobRequest,
+    RoomInputOptions,
+    RoomIO,
+    RoomOutputOptions,
```
lebaudantoine, line 24: "message": "RoomInputOptions and RoomOutputOptions are deprecated, use RoomOptions instead", "level": "WARNING", "name": "livekit.agents", "pid": 50, "job_id": "AJ_H3fwrGmgyQ5z", "room_id": "RM_FBCBg3FmnfqE", "timestamp": "2026-04-16T15:40:58.282963+00:00"}

src/agents/metadata_collector.py
```diff
@@ -0,0 +1,384 @@
...
+
+    async def cleanup():
+        logger.info("Shutting down metadata collector...")
+        await metadata_collector.aclose()
+
+    ctx.add_shutdown_callback(cleanup)
+
+
+if __name__ == "__main__":
+    cli.run_app(server)
```
lebaudantoine, line 382: I got this error stopping the recording : 

```
{"message": "Error closing session\nTraceback (most recent call last):\n  File \"/app/metadata_collector.py\", line 329, in _close_session\n    await session.drain()\n  File \"/usr/local/lib/python3.13/site-packages/livekit/agents/voice/agent_session.py\", line 804, in drain\n    raise RuntimeError(\"AgentSession isn't running\")\nRuntimeError: AgentSession isn't running", "level": "ERROR", "name": "metadata-collector", "exc_info": "Traceback (most recent call last):\n  File \"/app/metadata_collector.py\", line 329, in _close_session\n    await session.drain()\n  File \"/usr/local/lib/python3.13/site-packages/livekit/agents/voice/agent_session.py\", line 804, in drain\n    raise RuntimeError(\"AgentSession isn't running\")\nRuntimeError: AgentSession isn't running", "pid": 48, "job_id": "AJ_pdoT8MRwGodG", "room_id": "RM_FBCBg3FmnfqE", "timestamp": "2026-04-16T15:44:12.079380+00:00"}
```

src/agents/metadata_collector.py
```diff
@@ -0,0 +1,384 @@
...
+
+        logger.info("Started listening for participant events")
+
+    async def on_chat_message_received(
+        self, reader: rtc.TextStreamReader, participant_identity: str
+    ):
+        """Read a complete chat message and record it as an event."""
+        full_text = await reader.read_all()
+        logger.info(
+            "Received chat message from %s: '%s'", participant_identity, full_text
```
lebaudantoine: PII leak, chat message should not be logged. I guess we should reverify every logger.

src/agents/metadata_collector.py
lebaudantoine: Will be addressed in a dedicated PR.

src/agents/metadata_collector.py
lebaudantoine: Will be addressed in a dedicated PR.

PR #1169, opened 2026-03-18
feat: pass session ID to native apps via OIDC redirect URL

mmaudet:
## Summary

- When native apps (iOS, Android, Desktop) use OIDC login with a custom URL scheme as `returnTo` (e.g. `visio://auth-callback`), they cannot access the `sessionid` cookie set by Django because the system browser and the native app run in separate processes
- This adds a custom OIDC callback view that detects non-HTTPS redirect schemes and appends the session ID as a query parameter to the redirect URL
- Standard HTTPS redirects (web browser) are not affected — they continue using cookies as before

## Changes

- `src/backend/core/authentication/views.py` — New `OIDCAuthenticationCallbackView` that overrides `login_success()` to append session ID for custom-scheme redirects
- `src/backend/meet/settings.py` — Point `OIDC_CALLBACK_CLASS` to the new view

## Context

Native mobile/desktop apps (Visio Mobile) need to authenticate via OIDC. On iOS (`ASWebAuthenticationSession`) and Android (`Chrome Custom Tabs`), the system browser handles OIDC but cannot share cookies back to the app. With this change, the redirect URL becomes `visio://auth-callback?meet_sessionid=<session_key>`, allowing the app to extract the session ID directly.

## Test plan

- [ ] Web browser OIDC login still works (returnTo=https://...) — cookie-based, no change
- [ ] Native app OIDC login with returnTo=visio://auth-callback receives session ID in URL
- [ ] Session ID in URL is valid and can be used for API calls

### DIFF
--- a/src/backend/core/authentication/views.py
+++ b/src/backend/core/authentication/views.py
@@ -1,0 +1,47 @@
+"""Custom OIDC authentication views for native app support.
+
+When a native app (iOS, Android, Desktop) initiates OIDC login, it sets
+`returnTo` to a custom URL scheme (e.g. `visio://auth-callback`). After
+the OIDC flow completes, Django sets the session cookie and redirects to
+that URL. However, native apps cannot read browser cookies — they need
+the session ID passed explicitly in the redirect URL.
+
+This module overrides the OIDC callback view to append the session ID as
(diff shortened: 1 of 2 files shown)


lebaudantoine:
I guess your agent has wrongly reopened the PR you closed before.

PR #1186, opened 2026-03-19
🐛(frontend) proxy /media requests to nginx

FloChehab:
When running the build image with docker compose we need to proxy incoming /media request to nginx so that background images can be properly served.

We didn't have the issue with local dev with vite as vite is configured to do the proxy.
Nor do we have the issue with tilt as medias are served directly.

We need this because if the background image are hosted on another domain, by default (and that's to be fix in LiveKit), request to get the image don't have auth headers so they get rejected.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,10 @@ and this project adheres to
 - ♿️(frontend) fix sidepanel accessibility aria-label #1182
 - ♿️(frontend) fix more tools heading hierarchy #1181
 
+### Fixed
+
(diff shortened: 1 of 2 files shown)


FloChehab:
As discussed with @lebaudantoine, 

* We likely don't want to update the prod ngix config in the image,
* We would need to update `default.conf.template` mentionned in the docs for prod deployments (which also need to be updated for minio, etc.)
* And for the dev env, we could add a specific nginx conf mounted as volume in the root docker compose.
To be digged again by @lebaudantoine .

PR #1206, opened 2026-03-25
🧑‍💻(bin) update release tooling to support uv-based deps management

lebaudantoine:
Following the switch from pip to uv, prepare the release workflow to automatically run `uv lock` on backend and
keep dependencies up to date.

### DIFF
--- a/bin/prepare-release.sh
+++ b/bin/prepare-release.sh
@@ -101,6 +101,12 @@ update_npm_version "mail"
 # Update backend pyproject.toml
 update_python_version "backend"
 
+# Run uv lock in backend
+print_info "Running uv lock in backend..."
+cd "src/backend"
+uv lock
+cd -
+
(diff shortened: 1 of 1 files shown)


bin/prepare-release.sh
```diff
@@ -101,6 +101,12 @@ update_npm_version "mail"
 # Update backend pyproject.toml
 update_python_version "backend"
 
+# Run uv lock in backend
+print_info "Running uv lock in backend..."
+cd "src/backend"
+uv lock
```
FloChehab, line 107: I wouldn't run uv lock in CI

FloChehab (changes requested):

bin/prepare-release.sh
FloChehab: Or maybe it's linked to the auto update deps bot ?

PR #1247, opened 2026-04-07
✨(helm) add support multiple transcribe worker / endpoint

FloChehab:
Update the helm chart to support multiple transcribe worker in the summary service.
This is useful when using multiple WhisperX instances to have one deployment for each endpoint. This enables some kind of horizontal scaling (we still keep one call per WhisperX endpoint but can have multiple WhisperX endpoints)

Tested with tilt in local dev.


lebaudantoine (approved):
LGTM,

I tested this locally and went all the way through template generation and deployment on Tilt. Multiple instances are created. You currently have to manually update the Tiltfile to get multiple instances working in dev, but that’s not a big issue, it's acceptable.

However, the way it currently works is that extra-envs are appended after the default environment variables at deployment time. There’s no override mechanism, meaning the first value defined takes precedence. As a result, extra-envs come after the default env vars and don’t override them.

This raises a question: would it make more sense to place extra-envs first in the template? That way, if someone assumes they act as overrides (as I initially did), they would actually override the shared default values across pods. The current behavior might be a bit misleading.

Otherwise, this would imply that any values meant to be specifically defined via extra-envs should be removed from the shared env block.

That said, overall this looks great, thanks a lot, let's scale 🚀

FloChehab:
> This raises a question: would it make more sense to place extra-envs first in the template? That way, if someone assumes they act as overrides (as I initially did), they would actually override the shared default values across pods. The current behavior might be a bit misleading.

Ahh I was not aware of that, I would have thought the last one should be the truth would be the bahavior, we should definitely switch if with stick to this setup.

rouja (approved):

FloChehab:
Here's a quick write-up (reviewed AI summary for transparency) of the off-GitHub discussions we had, so the reasoning is documented and accessible to everyone. We're not set in stone on this — open to revisiting if better approaches emerge.

## Context & Problem

We're hitting a **scaling bottleneck on the WhisperX transcription service** (which we may eventually move out of Meet entirely):

* Queue times have been climbing as load increases on our single transcription worker
* The core constraint: **WhisperX is inherently single-request per GPU / per instance** (deployed from [meet-whisperx](https://github.com/suitenumerique/meet-whisperx))
* GPU capacity is the hard limit here

Bottom line:

> Each transcription needs exclusive access to a GPU-backed WhisperX instance.

---

## Constraints

1. **1 concurrent request per WhisperX API instance**
2. **1 GPU per active transcription — no real parallelism within an instance**
3. Scaling means **adding more GPU-backed instances**

---

## Options We Considered

Two main paths to scale transcription:

### Option 1 — Deploy "GPU Workers" directly on GPU nodes

✅ **Pros**

* Better performance (fewer network hops)
* No long-lived HTTP calls across clusters
* Tight coupling between worker and model instance
* STT API could be opened/shared with other services

❌ **Cons**

* Needs a new worker technology — Celery doesn't support keeping a model in RAM between tasks. Alternatives would be a custom pod (worker container + model container) using something like RedisQueue or Dramatiq
* Harder to span GPUs across different clusters or providers (doable, but not trivial)

---

### Option 2 — Deploy more WhisperX API instances and pair each with a dedicated worker

✅ **Pros**

* Uses the existing Docker image — already tested and deployed
* GPU nodes stay isolated (dedicated cluster, smaller attack surface)
* Works across multiple data centers / providers
* Easier to roll out incrementally
* Queuing and app-level logic stay close to the application for now

❌ **Cons**

* Added latency from HTTP calls between worker and API
* Potential inconsistency: worker healthy, WhisperX endpoint not (and vice versa)
* Long-lived HTTP requests for the duration of a transcription job (with the security implications that come with that)
* Version upgrades are more involved to coordinate

---

## ✅ Decision: Go with Option 2 (for now)

Option 2 fits better with our current deployment setup and short/mid-term strategy. It's not the most elegant solution, but it's the good call for where we are.

We're keeping the current PR design as-is because:

* It's the **simplest path that actually works reliably right now**
* It respects the **hard 1-transcription-per-GPU constraint**
* It avoids pulling in extra complexity:
  * No distributed locking
  * No additional coordination infra (Redis, etc.)
* Scaling stays **explicit and predictable**:

  > 1 Celery worker ↔ 1 WhisperX API ↔ 1 GPU

It also protects against misconfiguration:

> The Helm loop makes it structurally impossible to accidentally over-subscribe GPUs.

---

## Where We're Headed

This isn't the end state. We'll likely revisit the "true GPU worker" approach down the line, and we're keeping an eye on what's coming from the broader community.

The transcription service will most likely be decoupled from Meet at some point, to make it cleaner to share across applications.

lebaudantoine:
@coderabbitai full review

lebaudantoine:
discussed IRL with @qbey. He agrees to merge it.

PR #1251, opened 2026-04-08
✨(docker) add compose support for multi-user-transcriber

cameledev:
The multi-user transcriber service was not integrated into docker compose, limiting its usability in local and development environments. Add the necessary configuration to enable running and orchestrating the multi-user transcriber via compose.


Makefile
```diff
@@ -106,6 +108,10 @@ build-frontend: ## build the frontend container
 	@$(COMPOSE) build frontend
 .PHONY: build-frontend
 
+build-agents: ## build the multi-user-transcriber agent container
```
lebaudantoine, line 111: maybe, let's specify this command; agents is too vague, and would overlap with the metadata extractor.

Makefile
```diff
@@ -126,10 +132,15 @@ run-summary: ## start only the summary application and all needed services
 	@$(COMPOSE) up --force-recreate -d celery-summary-summarize
 .PHONY: run-summary
 
+run-agents: ## start the multi-user-transcriber agent
```
lebaudantoine, line 135: same

lebaudantoine (approved):
nit

Makefile
cameledev: I was thinking this would become:

```
build-agents: ## build the multi-user-transcriber agent container
	@$(COMPOSE) build multi-user-transcriber
	@$(COMPOSE) build metadata-collector
.PHONY: build-agents
```

once metadata-collector was implemented. 

What do you think ?

PR #1255, opened 2026-04-08
Fix misc issue with the tilt stack.

lebaudantoine:
Automatically copy required Kubernetes secrets when using the
Tilt development stack, as `make bootstrap` is not documented
as a prerequisite.

feedback from @arnaud-robin


FloChehab (approved):

arnaud-robin:
nice thanks for the fix

PR #1258, opened 2026-04-09
🩹(backend) fix swagger and redoc documentation urls

merahul22:
🩹(backend) fix swagger and redoc documentation urls

## Purpose

The Swagger and ReDoc routes were defined with a double slash in the Django URL configuration, which exposed the documentation on non-canonical paths like `/v1.0//swagger/` and `/v1.0//redoc/`. This change fixes the route definitions so the documentation is served on the expected URLs.

## Proposal

Update the backend URL patterns to use canonical documentation paths and add a regression test to prevent this issue from coming back.

- [x] replace the double-slash Swagger route with `/v1.0/swagger/`
- [x] replace the double-slash ReDoc route with `/v1.0/redoc/`
- [x] add a backend test that checks both documentation endpoints return `200`
- [x] add a changelog entry for the fix


lebaudantoine:
@coderabbitai full review

merahul22:
Good catch, thanks. You're right that `re_path()` treats `v1.0` as a regex pattern, so `.` can overmatch. I'll switch the ReDoc route to `path()` for literal matching.

I’ll also double-check the double-slash route strings to make sure Swagger and ReDoc stay consistent.

merahul22:
Good catch, thanks. I updated the ReDoc route to use `path()` instead of `re_path()` so the API version is matched literally, and kept it consistent with the Swagger routes.

merahul22:
@lebaudantoine can u check this pr ?

merahul22:
thanks @lebaudantoine

PR #1263, opened 2026-04-11
🐛(frontend) announce keyboard shortcut hint to screen readers

### LINKED ISSUE #1175: the keyboard shortcut help message is not rendered by screen readers
## Bug Report

**Problematic behavior**
When the video receives keyboard focus, a message appears visually: "Ctrl+Shift+/: access shortcuts directly." This message is contained in a `div` without an ARIA role or live region. It is not announced by screen readers.

Note: the given shortcut does not work on Mac. On Mac it is `Control + Shift + /`

**Expected behavior/code**
The keyboard shortcut help message must be announced by the screen reader when it becomes visible, i.e. when the video receives focus. This can be implemented via a role="status" or aria-live="polite" attribute on the message container.

**Steps to Reproduce**
1. Enable a screen reader
2. Navigate with the keyboard to the participant's video tile
3. Note that the message "Ctrl+Shift+/: access shortcuts directly." is not announced

**Additional context/Screenshots**
Image

**Critères RGAA**

[Critère 7.1 : Chaque script est-il, si nécessaire, compatible avec les technologies d'assistance ?](https://accessibilite.numerique.gouv.fr/methode/criteres-et-tests/#7.1)

**Impact**
A blind user navigating with the keyboard reaches the video without being informed of the `Ctrl+Shift+/` shortcut that allows access to the application's keyboard shortcuts. They cannot discover this navigation feature without seeing the screen.

merahul22:
## Summary

Fixes #1175 by making the keyboard shortcut hint available to screen readers when a participant tile receives keyboard focus.

## Changes

- keep the existing visual shortcut hint for keyboard users
- hide the visual-only hint from assistive technologies to avoid duplicate output
- add a polite live-region announcement for the shortcut help message
- trigger the announcement only while the participant tile has keyboard focus

## Testing

- `npm run lint`
- `npm run check`

## Notes

This keeps the current visual behavior unchanged while exposing the same shortcut guidance to screen-reader users.

### DIFF
--- a/src/frontend/src/features/rooms/livekit/components/KeyboardShortcutHint.tsx
+++ b/src/frontend/src/features/rooms/livekit/components/KeyboardShortcutHint.tsx
@@ -1,5 +1,6 @@
 import React, { ReactNode } from 'react'
 import { styled } from '@/styled-system/jsx'
+import { srOnly } from '@/styles/a11y'
 
 const Hint = styled('div', {
   base: {
@@ -26,6 +27,7 @@ const Hint = styled('div', {
 
 export interface KeyboardShortcutHintProps {
   children: ReactNode
+  announce?: boolean
 }
 
 /**
@@ -34,6 +36,21 @@ export interface KeyboardShortcutHintProps {
  */
 export const KeyboardShortcutHint: React.FC<KeyboardShortcutHintProps> = ({
   children,
+  announce = false,
 }) => {
-  return <Hint>{children}</Hint>
+  return (
+    <>
+      <Hint aria-hidden="true">{children}</Hint>
+      {announce ? (
+        <span
+          role="status"
+          aria-live="polite"
+          aria-atomic="true"
+          className={srOnly}
+        >
+          {children}
+        </span>
+      ) : null}
+    </>
+  )
 }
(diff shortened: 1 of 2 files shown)


Ovgodd (commented):
Hello thank you, 

Did you reproduce the problem first? As I mentioned in the issue, I wasn't able to reproduce it, 
NVDA already announces the shortcut hint correctly when the tile receives keyboard focus with the aria-label on the container and the existing toolbarHint translation.

Ovgodd (commented):
Hello thank you, 

Did you reproduce the problem first? As I mentioned in the issue, I wasn't able to reproduce it, 
NVDA already announces the shortcut hint correctly when the tile receives keyboard focus with the aria-label on the container and the existing toolbarHint translation.

merahul22:
Thanks for checking this.

I didn’t reproduce it with NVDA before opening the PR, so that’s on me. I inferred the issue from the code path and the original report, but if NVDA already announces the shortcut hint correctly when the tile receives keyboard focus, then this PR is probably not the right fix and may even be redundant.

I’m happy to close this PR if you think the current behavior is already correct.

PR #1264, opened 2026-04-13
🩹(make) fix indentation in Makefile

lebaudantoine:
Correct indentation issues introduced while updating Kubernetes commands, ensuring targets execute properly.


Ovgodd (approved):
thanks mate

PR #1270, opened 2026-04-14
♿️(frontend) add customizable accessibility fonts

### LINKED ISSUE #1069: Custom Font
## Feature Request

We can go even further by offering custom fonts for the user.
 I'm thinking of dyslexic people, if they want a sans-serif font. But there are certainly other possible cases.

Ovgodd:
## Purpose

This PR introduces a new accessibility preference that allows users to choose
the interface font for improved reading comfort.

It targets users with different readability needs (general readability, low
vision, dyslexia) and follows the existing accessibility settings architecture.

Capture d&#39;écran 2026-05-05 091238

## Proposal

- [x] Add `uiFont` preference in accessibility store
- [x] Install Fontsource packages and TypeScript module declarations
- [x] Pin new Fontsource dependency versions
- [x] Apply selected font globally via `--app-font-family`
- [x] Keep runtime theming compatibility for self-hosters (`--fonts-sans`)
- [x] Make LiveKit inherit the same font variable
- [x] Add font selector to Accessibility settings UI
- [x] Add FR/EN/NL/DE translations for the new setting
- [x] Add dynamic import error handling in `useApplyUiFont`


Ovgodd:
@coderabbitai

src/frontend/package.json
```diff
@@ -13,8 +13,11 @@
     "check": "prettier --check ./src"
   },
   "dependencies": {
+    "@fontsource-variable/atkinson-hyperlegible-next": "^5.2.6",
+    "@fontsource-variable/lexend": "^5.2.11",
     "@fontsource-variable/material-symbols-outlined": "5.2.34",
     "@fontsource/material-icons-outlined": "5.2.6",
+    "@fontsource/opendyslexic": "^5.2.5",
```
lebaudantoine: nit: pin dependency

src/frontend/src/locales/fr/settings.json
```diff
@@ -117,6 +117,16 @@
     "announceReactions": {
       "label": "Vocaliser les réactions"
     },
+    "font": {
+      "label": "Police d'affichage",
+      "description": "Personnalisez la police utilisée dans l'interface pour améliorer votre confort de lecture.",
+      "options": {
+        "default": "Par défaut (système)",
```
lebaudantoine: Actually, the system font is only used in the white-labeled image. In our custom Dinum image, the font used is Marianne.

I suggest not to mention `(système)` in the label

src/frontend/src/stores/accessibility.ts
```diff
@@ -62,13 +95,15 @@ type AccessibilityState = {
...
   captionBackgroundColor: CaptionColor
+  uiFont: UiFont
 }
 
 const DEFAULT_STATE: AccessibilityState = {
   announceReactions: false,
   captionTextSize: 'medium',
   captionFontColor: 'default',
   captionBackgroundColor: 'default',
+  uiFont: 'default',
```
lebaudantoine, line 86: I don't think it's the accessibility store's responsibility to define and maintain the default font stack. The store should only know how to override the default, the default itself belongs to the design system / global CSS.

Right now UI_FONT_STACKS.default duplicates what global CSS already declares, and the two can drift. I'd suggest:

- Define the base font stack once in global CSS as --app-font-family-base, with --app-font-family defaulting to it.
- In the accessibility store, drop default from the stacks map and keep only the overrides (lexend, atkinson, opendyslexic).
- The hook either sets --app-font-family to "<override>, var(--app-font-family-base)", or removes the property entirely to fall back to the base.

That way the store stays purely additive and doesn't need to be updated if the design system changes its default font.

lebaudantoine, line 86: Worth flagging: the current setup is also a breaking change for self-hosters. [We document a way to inject a custom CSS file at runtime](https://github.com/suitenumerique/meet/blob/main/docs/theming.md) to override design tokens and fonts without rebuilding the Docker image (e.g. swapping in Marianne for DINUM deployments). 

With the accessibility store hardcoding SYSTEM_SANS_STACK as the default and writing it to --app-font-family on every mount, any custom font set via the override CSS gets clobbered as soon as the hook runs.

This reinforces the point: the accessibility store shouldn't own the default font. If the base stack lives in CSS as --app-font-family-base, self-hosters keep their override path (they just redefine that variable in their custom CSS), and the accessibility layer stays purely additive on top of whatever the deployment has configured.

lebaudantoine:
@coderabbitai full review

src/frontend/src/locales/fr/settings.json
Ovgodd: I understand !

src/frontend/src/stores/accessibility.ts
Ovgodd: Yes that's make sense

Ovgodd: hm I see yes, thanks for explanations,

so I removed `SYSTEM_SANS_STACK` from the store. 
The base font stack now lives in CSS as `--app-font-family-base: var(--fonts-sans)`, and the hook only sets `--app-font-family` for non-default overrides (or removes the property to fall back to the base). Self-hosters redefining `--fonts-sans` via `FRONTEND_CSS_URL` are no longer clobbered.

src/frontend/src/locales/fr/settings.json
```diff
@@ -121,7 +121,7 @@
       "label": "Police d'affichage",
       "description": "Personnalisez la police utilisée dans l'interface pour améliorer votre confort de lecture.",
       "options": {
-        "default": "Par défaut (système)",
+        "default": "Par défaut",
```
lebaudantoine, line 124: I sugest removing (système) for every locale.

lebaudantoine (approved):

PR #1276, opened 2026-04-16
📌(agents) pin Docker image to a specific tag for reproducible builds

lebaudantoine:
Avoid using floating tags and pin the image to an explicit version to ensure consistent and reproducible agent builds.


cameledev (approved):
yes

PR #1277, opened 2026-04-16
✨ Support for voxtral realtime stt

cameledev:
## Purpose

Add support for voxtral realtime stt (live transcription).

## Why we can't use mistralai plugin

The official `livekit-plugins-mistralai` plugin is hardwired to **Mistral's hosted cloud API**   both the URL path and the wire protocol. We self-host Voxtral on **vLLM**, which speaks a different protocol on a different path, so the plugin can't be pointed at it. Four distinct issues, in the order they occured:

### 1. Hardcoded WebSocket path

The Mistral SDK's `RealtimeTranscription._build_url` hardcodes the path `/v1/audio/transcriptions/realtime`. `server_url` only controls scheme + host   there is no kwarg to change the suffix, and the LiveKit plugin calls `connect()` without overriding it either. vLLM serves Voxtral realtime at **`/v1/realtime`**. That route doesn't exist on vLLM, so every WebSocket connnection attemps returned **HTTP 500**. A monkeypatch on `_build_url` (`/v1/audio/transcriptions/realtime` → `/v1/realtime`) fixed this issue.

### 2. Incompatible wire protocol (most important issue)

Even with the path fixed, the two sides speak different languages:

| | Mistral plugin/SDK | vLLM Voxtral |
|---|---|---|
| **Protocol** | Mistral proprietary | OpenAI-Realtime derived |
| **Server msgs** | `RealtimeTranscriptionSessionCreated`, `TranscriptionStreamTextDelta`, `TranscriptionStreamDone` | `session.created`, `transcription.delta`, `transcription.done` |
| **Client msgs** | `send_audio` / `flush_audio` / `end_audio` | `input_audio_buffer.append` / `.commit` |

The Mistral SDK cannot parse vLLM's messages. Monkeypatching this would mean reimplementing the SDK's entire receive path.

### 3. The handshake failures (how the mismatch actually surfaced)

`rt.connect()` does two sequential handshakes, and the plugin broke at both:

- **WS upgrade handshake (transport).** Before the path fix, the upgrade itself was rejected: `websockets.exceptions.InvalidStatus: server rejected WebSocket connection: HTTP 500` thrown out of `ConnectionPool.prewarm`. 
- **Realtime session handshake (application).** After the `_build_url`
monkeypatch, the upgrade *succeeded* (`[accepted]` / `connection open` in vLLM logs), but `_recv_handshake` then blocked on `await websocket.recv()` waiting for **Mistral's** `session.created` event. vLLM never sends that event (its OpenAI-style handshake has a different shape), so the 10s LiveKit connect timeout fired → `CancelledError` → `TimeoutError` → `APIConnectionError: Connection error.`, retried, then session closed as unrecoverable.

### 4. Dependency resolution failure (packaging)

Independently, `livekit-plugins-mistralai==1.5.4` pulls in`mistralai[realtime]>=2.0.0`. Under `uv`'s *universal* resolution across
`requires-python = ">=3.12"` (which now includes 3.14), that extra is
unresolvable → `requirements are unsatisfiable`


src/agents/multi-user-transcriber.py
```diff
@@ -46,6 +47,15 @@ def create_stt_provider():
...
     elif STT_PROVIDER == "kyutai":
         _stt_instance = kyutai.STT(base_url=os.getenv("KYUTAI_STT_BASE_URL"))
+    elif STT_PROVIDER == "voxtral-realtime":
+        client = Mistral(
+            api_key=os.getenv("VOXTRAL_REALTIME_STT_API_KEY"),
+            server_url=os.getenv("VOXTRAL_REALTIME_STT_BASE_URL"),
+        )
+        _stt_instance = mistralai.STT(
+            client=client,
+            model="voxtral-mini-latest",
```
lebaudantoine: should we make it an env variable?

src/agents/multi-user-transcriber.py
```diff
@@ -46,6 +47,15 @@ def create_stt_provider():
...
+        )
```
lebaudantoine: we could refactor it to use a factory pattern with a declarative approach based on a decorator. 

```
from typing import Callable, Dict
import os

# Registry of STT provider factories
_STT_PROVIDERS: Dict[str, Callable] = {}


def register_stt_provider(name: str):
    """Decorator to register an STT provider factory."""
    def decorator(func: Callable) -> Callable:
        _STT_PROVIDERS[name] = func
        return func
    return decorator


@register_stt_provider("deepgram")
def _create_deepgram_stt():
    # Note: Not all Deepgram API parameters are supported by the LiveKit plugin.
    # detect_language is NOT supported for real-time streaming.
    # Use language="multi" instead for automatic multilingual support.
    return deepgram.STT(
        model=os.getenv("DEEPGRAM_STT_MODEL", "nova-3"),
        language=os.getenv("DEEPGRAM_STT_LANGUAGE", "multi"),
    )


@register_stt_provider("kyutai")
def _create_kyutai_stt():
    return kyutai.STT(base_url=os.getenv("KYUTAI_STT_BASE_URL"))


@register_stt_provider("voxtral-realtime")
def _create_voxtral_realtime_stt():
    client = Mistral(
        api_key=os.getenv("VOXTRAL_REALTIME_STT_API_KEY"),
        server_url=os.getenv("VOXTRAL_REALTIME_STT_BASE_URL"),
    )
    return mistralai.STT(
        client=client,
        model=os.getenv("VOXTRAL_REALTIME_STT_MODEL", "voxtral-mini-latest"),
    )


def create_stt_provider():
    """Create STT provider based on environment configuration."""
    try:
        factory = _STT_PROVIDERS[STT_PROVIDER]
    except KeyError:
        raise ValueError(
            f"Unknown STT provider: {STT_PROVIDER!r}. "
            f"Available providers: {sorted(_STT_PROVIDERS)}"
        )
    return factory()
```

src/agents/multi-user-transcriber.py
cameledev: agreed. will do

src/agents/multi-user-transcriber.py
cameledev: As long as we are dealing with a limited amount of models, I would keep the current implem. I find The suggested approach adds code and abstraction without improving performance or clarity.

src/frontend/src/features/subtitle/component/Subtitles.tsx
```diff
@@ -75,15 +77,15 @@ const useTranscriptionState = () => {
...
-      ]
+      const existingIndex = prevSegments.findIndex(
+        (s: TranscriptionSegmentWithParticipant) => s.id === segment.id
+      )
+      if (existingIndex === -1) {
+        return [...prevSegments, { participant, ...segment }]
+      }
+      const next = prevSegments.slice()
+      next[existingIndex] = { ...next[existingIndex], ...segment }
+      return next
```
cameledev, line 86: allow update of previously received segments

src/backend/core/api/__init__.py
```diff
@@ -65,7 +65,7 @@ def get_frontend_configuration(request):
             else None,
             "default_country": settings.ROOM_TELEPHONY_DEFAULT_COUNTRY,
         },
-        "subtitle": {"enabled": settings.ROOM_SUBTITLE_ENABLED},
+        "subtitle": {"enabled": True},  # settings.ROOM_SUBTITLE_ENABLED},
```
cameledev: REVERT BEFORE MERGE

src/backend/core/api/viewsets.py
cameledev, line 627: REVERT BEFORE MERGE (2/2)

cameledev (commented):
.

src/backend/core/api/viewsets.py
```diff
@@ -627,8 +586,7 @@ def invite(self, request, pk=None):  # pylint: disable=unused-argument
             permissions.HasLiveKitRoomAccess,
         ],
         authentication_classes=[LiveKitTokenAuthentication],
-    )
-    @FeatureFlag.require("subtitle")
+    )  # @FeatureFlag.require("subtitle")
```
lebaudantoine: revert before merge

compose.yml
```diff
@@ -342,6 +332,17 @@ services:
         - action: rebuild
           path: ./src/summary
 
+  multi-user-transcriber:
+    build:
+      context: ./src/agents
+      target: development    
+    command: ["python", "multi_user_transcriber.py", "dev"]
```
lebaudantoine: It's not necessary, but I understand if you want to keep more control over the compose service

lebaudantoine:
@greptileai

lebaudantoine:
@coderabbitai full review

lebaudantoine:
@greptileai

PR #1281, opened 2026-04-17
Validate room configuration

cameledev:
## Purpose

(from https://github.com/orgs/suitenumerique/projects/11/views/1?pane=issue&itemId=177109644)

### Summary

`Room.configuration` is a `JSONField` with no schema validation, allowing arbitrary/malicious payloads to be stored. This has already been flagged in bug bounty reports.

---

### Problem

* Accepts any JSON → no structure enforcement
* Potential security risks (unexpected/malicious data)
* Frontend assumes a defined shape, backend does not enforce it
* No clear contract for this field

---

### Proposed Fix

**1. Define a schema (Pydantic)**
Derive expected structure from frontend usage. This data is only an example and doesn't reflect at all the expected model.

```python
from pydantic import BaseModel, Field
from typing import Optional

class RoomConfiguration(BaseModel):
    is_recording_enabled: bool = False
    max_participants: Optional[int] = Field(default=None, ge=1)
    waiting_room_enabled: bool = False
    allow_guest_access: bool = True
```

---

**2. Validate at API level (serializer)**

```python
def validate_configuration(value):
    try:
        return RoomConfiguration(**value).dict()
    except ValidationError as e:
        raise serializers.ValidationError(e.errors())
```

you can draw inspiration from https://github.com/suitenumerique/meet/blob/aaf21e97e85f48fdd051359414c1139530427bfc/src/backend/core/api/serializers.py#L224

---

**Impact:** closes a security gap and enforces a clear data contract.

### DIFF
--- a/src/backend/core/api/serializers.py
+++ b/src/backend/core/api/serializers.py
@@ -14,6 +14,7 @@ from django.utils.translation import gettext_lazy as _
 
 from django_pydantic_field.rest_framework import SchemaField
 from pydantic import BaseModel, Field
+from pydantic import ValidationError as PydanticValidationError
 from rest_framework import serializers
 from rest_framework.exceptions import PermissionDenied
 from timezone_field.rest_framework import TimeZoneSerializerField
@@ -130,6 +131,16 @@ class RoomSerializer(serializers.ModelSerializer):
         model = models.Room
         fields = ["id", "name", "slug", "configuration", "access_level", "pin_code"]
         read_only_fields = ["id", "slug", "pin_code"]
+
+    def validate_configuration(self, value):
+        """Validate room configuration against the RoomConfiguration schema."""
+        if value is None or value == {}:
+            return value
+        try:
+            RoomConfiguration.model_validate(value)
+        except PydanticValidationError as e:
+            raise serializers.ValidationError(e.errors()) from e
+        return value
 
     def to_representation(self, instance):
         """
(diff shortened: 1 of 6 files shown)


src/backend/core/api/serializers.py
```diff
@@ -306,6 +317,22 @@ class MuteParticipantSerializer(BaseParticipantsManagementSerializer):
     )
 
 
+RoomConfigurationTrackSource = Literal[
+    "camera", "microphone", "screen_share", "screen_share_audio"
+]
```
cameledev, line 322: not sure I understand why it exists both capitalized and un-capitalized

lebaudantoine: You're right, that’s on me.

This should be straightforward to fix. The issue comes from the backend exposing lowercase strings to the frontend, whereas uppercase values are expected.

We’ll likely need to introduce a mechanism to handle case normalization (lower ↔ upper) for existing rooms in the database. Additionally, it might be worth running a migration script to update existing entries for consistency.

Relevant code reference:
https://github.com/suitenumerique/meet/blob/main/src/backend/meet/settings.py#L629

Can you handle it? Thus we can reuse the existing TrackSource type.

src/backend/core/tests/rooms/test_api_rooms_update.py
```diff
@@ -104,7 +104,97 @@ def test_api_rooms_update_administrators():
...
+    client.force_login(user)
+
+    response = client.patch(
+        f"/api/v1.0/rooms/{room.id!s}/",
+        {"configuration": {"can_publish_sources": ["INVALID_SOURCE"]}},
+        format="json",
+    )
+    assert response.status_code == 400
+    room.refresh_from_db()
+    assert room.configuration == {}
```
lebaudantoine: nit: building on your comment, it would be useful to add a test case where the track source is passed in a capitalized form. That way, if we later refactor this quirk, the test suite will catch it by failing and signal that the behavior has changed.

lebaudantoine (commented):
LGTM

src/backend/core/tests/rooms/test_api_rooms_update.py
cameledev: Proposed change:

```
@pytest.mark.parametrize("invalid_source", ["invalid_source", "CAMERA"])
def test_api_rooms_update_configuration_invalid_source_value(invalid_source):
    """Invalid source values should be rejected."""
    user = UserFactory()
    room = RoomFactory(users=[(user, "owner")])
    client = APIClient()
    client.force_login(user)

    response = client.patch(
        f"/api/v1.0/rooms/{room.id!s}/",
        {"configuration": {"can_publish_sources": [invalid_source]}},
        format="json",
    )
    assert response.status_code == 400
    room.refresh_from_db()
    assert room.configuration == {}
```

src/backend/core/api/serializers.py
cameledev: https://github.com/suitenumerique/meet/issues/1282

lebaudantoine (approved):

PR #1283, opened 2026-04-19
Allow room configuration and access level via external API

### LINKED ISSUE #1260: Add API endpoint (or creation parameter) to set admin & access options for virtual rooms via a service account
## Feature Request

We use the API with a service account to create video rooms on behalf of users in our organization.
Problem: The API does not let us set the room’s admin parameters; the user has to log in before the meeting to configure them manually.
Desired feature: An API endpoint that allows a service account to set a room’s administration settings for a user.

We need an admin API endpoint that, using a service account, can set a user’s virtual‑room parameters:

Moderation settings – turn on microphone, enable video, allow screen sharing.
Access settings – Open, Open to trusted people, or Restricted.
If security restrictions prevent a service account from modifying an existing user’s room, these settings should at least be accepted by the room‑creation API.

We've developed a bot that is able to organize meetings and provide Visio links. The n8n workflow that maganes to room creation could be adapted so that the bot is able to set moderation and security parameters in advance.

I can do some tests and share the n8n workflow

merahul22:
## Purpose
This Pull Request addresses issue #1260 by enhancing the external API to support granular room configuration. This allows service accounts (like n8n bots or other integrations) to set moderation and access settings programmatically during room creation and updates.

### Backend
- **RoomSerializer**: Modified to whitelist `configuration` (JSON) and `access_level` as input fields. Updated the `create` logic to respect provided values while maintaining secure defaults.
- **RoomViewSet**: Added `UpdateModelMixin` and `DestroyModelMixin` to expose `PATCH`, `PUT`, and `DELETE` endpoints for rooms via the external API.

### Tests
- Updated `test_external_api_rooms.py` with new test cases:
    - `test_api_rooms_create_with_params_success`: Verifies custom config injection.
    - `test_api_rooms_update_success`: Verifies room modification.
    - `test_api_rooms_delete_success`: Verifies room deletion.
- Adjusted existing assertions to account for the new `configuration` field in API responses.

## Verification
- Verified manually using a local dev environment.
- All backend tests passed: `make test-back args="core/tests/test_external_api_rooms.py"`.

Closes #1260

### DIFF
--- a/src/backend/core/external_api/serializers.py
+++ b/src/backend/core/external_api/serializers.py
@@ -34,10 +34,16 @@ class RoomSerializer(serializers.ModelSerializer):
     following the principle of least privilege.
     """
 
+    access_level = serializers.ChoiceField(
+        choices=models.RoomAccessLevel.choices,
+        required=False,
+    )
+    configuration = serializers.JSONField(required=False)
+
     class Meta:
         model = models.Room
-        fields = ["id", "name", "slug", "pin_code", "access_level"]
-        read_only_fields = ["id", "name", "slug", "pin_code", "access_level"]
+        fields = ["id", "name", "slug", "pin_code", "access_level", "configuration"]
+        read_only_fields = ["id", "name", "slug", "pin_code"]
 
     def to_representation(self, instance):
         """Enrich response with application-specific computed fields."""
@@ -68,6 +74,7 @@ class RoomSerializer(serializers.ModelSerializer):
 
         # Set secure defaults
         validated_data["name"] = utils.generate_room_slug()
-        validated_data["access_level"] = models.RoomAccessLevel.TRUSTED
+        validated_data.setdefault("access_level", models.RoomAccessLevel.TRUSTED)
+        validated_data.setdefault("configuration", {})
 
         return super().create(validated_data)
--- a/src/backend/core/external_api/serializers.py
+++ b/src/backend/core/external_api/serializers.py
@@ -4,10 +4,11 @@
 
 from django.conf import settings
 
+from pydantic import ValidationError
 from rest_framework import serializers
 
 from core import models, utils
-from core.api.serializers import BaseValidationOnlySerializer
+from core.api.serializers import BaseValidationOnlySerializer, RoomConfiguration
 
 OAUTH2_GRANT_TYPE_CLIENT_CREDENTIALS = "client_credentials"
 
@@ -34,10 +35,37 @@ class RoomSerializer(serializers.ModelSerializer):
     following the principle of least privilege.
     """
 
+    configuration = serializers.JSONField(required=False)
+
     class Meta:
         model = models.Room
-        fields = ["id", "name", "slug", "pin_code", "access_level"]
-        read_only_fields = ["id", "name", "slug", "pin_code", "access_level"]
+        fields = ["id", "name", "slug", "pin_code", "access_level", "configuration"]
+        read_only_fields = ["id", "name", "slug", "pin_code"]
+
+    def validate_configuration(self, value):
+        """Validate room configuration against the RoomConfiguration schema."""
+        if value is None or value == {}:
+            return value
+        try:
+            RoomConfiguration.model_validate(value)
+        except ValidationError as e:
+            raise serializers.ValidationError(e.errors()) from e
+        return value
+
+    def validate_access_level(self, access_level):
+        """Reject public access_level unless explicitly allowed or the default is already public."""
+
+        if settings.EXTERNAL_API_DEFAULT_ACCESS_LEVEL == models.RoomAccessLevel.PUBLIC:
+            return access_level
+
+        if (
+            access_level == models.RoomAccessLevel.PUBLIC
+            and not settings.EXTERNAL_API_ALLOW_PUBLIC_ACCESS
+        ):
+            raise serializers.ValidationError(
+                "Public rooms are disabled for the external API."
+            )
+        return access_level
 
     def to_representation(self, instance):
         """Enrich response with application-specific computed fields."""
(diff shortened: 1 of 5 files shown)


lebaudantoine:
@coderabbitai full review

src/backend/core/external_api/serializers.py
```diff
@@ -34,10 +34,16 @@ class RoomSerializer(serializers.ModelSerializer):
     following the principle of least privilege.
     """
 
+    access_level = serializers.ChoiceField(
+        choices=models.RoomAccessLevel.choices,
+        required=False,
+    )
+    configuration = serializers.JSONField(required=False)
```
lebaudantoine, line 38: We should replicate / reuse elements from this PR https://github.com/suitenumerique/meet/pull/1281, to validate the room's configuration.

src/backend/core/external_api/serializers.py
```diff
@@ -68,6 +74,7 @@ def create(self, validated_data):
 
         # Set secure defaults
         validated_data["name"] = utils.generate_room_slug()
-        validated_data["access_level"] = models.RoomAccessLevel.TRUSTED
+        validated_data.setdefault("access_level", models.RoomAccessLevel.TRUSTED)
```
lebaudantoine: Great, thank you.

From a security standpoint, and to avoid introducing a breaking change, I would suggest adding a setting to control whether the access level can be set to public via the external API.

Unlike /api, the /external-api was not originally designed to accept rooms without a lobby. It would be beneficial to make this behavior configurable through a Django setting.

For example, something like:

```
EXTERNAL_API_ALLOW_PUBLIC_ACCESS = False
EXTERNAL_API_ALLOW_ROOMS_WITHOUT_LOBBY = False

```

This would allow projects to explicitly opt in to this behavior while preserving the current default and ensuring backward compatibility.

src/backend/core/external_api/viewsets.py
```diff
@@ -156,6 +156,8 @@ def generate_jwt_access_token(self, request, *args, **kwargs):
 class RoomViewSet(
     mixins.CreateModelMixin,
     mixins.RetrieveModelMixin,
+    mixins.UpdateModelMixin,
+    mixins.DestroyModelMixin,
```
lebaudantoine: Adding `DestroyModelMixin` or `UpdateModelMixin` is out of scope for this PR. Let’s stay focused on enabling additional configuration when creating a room through the external API.

Alternatively, these changes could be introduced in separate, dedicated commits with proper unit test coverage.

lebaudantoine:
@merahul22 do you want me to adress my comments?

merahul22:
@lebaudantoine  Yes, I’ll address your comments. I’m planning to update the PR to keep it focused on create-time configuration, and I’ll follow up once I’ve pushed the changes.

src/backend/core/external_api/serializers.py
merahul22: I reused the room-configuration validation pattern from #1281 so the external API only accepts the supported configuration keys and values, instead of allowing free-form JSON.

src/backend/core/external_api/serializers.py
merahul22: Agreed. I added opt-in settings so public access and no-lobby behavior stay disabled by default and only work when explicitly enabled.

src/backend/core/external_api/viewsets.py
merahul22: Understood. I removed the write endpoints and kept this PR focused on create-time room configuration through the external API.

src/backend/core/tests/test_external_api_rooms.py
```diff
@@ -565,34 +597,56 @@ def test_api_rooms_create_readonly_enforcement():
...
 
     client = APIClient()
     client.credentials(HTTP_AUTHORIZATION=f"Bearer {token}")
-    response = client.patch(f"/external-api/v1.0/rooms/{room.id}/")
+    response = client.delete(f"/external-api/v1.0/rooms/{room.id}/")
 
-    assert response.status_code == 405
-    assert 'method "patch" not allowed.' in str(response.data).lower()
+    assert response.status_code == 204
+    assert not Room.objects.filter(id=room.id).exists()
```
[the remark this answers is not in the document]
merahul22: I see the gap. I’m going to keep this PR focused on create-time room configuration and remove the unsupported write-endpoint path from scope here, so I won’t add the PATCH/DELETE negative cases in this PR. If we keep write endpoints later, I’ll add the member/admin denial and missing-scope tests then.

src/backend/core/external_api/serializers.py
```diff
@@ -34,10 +34,16 @@ class RoomSerializer(serializers.ModelSerializer):
...
+        required=False,
+    )
+    configuration = serializers.JSONField(required=False)
+
     class Meta:
         model = models.Room
-        fields = ["id", "name", "slug", "pin_code", "access_level"]
-        read_only_fields = ["id", "name", "slug", "pin_code", "access_level"]
+        fields = ["id", "name", "slug", "pin_code", "access_level", "configuration"]
+        read_only_fields = ["id", "name", "slug", "pin_code"]
```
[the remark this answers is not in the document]
merahul22: I’ve added explicit validation for configuration and limited it to the supported room-config keys/values. Since this PR is being kept create-only, the PATCH merge issue is no longer part of the final scope here, but I agree the same validation/merge behavior would be needed if partial updates are added later.

lebaudantoine:
@coderabbitai full review

lebaudantoine:
@coderabbitai full review

lebaudantoine:
@coderabbitai full review

lebaudantoine (approved):
LGTM

merahul22:
thanks @lebaudantoine

PR #1284, opened 2026-04-19
fix: handle missing contentType in MinIO webhook payload

### LINKED ISSUE #973: Bug: MinIO webhook returns 400 - Missing contentType in payload
### Environment
- Deployment: Docker Compose
- Version: latest (lasuite/meet-backend:latest)
- MinIO version: RELEASE.2025-09-07T16-13-09Z

### Description
The MinIO webhook integration for recordings fails with 400 Bad Request because the `MinioParser` expects a `contentType` field that MinIO doesn't always send in webhook payloads.

### Configuration
```yaml
# env.d/common
RECORDING_ENABLE=True
RECORDING_STORAGE_EVENT_ENABLE=True
RECORDING_ENABLE_STORAGE_EVENT_AUTH=False
RECORDING_EVENT_PARSER_CLASS=core.recording.event.parsers.MinioParser
```

### Error
```python
ParsingEventDataError: "Missing or malformed key: 'contentType'."
```

### Root Cause
In `core/recording/event/parsers.py`, line ~95:
```python
filetype = file_object["contentType"]  # This key is not always present
```

MinIO webhook payloads don't always include `contentType`, especially for programmatically uploaded files.

### Proposed Solution
Make `contentType` optional and infer from file extension:
```python
if "contentType" in file_object:
    filetype = file_object["contentType"]
else:
    # Infer from extension
    if filepath.endswith('.mp4'):
        filetype = "video/mp4"
    elif filepath.endswith('.ogg'):
        filetype = "audio/ogg"
```

### Workaround
Created a custom parser that handles missing contentType.

Sanjays2402:
## Purpose

Fixes #973.

MinIO doesn't always include `contentType` in the S3 event object payload, causing a `KeyError` that results in a 400 response with "Missing contentType in payload". This PR makes the parser tolerate a missing/`None` `contentType` by falling back to `mimetypes.guess_type()` on the file path.

## Proposal

- [x] `src/backend/core/recording/event/parsers.py`: use `.get("contentType")` with a `mimetypes.guess_type()` fallback instead of direct dict access.
- [x] Tests updated and added for missing / `None` contentType scenarios.

### How it works

```python
filetype = file_object.get("contentType") or mimetypes.guess_type(filepath)[0] or ""
```

1. Try to get `contentType` from the payload.
2. If missing or `None`, guess from the file extension.
3. If unguessable, default to an empty string (validation catches invalid types downstream).

### DIFF
--- a/src/backend/core/recording/event/parsers.py
+++ b/src/backend/core/recording/event/parsers.py
@@ -1,6 +1,7 @@
 """Meet storage event parser classes."""
 
 import logging
+import mimetypes
 import re
 from dataclasses import dataclass
 from functools import lru_cache
@@ -104,7 +105,7 @@ class MinioParser:
             bucket_name = s3["bucket"]["name"]
             file_object = s3["object"]
             filepath = file_object["key"]
-            filetype = file_object["contentType"]
+            filetype = file_object.get("contentType") or mimetypes.guess_type(filepath)[0] or ""
         except (KeyError, IndexError) as e:
             raise ParsingEventDataError(f"Missing or malformed key: {e}.") from e
         try:
--- a/src/backend/core/tests/recording/event/test_parsers.py
+++ b/src/backend/core/tests/recording/event/test_parsers.py
@@ -81,25 +81,65 @@ def test_parse_missing_keys(minio_parser):
         minio_parser.parse(invalid_minio_event)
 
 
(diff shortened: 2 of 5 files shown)


lebaudantoine:
I'll close it as duplicate of #1291

PR #1300, opened 2026-04-29
♻️(backend) align CSRF token header with Django conventions

lebaudantoine:
Update the CSRF header naming to follow Django standards, avoiding duplicated client-side logic with inconsistent header names.


lebaudantoine:
@coderabbitai full review

PR #1309, opened 2026-04-30
🩹(frontend) fix reconnect loop caused by connectionObserverStore updates

lebaudantoine:
Updating connectionObserverSnapshot triggered page re-renders, causing participants to reconnect due to a race condition.

Read directly from the underlying store instead of using the snapshot to avoid unnecessary re-renders.


lebaudantoine:
@coderabbitai full review

lebaudantoine:
@coderabbitai full review

PR #1316, opened 2026-04-30
🐛(backend) make start-recording atomic and unblock room on worker failure

EpsilonFO:
## Purpose

Fix two concurrency issues in `POST /rooms/{id}/start-recording/`:

1. **Partial failure leaves orphan rows** — `Recording` and `RecordingAccess`
   were created in two separate auto-committed queries. If `RecordingAccess`
   creation failed, a `Recording` row remained without owner.

2. **Worker failure locks the room** — when `worker_manager.start()` raised
   `RecordingStartError`, the `Recording` stayed in `INITIATED`. Because of
   the partial unique constraint
   `unique_initiated_or_active_recording_per_room`, no new recording could
   be started on the same room without manual DB cleanup.

## Proposal

- [x] Wrap `Recording` + `RecordingAccess` creation in `transaction.atomic()`
- [x] Return **409 Conflict** instead of 500 when a recording is already in
      progress (catches both `IntegrityError` and `ValidationError` from
      `full_clean()`)
- [x] On `RecordingStartError`, transition the `Recording` to
      `FAILED_TO_START` so the room is unblocked for future attempts
- [x] Add two new tests covering the 409 path and the unblock-after-failure
      path
- [x] Update existing `test_start_recording_worker_error` to assert the new
      status
- [x] Changelog entry added

### DIFF
--- a/src/backend/core/api/viewsets.py
+++ b/src/backend/core/api/viewsets.py
@@ -6,7 +6,9 @@ from logging import getLogger
 from urllib.parse import unquote, urlparse
 
 from django.conf import settings
+from django.core.exceptions import ValidationError as DjangoValidationError
 from django.core.files.storage import default_storage
+from django.db import IntegrityError, transaction
 from django.db.models import Q
 from django.http import Http404
 from django.shortcuts import get_object_or_404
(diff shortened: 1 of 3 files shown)


lebaudantoine (commented):
LGTM; I'll just run a coderabbitai review before merging

lebaudantoine:
@coderabbitai full review

lebaudantoine:
@EpsilonFO don't lose time fixing the lint-git job !

src/backend/core/api/viewsets.py
```diff
@@ -320,23 +322,37 @@ def start_room_recording(self, request, pk=None):  # pylint: disable=unused-argu
...
             worker_manager.start(recording)
         except RecordingStartError:
+            # Move out of the unique-constraint window so the room is not
+            # locked from future recording attempts by this orphan row.
+            models.Recording.objects.filter(pk=recording.pk).update(
+                status=models.RecordingStatusChoices.FAILED_TO_START
+            )
             return drf_response.Response(
                 {"error": f"Recording failed to start for room {room.slug}"},
                 status=drf_status.HTTP_500_INTERNAL_SERVER_ERROR,
```
lebaudantoine: We probably should return a 502 here. These errors usually happen when the egress worker is busy, and a retry shortly afterward would likely succeed.

PR #1321, opened 2026-05-07
⬆️(mail) fix dependencies not having resolved or integrity field by updating

soyouzpanda:
This is needed for packaging as many distribution retrieve node modules into the npm cache and then tries to install node modules into the project without any internet connection. Since there is no resolved/integrity field, it fails to get packages from the cache.


soyouzpanda:
Maybe it would be nicer to use docs' modification of the mail template system (mjml v5 and new fonts).

But it may require a common base (maybe another repo like `django-lasuite`) so modifications from one repo are propagated to other lasuite projects.

lebaudantoine (approved):

lebaudantoine:
@qbey @lunika what do you think of @soyouzpanda's idea?

PR #1323, opened 2026-05-07
Improve `resolve_speaker_identities`  performance

cameledev:
## Purpose

Currently, we get diarization from whisperX, which returns generic speakers (SPEAKER_0, SPEAKER_1). We then assign actual users using a heuristic `resolve_speaker_identities` which tries to match whisperX output with VAD metadata, computing overlap. 

The problem is that the output of whisperX assigns wrong start/end values when words are followed by silence:

```
# Word in reality takes about 0.5s to be said. WhisperX sais it took about ~4s
{'word': 'pense', 'start': 3.989, 'end': 7.769, 'score': 0.963, 'speaker': 'SPEAKER_00'}
````

## Proposal

Add heuristic on speaker timeline to cut intervals at MAX_WORD_DURATION after any word start. Cut interval short if long word is at the end, or split into multiple intervals if long word(s) is in the middle.

## TODO
- [x] Heuristic for speaker timeline
- [x] New tests for heuristic
- [x] Update existing tests 
- [x] Debug logging to compare timelines

## NB: Output of debug logging

image

### DIFF
--- a/src/summary/summary/core/user_assign.py
+++ b/src/summary/summary/core/user_assign.py
@@ -14,11 +14,11 @@ from dataclasses import dataclass, field
 from datetime import datetime
 from typing import Any
 
-logger = logging.getLogger(__name__)
+from summary.core.config import get_settings
 
-# Minimum fraction of a speaker's total duration that must overlap with a
-# participant's VAD to accept the assignment.
-DEFAULT_OVERLAP_THRESHOLD = 0.5
+settings = get_settings()
+
+logger = logging.getLogger(__name__)
 
 
 @dataclass
@@ -74,7 +74,7 @@ class AssignmentResult:
                     f" ({item['speaker']})" if name_to_speaker_count[name] > 1 else ""
                 )  # Add suffix only if there are multiple detected speakers per user
                 return {**item, "speaker": f"{name}{suffix}"}
-            return item
+            return {**item}
 
         def _process_segment(
             item: dict[str, Any], include_words: bool = False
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: src/summary/summary/core/user_assign.py [hunk 1]
```diff
@@ -138,6 +141,78 @@ def _overlap_duration(
     return overlap
 
 
+def _format_timelines_debug(
```
cameledev, line 141: remove ?

lebaudantoine: Let's keep it until you have reached a stable version.

lebaudantoine (commented):
If I understand correctly, the main change here is that the speaker timeline is now reconstructed from word-level timestamps instead of segment boundaries.

The assumption is that any word lasting more than 1 second is likely either silence or a WhisperX hallucination, so the interval gets cut at that point to avoid artificially inflating the speaker’s speaking time, which would otherwise skew the overlap computation with participants.

I’m not entirely sure how to review this PR. My impression is that we’re introducing a correction in the WhisperX's output, but I’m having a hard time evaluating the concrete impact and how much this changes the behavior of your algorithm.

have you checked this module https://github.com/m-bain/whisperX/blob/main/whisperx/diarize.py?

From the tests, the scenarios you cover look reasonable to me. That said, I wonder whether the 1-second threshold might be a bit too aggressive: a long word spoken slowly (“approximately”, numbers read digit by digit, etc.) can legitimately exceed 1 second without being silence or a hallucination. In that case, we might end up truncating real speech and fragmenting the speaker timeline more than necessary. Do you have any intuition about the actual distribution of word durations in your samples?

In any case, these hardcoded values (MAX_WORD_DURATION, and while we’re at it DEFAULT_OVERLAP_THRESHOLD) should probably be exposed through settings so they can be adjusted without releasing a new version. This is exactly the kind of heuristic we’ll likely want to tune in production based on real-world observations.

Also you could enable/disable this correction using a backend setting.

IMO, we should quickly collect a small dataset of outputs (from both WhisperX and the metadata agent) in staging, so you can start iterating with real data and properly backtest the algorithm.

cameledev:
> If I understand correctly, the main change here is that the speaker timeline is now reconstructed from word-level timestamps instead of segment boundaries.

It use a mix of segment and word level timestamps. word level timestamps are only used to split up or trim segments when a word is too long. That being said, maybe just using word timestamps would be better. Ideally, we should time total interval duration with VAD and with both strategies, to choose the one that matches VAD best.

> The assumption is that any word lasting more than 1 second is likely either silence or a WhisperX hallucination, so the interval gets cut at that point to avoid artificially inflating the speaker’s speaking time, which would otherwise skew the overlap computation with participants.

Yes

> have you checked this module https://github.com/m-bain/whisperX/blob/main/whisperx/diarize.py?

Looked briefly but could not find anything relative to capping word duration

> From the tests, the scenarios you cover look reasonable to me. That said, I wonder whether the 1-second threshold might be a bit too aggressive: a long word spoken slowly (“approximately”, numbers read digit by digit, etc.) can legitimately exceed 1 second without being silence or a hallucination. In that case, we might end up truncating real speech and fragmenting the speaker timeline more than necessary. Do you have any intuition about the actual distribution of word durations in your samples?

This is mostly intuition, but the tradeoff seems good. Words over 1 second are infrequent (average words per minute (FR, EN) are between 110 and 170). On the other hand, whisperX potentially assigns wrong values at each word followed by a blank. 

> In any case, these hardcoded values (MAX_WORD_DURATION, and while we’re at it DEFAULT_OVERLAP_THRESHOLD) should probably be exposed through settings so they can be adjusted without releasing a new version. This is exactly the kind of heuristic we’ll likely want to tune in production based on real-world observations.

Agreed

> Also you could enable/disable this correction using a backend setting.

Agreed

lebaudantoine (approved):
LGTM, I've not tested it

PR #1328, opened 2026-05-11
🐛(summary) complete webm support

FloChehab:
When duration is not reported in the files metadata, we directly infer the duration from the audio packets. This prevents errors on webm files.

Very simple audio & video test files have been added that cover relevant usecases to prevent regressions.


FloChehab:
@coderabbitai full review

src/summary/summary/core/file_service.py
```diff
@@ -23,6 +23,93 @@
...
+            local_path,
+        ],
+        check=False,
+        stdout=subprocess.PIPE,
+        stderr=subprocess.PIPE,
+        text=True,  # Decode stdout/stderr as strings instead of bytes.
+    )
+
+    data = json.loads(result.stdout)
+    # Build a list containing the end time of each audio packet.
```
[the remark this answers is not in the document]
FloChehab: I think it's ok like this for now

.github/workflows/meet.yml
```diff
@@ -322,6 +322,11 @@ jobs:
       - name: Checkout repository
         uses: actions/checkout@v6
 
+      - name: Install ffmpeg
+        run: |
+          sudo apt-get update
+          sudo apt-get install -y ffmpeg
```
lebaudantoine, line 328: Should we make sure the same ffmpeg version is installed?
The Docker image pins it to `8.0.1-r1`

src/summary/summary/core/file_service.py
```diff
@@ -23,6 +23,93 @@
...
+
+    data = json.loads(result.stdout)
+    duration_value = data.get("format", {}).get("duration")
+
+    if duration_value in (None, "N/A"):
+        duration = _get_duration_from_packets(local_path)
+    else:
+        duration = float(duration_value)
+
+    return duration
```
lebaudantoine:
```suggestion
    if duration_value in (None, "N/A"):
        return _get_duration_from_packets(local_path)

    return float(duration_value)
```

src/summary/summary/core/file_service.py
```diff
@@ -23,6 +23,93 @@
...
+            local_path,
+        ],
+        check=False,
+        stdout=subprocess.PIPE,
+        stderr=subprocess.PIPE,
+        text=True,
+    )
+
+    data = json.loads(result.stdout)
+    duration_value = data.get("format", {}).get("duration")
```
lebaudantoine, line 109: As a fallback, you could try retrieving the file duration from the stream `duration` property itself.

I just tested locally with:

```bash
ffprobe -select_streams a:0 -show_streams ./audio-sample-ios.m4a -print_format json
```

and noticed that some streams also expose a `duration` field when available.

This obviously does not guarantee that a duration will always be present, but when it is, it could help us avoid falling back to the manual computation method.

src/summary/summary/core/file_service.py
```diff
@@ -23,6 +23,93 @@
...
+    #   end time = packet start time + packet duration
+    #
+    # pts_time is the packet presentation timestamp, meaning when that packet
+    # starts during playback.
+    #
+    # duration_time may be missing, so it defaults to 0.
+    packet_ends = [
+        float(packet["pts_time"]) + float(packet.get("duration_time", 0))
+        for packet in data.get("packets", [])
+        if "pts_time" in packet
```
lebaudantoine, line 76: I'm a bit concerned about the performance implications of this approach at scale, especially since this code acts as a gatekeeper and -show_packets requires iterating through all audio packets.

Could we explore a more lightweight fallback first?

For example, streams may expose enough metadata to compute the duration directly. If we can retrieve the sample rate together with the total number of samples (or a similar field), we could derive the duration with a simple computation instead of parsing every packet.

something like `duration_seconds = nb_samples / sample_rate`.

it's pure brainstorming.

.github/workflows/meet.yml
FloChehab: Yes we should, I'll check the feasability

FloChehab: I have checked and it's a bit tricky to do. there is this but it's a bit outdated https://github.com/federicocarboni/setup-ffmpeg
Apt repositories don't have all the versions.
My suggestion => leave it like this.

src/summary/summary/core/file_service.py
FloChehab: I have tweaked

src/summary/summary/core/file_service.py
FloChehab: So with VBR (variable bit rate), this is not reliable enough, or would require some scanning too.

I have tweaked the code to only fetch the last packets which avoids memory leaks and might be optimized internally by ffmpeg (not 100% sure though).

lebaudantoine (approved):

PR #1337, opened 2026-05-13
🔒️ (security) implement hash mode encryption

sneko:
## Purpose

First step to bring more privacy to meetings without requiring a specific user onboarding or a third-party server managing keys distribution.

The shared secret is passed within the URL hash so it's never sent to the `meet` server. The meeting creator is responsible to provide the link having the hash over secured channels (Tchap, ...). Knowing those meetings require the waiting room, so even if the link leaked across an email, the meeting admin still accept or not new participants.

Also:
* the PR also brings more visibility about users being authenticated or not
* encryption prevent SIP users and server features like recording or transcript

EDIT:
Due to first thoughts I had "default_encryption_mode" in the database to also prepare for potential advanced encryption mode, but then we switched to an additional step on the home screen to do "per-meeting" choice. Maybe I should change this for now, thoughts?


lebaudantoine:
@coderabbitai full review

PR #1343, opened 2026-05-14
🩹(frontend) fix subtitle background regression

lebaudantoine:
Restore transparent background as the default subtitle background to match previous behavior.


lebaudantoine:
@coderabbitai full review

Issue #1345, opened 2026-05-14
Automatically equalize participant audio levels


lebaudantoine:
## Feature Request

**Is your feature request related to a problem or unsupported use case? Please describe.**

In meetings with several participants, audio levels are often uneven, some people come through loud and clear, others are quiet or barely audible (poor mic, low input gain, distance from the mic). As a participant, this means constantly straining to hear some people while others are jarringly loud. 

The only fix today is manual: adjusting computer volume when the speaker changes. That's tedious, most users won't think to do it at all, they'll just accept a degraded experience.

**Describe the solution you'd like**

Automatic per-participant audio leveling (auto-gain / auto-normalization), so every participant lands at roughly the same perceived loudness without the user touching anything. 

The LiveKit JavaScript client exposes per-participant audio levels; a basic adaptive algorithm would use those to continuously adjust each remote participant's gain during the meeting, converging toward a consistent overall level.

Key product principles:
- Fully automatic — the user should never feel they *have* to manage participant volumes manually.
- Adaptive over the course of the meeting, not a one-time pass — it should react as people join, change environment, or move relative to their mic.
- "Basic and adaptive" by design — a simple, well-behaved algorithm is preferable to an aggressive one.
- Should be togglable in the settings.

**Describe alternatives you've considered**

- **Manual per-participant volume** — rejected as the primary solution because it puts the burden on the user and won't be discovered or used by most.
- **A single global output normalization** — simpler, but doesn't solve the core problem, which is *relative* differences *between* participants.
- **Asking participants to fix their own input gain** — not reliable; depends on every participant's awareness and setup.

**Discovery, Documentation, Adoption, Migration Strategy**

Ideally this "just works" with no discovery needed — participants simply notice the meeting sounds more balanced. If it's toggleable, expose it as a clearly labelled setting (e.g. "Automatically balance participant audio levels") with a one-line explanation. No migration concerns since it's additive; the main decision is the default state (on vs. off) for existing users.

It would be release off first, allowing user to try it in production, once it's validated, it will become on by default.

**Do you want to work on it through a Pull Request?**

kushmunjal:
Hi! I'd like to work on this issue. Can I be assigned to it?

Prateek007rai:
I raised PR for this issue, please look into it @lebaudantoine

PR #1347, opened 2026-05-14
🩹(frontend) add participant audio leveling

SyedIshmumAhnaf:
## Purpose

Adds an optional participant audio leveling setting for calls, addressing #1345.

## Proposal

Automatically equalize participant audio levels

## Changes

- Add a persisted audio setting for participant audio leveling.
- Add an Audio settings switch with localized label and description.
- Add a LiveKit hook that gradually attenuates loud remote microphone tracks.
- Reset touched remote audio tracks to volume 1 when disabled or unmounted.

## Safety

This MVP is attenuation-only. `MAX_GAIN` is capped at `1`, so the hook never sets `HTMLMediaElement.volume` above the safe browser limit.

### DIFF
--- a/src/frontend/src/features/rooms/livekit/hooks/useParticipantAudioLeveling.ts
+++ b/src/frontend/src/features/rooms/livekit/hooks/useParticipantAudioLeveling.ts
@@ -1,0 +1,67 @@
+import { useEffect, useRef } from 'react'
+import { RemoteAudioTrack, Track } from 'livekit-client'
+import { useRoomContext } from '@livekit/components-react'
+import { useSnapshot } from 'valtio'
+import { userPreferencesStore } from '@/stores/userPreferences'
+
+const TARGET_AUDIO_LEVEL = 0.35
+const MIN_GAIN = 0.6
+const MAX_GAIN = 1
+const SMOOTHING_FACTOR = 0.08
+const UPDATE_INTERVAL_MS = 500
+const MIN_AUDIO_LEVEL = 0.02
+
+export const useParticipantAudioLeveling = () => {
+  const room = useRoomContext()
+  const { is_participant_audio_leveling_enabled: enabled } =
+    useSnapshot(userPreferencesStore)
+
+  // track sid → current smoothed gain
+  const gainMapRef = useRef<Map<string, number>>(new Map())
+  // direct track object refs so reset works even if tracks are unpublished/muted
+  const touchedTracksRef = useRef<Set<RemoteAudioTrack>>(new Set())
+
+  useEffect(() => {
+    const resetTouchedTracks = () => {
+      for (const track of touchedTracksRef.current) {
+        track.setVolume(1)
+      }
+      gainMapRef.current.clear()
+      touchedTracksRef.current.clear()
+    }
+
+    if (!enabled) {
+      resetTouchedTracks()
+      return
+    }
+
+    const interval = setInterval(() => {
+      for (const participant of room.remoteParticipants.values()) {
+        const pub = participant.getTrackPublication(Track.Source.Microphone)
+        const audioTrack = pub?.audioTrack
+        if (!audioTrack || !pub.trackSid) continue
+        if (!(audioTrack instanceof RemoteAudioTrack)) continue
+
+        const sid = pub.trackSid
+        const audioLevel = participant.audioLevel
+
+        if (audioLevel < MIN_AUDIO_LEVEL) continue
+
+        const prevGain = gainMapRef.current.get(sid) ?? 1
+        const desiredGain = TARGET_AUDIO_LEVEL / audioLevel
+        const clampedGain = Math.min(MAX_GAIN, Math.max(MIN_GAIN, desiredGain))
+        const smoothedGain =
+          prevGain + SMOOTHING_FACTOR * (clampedGain - prevGain)
+
+        gainMapRef.current.set(sid, smoothedGain)
+        touchedTracksRef.current.add(audioTrack)
+        audioTrack.setVolume(smoothedGain)
+      }
+    }, UPDATE_INTERVAL_MS)
+
+    return () => {
+      clearInterval(interval)
+      resetTouchedTracks()
+    }
+  }, [enabled, room])
+}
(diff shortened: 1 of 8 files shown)


cameledev (commented):
Hi. Thank you for your PR. This would be nice feature. I did not try this out yet. 

I have a few comments/questions:
- Do you have a reference for property `audioLevel` being used ? I have trouble finding a definitive answer on whether this is a smoothed value already. I guess yes, since values are btw 0 and 1, but it would be nice to confirm what it actually corresponds to.
- Do you have insight for the proposed constants ? How did you choose these values ?
- SMOOTHING_FACTOR / UPDATE_INTERVAL_MS with current values means corrections can be long (5 seconds from 0.02 to 0.3, with target 0.35). While this is okay for volume increase, it seems like a long wait if sound is too loud. Perhaps, volume decrease can be sped up.

- Additional comment: We should be careful that this integrates properly with participant muting to come

SyedIshmumAhnaf:
@cameledev Thanks for the review.

On `audioLevel`: I traced the value through LiveKit rather than assuming from the range alone. In `livekit-client`, `participant.audioLevel` is assigned from `SpeakerInfo.level` during active-speaker / speaker-update handling. The LiveKit protocol defines `SpeakerInfo.level` as an audio level from `0` to `1.0`, where `1` is loudest.

I have not found a definitive reference confirming whether that value is already smoothed server-side, or whether it represents RMS/peak/VAD-derived level. So I treated it as a LiveKit-provided relative activity/loudness signal rather than a calibrated loudness value.

On the constants: they are conservative MVP values rather than calibrated audio-normalization constants.

* `MAX_GAIN = 1` is deliberate because this first version is attenuation-only. Since the current room setup does not enable a WebAudio amplification path, `RemoteAudioTrack.setVolume()` may map to `HTMLMediaElement.volume`, where values above `1` are unsafe.
* `MIN_GAIN = 0.6` avoids making loud participants too quiet.
* `MIN_AUDIO_LEVEL = 0.02` avoids reacting to silence/background noise.
* `TARGET_AUDIO_LEVEL = 0.35` is a conservative target within LiveKit’s `0..1` level range.

Your point about correction speed makes sense. I updated the smoothing to be asymmetric: attenuation is faster when a participant is too loud, while recovery is slower to avoid audible pumping. That should make loud-participant correction more responsive without making the volume jump around.

On participant muting: the current implementation only touches remote microphone track playback volume and resets touched `RemoteAudioTrack`s to `1` when disabled/unmounted. It does not change publication mute state or subscription state. That said, I agree this should be kept compatible with participant muting if/when local per-participant mute/volume controls are introduced; if that feature adds its own per-participant playback state, this hook should compose with it rather than overwrite it.

lebaudantoine:
@greptileai

SyedIshmumAhnaf:
@lebaudantoine based on the review, do you think this is ready to be merged? Do let me know if you need anything from my end.

src/frontend/src/locales/fr/settings.json
```diff
@@ -39,6 +35,11 @@
       "ongoingTest": "Test du son…",
       "safariWarning": "La sélection du haut-parleur n'est pas encore disponible sur Safari en raison de limitations du navigateur."
     },
+    "participantAudioLeveling": {
+      "heading": "Égalisation audio des participants",
+      "label": "Équilibrer automatiquement les niveaux audio des participants",
+      "description": "Ajuste progressivement le volume des participants distants pendant l'appel pour que les voix soient plus homogènes."
```
davd-gzl, line 45: Pour homogèniser les voix

src/frontend/src/features/rooms/livekit/hooks/useParticipantAudioLeveling.ts
```diff
@@ -0,0 +1,72 @@
...
+      gainMapRef.current.clear()
+      touchedTracksRef.current.clear()
+    }
+
+    if (!enabled) {
+      resetTouchedTracks()
+      return
+    }
+
+    const interval = setInterval(() => {
```
davd-gzl, line 61: Suggestion: [`ActiveSpeakersChanged`](https://github.com/livekit/client-sdk-js/blob/v2.17.1/src/room/Room.ts#L1743) fires whenever [livekit moves these levels](https://github.com/livekit/client-sdk-js/blob/v2.17.1/src/room/Room.ts#L1731), so a listener replaces this timer and nothing runs through a silent call.

davd-gzl (changes requested):
- This setting never sees its own effect: [what it changes](https://github.com/SyedIshmumAhnaf/meet/blob/feat-smart-audio/src/frontend/src/features/rooms/livekit/hooks/useParticipantAudioLeveling.ts#L63) is [the volume of your own audio element](https://github.com/livekit/client-sdk-js/blob/v2.17.1/src/room/track/RemoteAudioTrack.ts#L49), and [what it reads](https://github.com/SyedIshmumAhnaf/meet/blob/feat-smart-audio/src/frontend/src/features/rooms/livekit/hooks/useParticipantAudioLeveling.ts#L47) [comes from the server](https://github.com/livekit/client-sdk-js/blob/v2.17.1/src/room/Room.ts#L1731).
- Nobody is ever made louder, though [the label](https://github.com/SyedIshmumAhnaf/meet/blob/feat-smart-audio/src/frontend/src/locales/en/settings.json?plain=1#L40) promises balance: the gain stops at [`MAX_GAIN`](https://github.com/SyedIshmumAhnaf/meet/blob/feat-smart-audio/src/frontend/src/features/rooms/livekit/hooks/useParticipantAudioLeveling.ts#L9) 1, and 1 is [what a track already plays at](https://github.com/SyedIshmumAhnaf/meet/blob/feat-smart-audio/src/frontend/src/features/rooms/livekit/hooks/useParticipantAudioLeveling.ts#L28).
- Someone made quieter comes back only while they talk: [silence is skipped](https://github.com/SyedIshmumAhnaf/meet/blob/feat-smart-audio/src/frontend/src/features/rooms/livekit/hooks/useParticipantAudioLeveling.ts#L49), and [the climb back is slow](https://github.com/SyedIshmumAhnaf/meet/blob/feat-smart-audio/src/frontend/src/features/rooms/livekit/hooks/useParticipantAudioLeveling.ts#L11) even then.

The volume applied over one call, flat through the silence in the middle:

the volume falls while Loud speaks, holds flat with no level reported, then climbs back once they speak again

PR #1349, opened 2026-05-15
(paas) remove buildpack requirements.txt to use the new uv.lock

zeylos:
Current version depended on poetry which wasn't handled by buildpack. Now it throws an error:
```
 !     Error: Multiple Python package manager files were found.
 !
 !     Exactly one package manager file should be present in your app's
 !     source code, however, several were found:
 !
 !     requirements.txt (pip)
 !     uv.lock (uv)
```

This fixes it.

Tested successfully on scalingo


lebaudantoine (approved):
I've tested it trust u

zeylos:
hi @lebaudantoine can we merge that ?

Issue #1344, opened 2026-05-14
Per-recording encoding configuration (quality presets in the start-recording API)


lebaudantoine:
**Type:** Enhancement · **Component:** Recording / `start-recording` API

### Why

Users can record at high quality, producing very large files — a one-hour meeting at LiveKit's default `H264_720P_30` preset is ~1.4 GB, heavier than necessary for talking-head and screen-share content. A recent change added server-wide `RECORDING_ENCODING_*` settings to globally lower quality, but that's an operator-level, all-or-nothing knob. **Users should be able to choose an appropriate quality for their specific recording** at start time — a slide-heavy deck review and a voice-first webinar have very different needs.

### What

Extend the `start-recording` API to accept an optional per-recording **encoding** config, validated as part of the existing `RecordingOptions` pydantic model and persisted in `recording.options` — no new model field, **no migration**.

Use a **hybrid model**:

- **Resolution** — explicit dropdown value (`1080p` / `720p` / `540p`) → maps to `EncodingOptions.width` / `height`.
- **Profile / intent** — an abstracted level rather than raw numbers, surfaced to users as plain-language intent, mapping to the documented reference profiles:
  - "Optimize for small file / talking heads" → `low_cpu` (15 fps, 600 kbps video, 64 kbps audio)
  - "Optimize for on-screen text" → `slide_heavy` (15 fps, 900 kbps video, 64 kbps audio)
  - "Balanced" → `balanced` (20 fps, 1000 kbps video, 96 kbps audio)

Raw framerate / bitrate / keyframe values are **not** exposed in the API. The profile name resolves server-side to concrete values, which populate LiveKit's `EncodingOptions` message (`width`, `height`, `framerate`, `video_bitrate`, `audio_bitrate`, `key_frame_interval`) passed as `advanced` on the egress request.

**Extensibility:** the resolution and profile maps must be defined as plain data structures (dicts/enums) so adding a new profile or resolution is a one-line addition with no logic changes. Treat the profile registry as the single extension point.

### Validation

Extend `RecordingOptions` with an optional `encoding` sub-model. Because `model_config = {"extra": "forbid"}` is already set, unknown keys are rejected automatically. Resolution and profile fields should be `Literal` types so only allowlisted values pass — invalid values fail validation at the serializer (`SchemaField`) and return `400`.

```python
class EncodingConfig(BaseModel):
    resolution: Literal["1080p", "720p", "540p"] | None = None
[...]

PR #1357, opened 2026-05-20
✨(backend) add per-recording encoding config to start-recording API

sarthakbahal:
## Purpose
Adds optional per-recording encoding configuration to the `start-recording` API, addressing #1344.

## Proposal
Allow callers to pass resolution and quality profile settings when starting a recording,
instead of relying solely on global server-side encoding environment variables.

## Changes
- Add `EncodingConfig` Pydantic sub-model with `resolution` and `profile` fields.
- Extend `RecordingOptions` with an optional `encoding: EncodingConfig | None` field.
- Add `encodings.py` with `RESOLUTION_MAP`, `PROFILE_MAP`, and resolver functions.
- Wire encoding options into `VideoCompositeEgressService` via LiveKit `advanced=` parameter.
- Apply audio bitrate from profile in `AudioCompositeEgressService`, resolution is ignored for audio-only recordings.

## Notes
No database migration needed — encoding config is stored in the existing `recording.options` JSON field.
Precedence chain: per-recording config → global `RECORDING_ENCODING_*` env vars → LiveKit default.


src/backend/core/recording/encodings.py
```diff
@@ -0,0 +1,80 @@
...
+    "540p": (960, 540),
+}
+
+# Maps profile string → (framerate fps, video_kbps, audio_kbps)
+# Values come from the reference profiles table in docs/features/recording.md
+PROFILE_MAP = {
+    "low_cpu": (15, 600, 64),
+    "slide_heavy": (15, 900, 64),
+    "balanced": (20, 1000, 96),
+}
```
cameledev: I believe this should go into settings to allow modifying this through ENV variables

src/backend/core/recording/worker/services.py
```diff
@@ -160,10 +169,24 @@ def start(self, room_name, recording_id):
...
+        request_kwargs = {
+            "room_name": room_name,
+            "file_outputs": [file_output],
+            "audio_only": True,
+        }
+        if encoding_options is not None:
+            request_kwargs["advanced"] = encoding_options
+
+        request = livekit_api.RoomCompositeEgressRequest(**request_kwargs)
+
```
lebaudantoine, line 245: Reading through the code, this looks promising overall. 

However, I think these recording-related options would fit better in `VideoCompositeEgress`, since that service is actually responsible for starting and managing the screen recording pipeline. It’s also the component that most needs the CPU and file-size optimizations, so keeping those concerns localized there would make the design cleaner and easier to maintain.

I think your current approach for `AudioCompositeEgress` is still fine.

src/backend/core/recording/encodings.py
```diff
@@ -0,0 +1,80 @@
...
+    only the profile's audio bitrate is applied.
+    """
+    if encoding_config is None:
+        return None
+
+    if encoding_config.profile:
+        _, _, audio_kbps = PROFILE_MAP[encoding_config.profile]
+    else:
+        _, _, audio_kbps = PROFILE_MAP["balanced"]
+
```
lebaudantoine: Let me think, but these functions could be defined as private methods of the BaseEgress Service.

src/backend/core/recording/encodings.py
```diff
@@ -0,0 +1,80 @@
...
+    if encoding_config.resolution:
+        width, height = RESOLUTION_MAP[encoding_config.resolution]
+    else:
+        width, height = RESOLUTION_MAP["720p"]
+
+    # Profile: use from config or fall back to balanced
+    if encoding_config.profile:
+        fps, video_kbps, audio_kbps = PROFILE_MAP[encoding_config.profile]
+    else:
+        fps, video_kbps, audio_kbps = PROFILE_MAP["balanced"]
```
lebaudantoine: I would just pass None value as fallback to LiveKit default value in EncodingOptions.
Every parameter in EncodingOptions seems optional.

lebaudantoine (commented):
I haven't reviewed the test yet. Thank you for your contribution.

cameledev (commented):
Hi. Thank you for your contribution !

Could you adapt your solution to include this fix, adapting bitrate to resolution ?

https://github.com/suitenumerique/meet/issues/1344#issuecomment-4565781660

Also, I believe better naming for profiles would be:

 `talking_heads`, `text`, `high_quality`

sarthakbahal:
Sure , I'll work on the necessary solutions .

sarthakbahal:
Hi, I've pushed the updated changes addressing the review feedback. The summary is : 

1. Replaced flat PROFILE_MAP with a 2D ENCODING_MAP keyed by (resolution, profile) so bitrate scales correctly with resolution as suggested
2. Renamed profiles to talking_heads, text, mixed
3. Moved resolver functions into BaseEgressService as private methods _resolve_encoding_options and                                     _resolve_audio_encoding_options
4. Removed hardcoded fallbacks , passing None defers to LiveKit defaults (confirmed via tests that LiveKit protobuf converts None to 0 i.e. unset)

All 13 resolver tests and 56 API tests pass.

src/backend/core/recording/encodings.py
sarthakbahal: Moved to _resolve_encoding_options and _resolve_audio_encoding_options as private methods on BaseEgressService.

src/backend/core/recording/encodings.py
sarthakbahal: Removed the hardcoded fallbacks. Interestingly, tested and found LiveKit protobuf converts None to 0 internally, so unset fields correctly go over to LiveKit's defaults.

src/backend/core/recording/encodings.py
sarthakbahal: Updated to a 2D, ENCODING_MAP keyed by (resolution, profile) and updated the old naming profile to talking_heads, text, high_quality. Also , my hesitation is that ENCODING_MAP is tightly coupled to the Literal choices in EncodingConfig, so changing one without the other would break things.

PR #1358, opened 2026-05-21
✨(summary) extended support for all video / audio files

FloChehab:
* Removed constraint on file extension
* Infer audio / video streams from the media with ffmpeg
* Infer the correct processed audio file extension based on actual codec to avoid ffmpeg errors.

### DIFF
--- a/src/summary/summary/core/celery_worker.py
+++ b/src/summary/summary/core/celery_worker.py
@@ -14,7 +14,11 @@ from requests import exceptions
 
 from summary.core.analytics import MetadataManager, get_analytics
 from summary.core.config import get_settings
-from summary.core.file_service import FileService, FileServiceException
+from summary.core.file_service import (
+    FileService,
+    FileServiceException,
+    TranscribeError,
+)
 from summary.core.llm_service import LLMException, LLMObservability, LLMService
 from summary.core.locales import get_locale
 from summary.core.models import (
@@ -540,14 +544,24 @@ def process_audio_transcribe_v2_task(
 
     job_id = self.request.id
 
-    transcription_res = WhisperXResponse(
-        **transcribe_audio(  # type: ignore
-            task_id=job_id,
-            cloud_storage_url=payload.cloud_storage_url,
-            language=payload.language,
-            raises=True,
-        ).model_dump()
-    )
+    try:
+        transcription_res = WhisperXResponse(
+            **transcribe_audio(  # type: ignore
+                task_id=job_id,
+                cloud_storage_url=payload.cloud_storage_url,
+                language=payload.language,
+                raises=True,
+            ).model_dump()
+        )
+    except TranscribeError as e:
+        failure_payload = TranscribeWebhookFailurePayload(
+            job_id=job_id,
+            error_code=e.error_code,
+        )
+        call_webhook_v2_task.apply_async(
+            args=[failure_payload.model_dump(), payload.tenant_id]
+        )
+        return failure_payload.model_dump()
 
     file_service.store_transcript(
         transcript=transcription_res,
--- a/src/summary/summary/core/file_service.py
+++ b/src/summary/summary/core/file_service.py
@@ -7,6 +7,7 @@ import os
 import subprocess
 import tempfile
 from contextlib import contextmanager
+from dataclasses import dataclass
 from datetime import timedelta
 from pathlib import Path
 from urllib.parse import urlparse
@@ -21,6 +22,24 @@ from summary.core.shared_models import WhisperXResponse
 settings = get_settings()
 
 logger = logging.getLogger(__name__)
+
+
+class TranscribeError(ValueError):
+    """Base class for transcribe-related errors."""
+
+    error_code: str = "unknown_error"
+
+
+class MediaDurationTooLongError(TranscribeError):
+    """Raised when the duration of a media file exceeds the allowed limit."""
+
+    error_code = "media_duration_too_long"
+
+
+class NoAudioInFileError(TranscribeError):
+    """Raised when a media file does not contain any audio."""
+
+    error_code = "no_audio_in_file"
 
 
 def _get_duration_from_packets(local_path: Path) -> float:
@@ -84,7 +103,7 @@ def _get_duration_from_packets(local_path: Path) -> float:
     return max(packet_ends)
 
 
-def get_media_duration(local_path: Path):
+def get_media_duration_seconds(local_path: Path) -> float:
     """Get media (audio or video) file duration in seconds."""
     # ruff: noqa: S607 Hard to know the ffprobe path, it depends on the deployment
     result = subprocess.run(
@@ -114,6 +133,123 @@ def get_media_duration(local_path: Path):
     return _get_duration_from_packets(local_path)
 
 
+@dataclass(frozen=True)
+class MediaInfo:
+    """Object containing information about the media file."""
+
+    path: Path
+    has_audio: bool
+    has_video: bool
+    audio_duration_seconds: float | None
+    audio_codec_name: str | None
+
+
+def get_media_info(local_path: Path) -> MediaInfo:
+    """Determines if a media file contains an audio and / or a video stream.
+
+    This function checks if the given media file contains at least one
+    audio / video stream. It utilizes ffmpeg to analyze the media and parse
+    its metadata to detect the presence of audio / video content.
+    If appropriate, it also retrieves the codec name for the audio stream
+    and its duration.
+    """
+    # ruff: noqa: S607 Hard to know the ffprobe path, it depends on the deployment
+    # ruff: noqa: S603 Input can be trusted
+    result = subprocess.run(
+        [
+            "ffprobe",
+            "-v",
+            "error",
+            "-show_entries",
+            "stream=codec_type,codec_name",
+            "-print_format",
+            "json",
+            str(local_path),
+        ],
+        check=False,
+        stdout=subprocess.PIPE,
+        stderr=subprocess.PIPE,
+        text=True,
+    )
+    data = json.loads(result.stdout)
+
+    streams = data.get("streams", [])
+    has_audio = any(el["codec_type"] == "audio" for el in streams)
+    has_video = any(el["codec_type"] == "video" for el in streams)
+    audio_codec_name = next(
+        (
+            stream.get("codec_name")
+            for stream in streams
+            if stream.get("codec_type") == "audio"
+        ),
+        None,
+    )
+    audio_duration_seconds = None
+    if has_audio:
+        audio_duration_seconds = get_media_duration_seconds(local_path)
+    return MediaInfo(
+        path=local_path,
+        has_audio=has_audio,
+        has_video=has_video,
+        audio_duration_seconds=audio_duration_seconds,
+        audio_codec_name=audio_codec_name,
+    )
+
+
+def extract_audio_from_video(media_info: MediaInfo) -> Path:
+    """Extracts the audio track from a video file and saves it as a separate audio file.
+
+    Based on the provided audio codec,
+    it determines an appropriate audio file extension.
+    It then runs
+    `ffmpeg` to extract the audio track without re-encoding it, ensuring quality is
+    preserved and speed.
+
+    Raises:
+        CalledProcessError: Raised if the `ffmpeg` command execution fails.
+    """
+    if media_info.audio_codec_name is None:
+        raise ValueError("Media file must have codec name must be provided")
+
+    # .mka can contain a bunch of audio codecs, using it as a default
+    # to avoid potential issues with the ffmpeg copy later on
+    suffix = settings.codec_to_extension.get(media_info.audio_codec_name, ".mka")
+
+    with tempfile.NamedTemporaryFile(
+        suffix=suffix,
+        delete=False,
+        prefix="audio_extract_",
+    ) as tmp:
+        output_path = Path(tmp.name)
+
+    extract_command = [
+        "ffmpeg",
+        "-i",
+        str(media_info.path),
+        "-vn",
+        "-acodec",
+        "copy",
+        "-y",
+        str(output_path),
+    ]
+
+    try:
+        subprocess.run(extract_command, check=True)
+    except BaseException as e:
+        if isinstance(e, FileNotFoundError):
+            logger.error("ffmpeg not found. Please install ffmpeg.")
+        elif isinstance(e, subprocess.CalledProcessError):
+            logger.error("Audio extraction failed: %s", e.stderr.decode())
+        else:
+            logger.error("Unexpected error during audio extraction: %s", e)
+
+        if output_path.exists():
+            os.remove(output_path)
+        raise RuntimeError("Failed to extract audio from file") from e
+
+    return output_path
+
+
 class FileServiceException(Exception):
     """Base exception for file service operations."""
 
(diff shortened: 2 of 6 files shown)


src/summary/summary/core/shared_models.py
```diff
@@ -89,9 +89,9 @@ class TranscribeWebhookFailurePayload(BaseWebhook):
 
     type: Literal["transcript"] = Field(default="transcript")
     status: Literal["failure"] = Field(default="failure")
-    error_code: Literal["unknown_error"] = Field(
-        title="Error code", description="The error code."
-    )
+    error_code: (
+        Literal["unknown_error", "no_audio_in_file", "media_duration_too_long"] | str
+    ) = Field(title="Error code", description="The error code.")
```
FloChehab, line 96: This is technically a breaking change, but it impacts only Transcripts and I have made the change.

src/summary/summary/core/file_service.py
```diff
@@ -114,6 +123,154 @@ def get_media_duration(local_path: Path):
     return _get_duration_from_packets(local_path)
 
 
+@dataclass(frozen=True)
+class MediaInfo:
+    """Object containing information about the media file."""
+
+    has_audio: bool
+    has_video: bool
+    audio_duration_seconds: float | None
```
cameledev, line 143:
```suggestion
    media_duration_seconds: float | None
```

for consistency

src/summary/summary/core/file_service.py
```diff
@@ -114,6 +123,154 @@ def get_media_duration(local_path: Path):
...
+
+
+def get_media_info(local_path: Path) -> MediaInfo:
+    """Determines whether a media file contains an audio and / or a video stream.
```
cameledev:
```suggestion
    """Determines if a media file contains an audio and / or a video stream.
```

src/summary/summary/core/file_service.py
```diff
@@ -114,6 +123,154 @@ def get_media_duration(local_path: Path):
...
+    audio / video stream. It utilizes ffmpeg to analyze the media and parse
+    its metadata to detect the presence of audio / video content.
+
+    Parameters:
+    local_path : Path
+        The filesystem path to the media file being analyzed.
+
+    Returns:
+    MediaInfo
+        Object containing information about the media file.
```
cameledev:
```suggestion
    Args:
        local_path : The filesystem path to the media file being analyzed.

    Returns:
        MediaInfo:  Object containing information about the media file.
```

to follow existing pattern

src/summary/summary/core/file_service.py
```diff
@@ -114,6 +123,154 @@ def get_media_duration(local_path: Path):
...
+
+    The function uses `ffprobe` to identify the audio codec used in the
+    input video file.
+    Based on the codec, it determines an appropriate audio file extension. It then runs
+    `ffmpeg` to extract the audio track without re-encoding it, ensuring quality is
+    preserved and speed.
+
+    Parameters:
+        video_path (Path): The path to the input video file from which the audio track
+            will be extracted.
```
cameledev: Same. Use "Args" , and no need to specify var type since variables are already typed :)

src/summary/summary/core/file_service.py
```diff
@@ -114,6 +123,154 @@ def get_media_duration(local_path: Path):
...
+        "json",
+        str(video_path),
+    ]
+
+    probe = subprocess.run(
+        probe_cmd,
+        capture_output=True,
+        text=True,
+        check=True,
+    )
```
cameledev: If I understand properly, we will be running ffprobe twice (with codec_type then codec_name) on the video. Why not run with codec_name, store it in `MediaInfo` and re-use it here ?

src/summary/summary/core/file_service.py
```diff
@@ -114,6 +123,154 @@ def get_media_duration(local_path: Path):
...
+    suffix = extension_map.get(codec, ".mka")
+
+    with tempfile.NamedTemporaryFile(
+        suffix=suffix,
+        delete=False,
+        prefix="audio_extract_",
+    ) as tmp:
+        output_path = Path(tmp.name)
+
+    command = [
```
cameledev:
```suggestion
    audio_extract_cmd = [
```
(similar to probe_cmd  above)

src/summary/summary/core/shared_models.py
```diff
@@ -126,7 +126,7 @@ class SummarizeWebhookFailurePayload(BaseWebhook):
 
     type: Literal["summary"] = Field(default="summary")
     status: Literal["failure"] = Field(default="failure")
-    error_code: Literal["unknown_error"] = Field(
+    error_code: Literal["unknown_error"] | str = Field(
```
cameledev, line 133: is there a reason to change this ?

cameledev (commented):
Looks mostly good to me.

One minor structure question/suggestion.

Otherwise, just naming stuff

src/summary/summary/core/celery_worker.py
```diff
@@ -540,14 +545,33 @@ def process_audio_transcribe_v2_task(
...
-    )
+    try:
+        transcription_res = WhisperXResponse(
+            **transcribe_audio(  # type: ignore
+                task_id=job_id,
+                cloud_storage_url=payload.cloud_storage_url,
+                language=payload.language,
+                raises=True,
+            ).model_dump()
+        )
```
lebaudantoine, line 555: not related with this PR, can't `transcribe_audio` returns directly a `WhisperXResponse` instance?

src/summary/summary/core/shared_models.py
lebaudantoine: Would you prefer to explicitly list the allowed error_code?

src/summary/summary/core/celery_worker.py
```diff
@@ -540,14 +545,33 @@ def process_audio_transcribe_v2_task(
...
+        return failure_payload.model_dump()
+    except MediaDurationTooLongError:
+        failure_payload = TranscribeWebhookFailurePayload(
+            job_id=job_id,
+            error_code="media_duration_too_long",
+        )
+        call_webhook_v2_task.apply_async(
+            args=[failure_payload.model_dump(), payload.tenant_id]
+        )
+        return failure_payload.model_dump()
```
lebaudantoine: i would go for smth : 

```
class TranscribeError(ValueError):
    error_code: str = "unknown_error"

class MediaDurationTooLongError(TranscribeError):
    """Raised when the duration of a media file exceeds the allowed limit."""
    error_code = "media_duration_too_long"

class NoAudioInFileError(TranscribeError):
    """Raised when a media file does not contain any audio."""
    error_code = "no_audio_in_file"
```

and then here : 

```
 except TranscribeError as exc:
        failure_payload = TranscribeWebhookFailurePayload(
            job_id=job_id,
            error_code=exc.error_code,
        )
        call_webhook_v2_task.apply_async(
            args=[failure_payload.model_dump(), payload.tenant_id]
        )
        return failure_payload.model_dump()
```

src/summary/summary/core/file_service.py
```diff
@@ -114,6 +123,154 @@ def get_media_duration(local_path: Path):
...
+
+    # Choose proper extension
+    extension_map = {
+        "aac": ".m4a",
+        "alac": ".m4a",
+        "mp3": ".mp3",
+        "opus": ".opus",
+        "vorbis": ".ogg",
+        "flac": ".flac",
+    }
```
lebaudantoine: This should be configurable through settings with a default value, allowing us to add new mappings via environment variables without needing to release a new version.

src/summary/summary/core/file_service.py
```diff
@@ -114,6 +123,154 @@ def get_media_duration(local_path: Path):
...
+
+    suffix = extension_map.get(codec, ".mka")
```
lebaudantoine: why `mka` is the new default one?

src/summary/summary/core/file_service.py
lebaudantoine: nit: 
```suggestion
    codec_to_extension = {
        "aac": ".m4a",
        "alac": ".m4a",
        "mp3": ".mp3",
        "opus": ".opus",
        "vorbis": ".ogg",
        "flac": ".flac",
    }
```

src/summary/summary/core/shared_models.py
FloChehab: Yes, so that if we add other error_codes in the future this doesn't cause a breaking change, from the consumer standpoint.

FloChehab:
> Looks mostly good to me.
> 
> One minor structure question/suggestion.
> 
> Otherwise, just naming stuff

Thanks for the review!
I'll try to clean things by the end of day :)

src/summary/summary/core/file_service.py
```diff
@@ -114,6 +123,154 @@ def get_media_duration(local_path: Path):
...
+        if isinstance(e, FileNotFoundError):
+            logger.error("ffmpeg not found. Please install ffmpeg.")
+        elif isinstance(e, subprocess.CalledProcessError):
+            logger.error("Audio extraction failed: %s", e.stderr.decode())
+        else:
+            logger.error("Unexpected error during audio extraction: %s", e)
+
+        if output_path.exists():
+            os.remove(output_path)
+        raise RuntimeError("Failed to extract audio from file") from e
```
lebaudantoine, line 248: in terms of python style I found the initial syntax with several except easier to scan/parse.
maybe, extract the cleanup in a function?

src/summary/summary/core/file_service.py
```diff
@@ -114,6 +123,154 @@ def get_media_duration(local_path: Path):
...
+
+    probe = subprocess.run(
+        probe_cmd,
+        capture_output=True,
+        text=True,
+        check=True,
+    )
+
+    data = json.loads(probe.stdout)
+    codec = data["streams"][0]["codec_name"]
```
lebaudantoine: extracting this code in a method with an explicit name could help code reading (e.g. `extract_codec`)

src/summary/summary/core/file_service.py
```diff
@@ -244,63 +391,19 @@ def _download_from_cloud_storage_url(self, cloud_storage_url: str) -> Path:
                 "Unexpected error while downloading object from cloud_storage_url."
             ) from e
 
-    def _validate_duration(self, local_path: Path) -> float:
+    def _validate_duration(self, duration_seconds: float) -> None:
```
lebaudantoine, line 373: we could have isolate the renaming `duration` to `duration_seconds` in a commit to facilitate the review.

src/summary/summary/core/file_service.py
```diff
@@ -244,63 +391,19 @@ def _download_from_cloud_storage_url(self, cloud_storage_url: str) -> Path:
...
-        duration = get_media_duration(local_path)
-
         logger.info(
             "Recording file duration: %.2f seconds",
-            duration,
+            duration_seconds,
         )
 
-        if self._max_duration is not None and duration > self._max_duration:
+        if self._max_duration is not None and duration_seconds > self._max_duration:
```
lebaudantoine: should we rename `_max_duration` to `_max_duration_seconds` to be consistent? Btw not related with the initial PR intent.

src/summary/summary/core/file_service.py
lebaudantoine: agree that would save a call. Let's reuse the MediaInfo

lebaudantoine (commented):
I think the current structure evolved in a way that makes the file service harder to extend safely.

My initial choice was to encapsulate file utilities inside a service class to support interchangeability of implementations. However, in practice we’ve ended up continuously modifying the same implementation for new features, which increases regression risk for v1 and reduces flexibility.

In hindsight, it may have been better to introduce a stable interface with separate implementations (e.g. v1/v2) and a factory or dependency injection layer to select between them, instead of evolving a single service over time.

We already adapt the behavior of the service with the `if`logic base on the file url/path.

This would likely have preserved v1 behavior while allowing faster iteration on v2 without affecting existing logic.

We can’t assume all summary callers will require the same file service implementation. This suggests we may need a more explicit abstraction (e.g. interface + multiple implementations) rather than a single tightly coupled service.

As a general rule, I think we should try to keep clear boundaries between implementations rather than mixing extension logic into a single evolving class. 

I think we should clarify the ownership boundaries between module-level utilities and service classes. Right now the split is not always consistent, which makes it harder to understand where new logic should live.

src/summary/summary/core/shared_models.py
FloChehab: Sorry, was responding to @cameledev. 
An explicit list of of error codes is nice, BUT, if you add one it could be treated as breaking given the parsing done by pydantic schemas (those schemas might be copied in other repos). So I prefer to be explicit and actualy authorize any other unkown string, so that other error code can be added later on without breaking the API.

FloChehab:
Regarding your big comment on the file service @lebaudantoine, in https://github.com/suitenumerique/meet/pull/1362 I get rid of old "read from minio" and other stuff from the file service. From my PoV summary was considered kind of experimental in regards to other hosters community (there is a bunch of forced posthog in config for instance).
So I think we are still ok in regards to merging those changes.

src/summary/summary/core/celery_worker.py
FloChehab: There is indeed work to be done around using from pydantic models, especially in the helpers by @cameledev. I left that for another PR.

src/summary/summary/core/celery_worker.py
FloChehab: Approved, much nicier :)

src/summary/summary/core/file_service.py
FloChehab: .mka: Used for audio only files, can contain any supported audio compression format, such as MP2, MP3, Vorbis, AAC, AC3, DTS, or PCM

From https://www.matroska.org/faq.html
So basically it's a good default to avoid ffmpeg complain. 
I have added a comment about that aspect.

src/summary/summary/core/file_service.py
FloChehab: True, I tried to be as explicit as possible, should have been in another PR.
To align I have renamed to _max_duration_seconds, but will leave the env parameter renaming to another RP.

src/summary/summary/core/file_service.py
FloChehab: It really is the audio duration, if a video without audio is "provided", then the audio_duration_seconds is None. So lgtm.

src/summary/summary/core/file_service.py
FloChehab: I have cleaned the code a bit

src/summary/summary/core/file_service.py
FloChehab: Merged into MediaInfo

src/summary/summary/core/file_service.py
FloChehab: I think it's better now that it's moved in the get_media_info

src/summary/summary/core/file_service.py
FloChehab: I think it's very theoretical considered the current list but I have made the change.

src/summary/summary/core/file_service.py
FloChehab: I think this is to be reworked as a bigger clean of those file handles

src/summary/summary/core/file_service.py
lebaudantoine: I'm not sure to get what's theoretical, but thanks!

lebaudantoine (approved):

PR #1362, opened 2026-05-26
feat(summary) Migrate visio integration to summary API v2

FloChehab:
## Overview

This PR migrates the visio transcription integration from the legacy v1 summary API to the streamlined v2 API, cleaning up old code and introducing several improvements to how transcription, summarization, and analytics are handled.

### 🔥 Breaking / Cleanup
- Removed most v1 route-related code from the summary service (first step toward full v1 deprecation)
=> Not considered as a real breaking change as the deployment of both services simultaneously is considered stable, and visio is still compatible with a summary v1 API.

### ✨ Summary Service — v2 Route Extensions
- v2 `create transcribe` payload now supports publishing directly to Docs and triggering automatic summarization
- Title computation is now the **caller's responsibility**, simplifying the endpoint interface
- All files are now sent as **signed URLs** (direct S3 reads removed)
- Docs integration is now explicit in summary settings
- User `sub` is now used as the PostHog `distinct_id` for consistent cross-project analytics; a PostHog event is emitted at request creation time to ensure feature flags are properly evaluated
- User email can now be provided to the API (optional, non-breaking) -> usefull for analytics
- Added a specific User-Agent for improved tracking

### ✨ Visio — v2 Compatibility
- Visio now calls the v2 summary API; v1 params are still sent for backward compatibility, but **users running their own deployments are advised to migrate** — this compatibility shim will be removed in a future major version
- Title for the transcription document is now computed in visio
- `RecordingStatusChoices` has two new values: `EXTERNAL_PROCESS_SUCCESSFUL` and `EXTERNAL_PROCESS_FAILED`, updated via a new webhook called by the transcribe service
- The new webhook is protected by its own bearer token
- Tests added and updated to cover the above changes

## Notes
- Some existing helpers in summary still use `model_dump()` in places due to incomplete Pydantic model adoption — left as-is to avoid scope creep

## Integration tests
* Tested on docker-compose / tilt / staging

TODO : 

* Discuss posthog alias id & sub => Create a new posthog projet dedictated to summary that works on sub only


FloChehab:
@coderabbitai full review

src/helm/env.d/dev-dinum/values.meet.yaml.gotmpl
```diff
@@ -105,8 +105,9 @@ backend:
     RECORDING_ENABLE: True
     RECORDING_STORAGE_EVENT_ENABLE: True
     RECORDING_STORAGE_EVENT_TOKEN: password
-    SUMMARY_SERVICE_ENDPOINT: http://meet-summary:80/api/v1/tasks/
+    SUMMARY_SERVICE_ENDPOINT: http://meet-summary:80/api/v2/async-jobs/transcribe/
     SUMMARY_SERVICE_API_TOKEN: password
+    SUMMARY_SERVICE_WEBHOOK_API_TOKEN: webhook-password
```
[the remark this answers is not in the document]
FloChehab: dev-dinum is not maintained

src/backend/core/api/viewsets.py
```diff
@@ -989,6 +992,47 @@ def on_storage_event_received(self, request, pk=None):  # pylint: disable=unused
...
+        authentication_classes=[RecordingProcessWebhookAuthentication],
+    )
+    def on_external_process_event_received(self, request, pk=None):  # pylint: disable=unused-argument
+        """Handle incoming external process events for recordings."""
+        logger.debug("Processing external process event %s", request.data)
+        data = request.data
+        if not data.get("job_id"):
+            raise drf_exceptions.ValidationError(detail="No job_id provided")
+
+        job_id = data["job_id"]
```
lebaudantoine: wip: use a serializer? with a clear validation of what format should a job_id be

src/backend/core/models.py
```diff
@@ -598,6 +605,14 @@ class Recording(BaseModel):
...
         help_text=_("Recording options"),
     )
+    external_process_id = models.CharField(
+        max_length=255,
+        null=True,
+        blank=True,
+        unique=True,
+        verbose_name=_("External Process ID"),
+        help_text=_("ID of the external process associated with the recording."),
+    )
```
lebaudantoine, line 615: wip: CharField seems very open right now. Is that by design, or should we enforce a UUID format?

src/summary/summary/core/config.py
```diff
@@ -112,16 +110,17 @@ class Settings(BaseSettings):
...
     # Summary related settings
     is_summary_enabled: bool = True
-    transcription_satisfaction_form_base_url: Optional[str] = None
+
+    # Docs service configuration
+    is_docs_integration_enabled: bool = False
+    docs_base_url: Url = "https://example.com"
+    docs_server_to_server_api_key: SecretStr = Field(
+        title="API key for using docs server to server api", default="NO_API_KEY"
+    )
```
lebaudantoine, line 123: WIP: Docs is probably vague for external contributors; I would suggest being more precise, LaSuite Docs ?

src/summary/summary/core/config.py
```diff
@@ -191,13 +162,21 @@ def validate_authorized_tenants(self):
...
+                )
+            if self.docs_server_to_server_api_key.get_secret_value() in {
+                "",
+                "NO_API_KEY",
+            }:
+                raise ValueError(
+                    "Valid docs_server_to_server_api_key is required when"
+                    " docs integration is enabled"
+                )
         return self
```
lebaudantoine, line 182: nit: early return would save one nesting level

```
if not self.is_docs_integration_enabled:
   return self
```

src/summary/summary/core/config.py
```diff
@@ -112,16 +110,17 @@ class Settings(BaseSettings):
     webhook_max_retries: int = 2
     webhook_status_forcelist: List[int] = [502, 503, 504]
     webhook_backoff_factor: float = 0.1
-
-    # Locale
-    default_context_language: Literal["de", "en", "fr", "nl"] = "fr"
```
lebaudantoine, line 117: wip: why removing?

src/backend/meet/settings.py
```diff
@@ -755,6 +755,14 @@ class Base(Configuration):
     SUMMARY_SERVICE_API_TOKEN = SecretFileValue(
         None, environ_name="SUMMARY_SERVICE_API_TOKEN", environ_prefix=None
     )
+    SUMMARY_SERVICE_WEBHOOK_API_TOKEN = SecretFileValue(
+        None, environ_name="SUMMARY_SERVICE_WEBHOOK_API_TOKEN", environ_prefix=None
+    )
```
lebaudantoine, line 762: WIP: open question, since these two components already share a secret, would it make sense to reuse the same one here for simplicity? My intuition is that it’s probably not a good idea, as it would introduce tighter coupling and reduce separation of concerns.

But it would make configuration easier

src/backend/core/utils.py
```diff
@@ -465,6 +465,19 @@ def generate_download_file_url(file, *, expires_in: int, override_domain: bool =
...
+
+
+def generate_download_s3_file_url(
+    key: str, *, expires_in: int, override_domain: bool = True
+):
+    """
+    Generate a S3 signed download url for a given key.
+    """
+    if not key:
+        raise ValueError("key cannot be empty")
```
lebaudantoine: wip: naming `generate_download_s3_file_url` and `generate_download_file_url` are close, and we don't get easily what's the difference between them.

src/backend/core/recording/event/notification.py
```diff
@@ -178,6 +182,52 @@ def _ns_to_utc(ns):
...
+            return default_template_by_locale.get(
+                locale, default_template_by_locale["en-us"]
+            )
+
+        template_by_locale = {
+            "de-de": 'Besprechung "{room}" am {room_recording_date} um {room_recording_time}',
+            "en-us": 'Meeting "{room}" on {room_recording_date} at {room_recording_time}',
+            "fr-fr": 'Réunion "{room}" du {room_recording_date} à {room_recording_time}',
+            "nl-nl": 'Vergadering "{room}" op {room_recording_date} om {room_recording_time}',
+        }
```
lebaudantoine: wip: can't we rely on Django i18n for these parts? to avoid these manipulations

src/backend/core/recording/event/notification.py
```diff
@@ -214,19 +262,57 @@ def _notify_summary_service(recording: models.Recording):
...
             "owner_id": str(owner_access.user.id),
             "recording_filename": recording.key,
             "metadata_filename": metadata_filename,
             "email": owner_access.user.email,
             "sub": owner_access.user.sub,
             "room": recording.room.name,
-            "language": recording.options.get("language"),
             "owner_timezone": str(owner_access.user.timezone),
             "download_link": f"{get_recording_download_base_url()}/{recording.id}",
-            "context_language": owner_access.user.language,
```
lebaudantoine, line 227: WIP: Removing these params is inherently a breaking change, don't you think?

src/backend/core/recording/event/notification.py
```diff
@@ -214,19 +262,57 @@ def _notify_summary_service(recording: models.Recording):
...
             "recording_start_at": (started_at.isoformat() if started_at else None),
             "recording_end_at": (ended_at.isoformat() if ended_at else None),
+            # V2 params
+            "user_sub": owner_access.user.sub,
+            "user_email": owner_access.user.email,
+            "cloud_storage_url": generate_download_s3_file_url(
+                recording.key, expires_in=60 * 60 * 24, override_domain=False
```
lebaudantoine: Expiration should be configurable from a django settings

src/backend/core/recording/event/notification.py
```diff
@@ -214,19 +262,57 @@ def _notify_summary_service(recording: models.Recording):
...
+                    locale=owner_access.user.language
+                    or recording.options.get("language", "fr-fr"),
+                    room=recording.room.name,
+                    recording_datetime=started_at,
+                    owner_timezone=str(owner_access.user.timezone),
+                ),
+                "download_link": f"{get_recording_download_base_url()}/{recording.id}",
+                "form_link": f"{form_base_url}?room_id={recording.room.id}"
+                if (form_base_url and metadata_filename is not None)
+                else None,
```
lebaudantoine: If you’re introducing a temporary form_base_url variable, I’d suggest moving the ternary logic into its definition and instantiating form_link directly afterward. That would make the flow a bit clearer.

src/backend/core/recording/event/notification.py
```diff
@@ -214,19 +262,57 @@ def _notify_summary_service(recording: models.Recording):
...
+                # For now the feature flag logic is handled on summary side
+                "auto_create_summary": True,
+            },
+            "metadata": (
+                {
+                    "cloud_storage_url": generate_download_s3_file_url(
+                        metadata_filename,
+                        expires_in=60 * 60 * 24,
```
lebaudantoine: same I guess, should probably be extracted in a setting, and maybe factorized

src/backend/core/recording/event/notification.py
```diff
@@ -214,19 +262,57 @@ def _notify_summary_service(recording: models.Recording):
...
+                        override_domain=False,
+                    ),
+                    "started_at": started_at.isoformat(),
+                    "ended_at": ended_at.isoformat(),
+                }
+                if (started_at and ended_at and metadata_filename)
+                else None
```
lebaudantoine: I would extract this part in a variable

src/backend/core/recording/event/notification.py
```diff
@@ -242,6 +328,35 @@ def _notify_summary_service(recording: models.Recording):
...
+                    "transcribe service."
+                    "We have released a new version and API contract for that service, "
+                    "we recommend checking it out"
+                    # pylint: disable=line-too-long
+                    "https://github.com/suitenumerique/meet/blob/19c2a378e7e66652afbfa3e749badd697e3917ac/src/summary/summary/api/route/tasks_v2.py "
+                    "and mimicking it's behavior. We will remove the legacy "
+                    "handling in a future version.",
+                    DeprecationWarning,
+                    stacklevel=2,
+                )
```
lebaudantoine: nit: I would have inverted the boolean to : 

`if is_legacy_implementation` or `if is_v1_implementation`

src/summary/summary/core/models.py
```diff
@@ -12,6 +14,56 @@ class SharedV2TaskCreation(BaseModel):
...
+        description="The URL of the metadata file for speaker assignement.",
+    )
+    started_at: AwareDatetime = Field(title="Start time of the recording to transcribe")
+    ended_at: AwareDatetime = Field(title="End time of the recording to transcribe")
+
+
+class PushToDocsBaseConfig(BaseModel):
+    """Model containing information for pushing transcript and summaries to docs."""
+
+    user_email: str = Field(
```
lebaudantoine: should it be a string?

src/summary/summary/core/models.py
```diff
@@ -12,6 +14,56 @@ class SharedV2TaskCreation(BaseModel):
...
+        title="User Email", description="The user's email, future owner of the docs."
+    )
+    title: str = Field(title="Title", description="The title for the created document.")
+
+
+class PushToDocsTranscriptConfig(PushToDocsBaseConfig):
+    """Model for push to docs information for transcripts."""
+
+    download_link: str | None = Field(
```
lebaudantoine: same should it be url?

src/summary/summary/core/models.py
```diff
@@ -12,6 +14,56 @@ class SharedV2TaskCreation(BaseModel):
...
+        default=None,
+        title="Download Link",
+        description="The link to download the recording.",
+    )
+    form_link: str | None = Field(
```
lebaudantoine: same

src/backend/core/recording/event/notification.py
```diff
@@ -242,6 +328,35 @@ def _notify_summary_service(recording: models.Recording):
...
             response.raise_for_status()
+            is_v2_implementation = False
+            try:
+                response_json = response.json()
+                # We do not require a job_id to avoid a breaking change
+                if job_id := response_json.get("job_id"):
+                    recording.external_process_id = job_id
+                    recording.save()
+                    is_v2_implementation = True
+            except json.JSONDecodeError:
```
lebaudantoine: Should we make it explicit via Django settings whether we use v1 or v2?

Right now, checking for the presence of job_id to infer the version feels a bit weak, it’s implicit that v1 has no job_id, but not very clear.

I think we’d benefit from an explicit setting-based switch, calling either a dedicated v1 or v2 method. That would make the logic easier to follow and avoid sending requests with both v1 and v2 params mixed together.

It would also make the v1 deprecation path cleaner: we could simply remove/vendor the v1 method instead of refactoring this one later.

src/backend/core/recording/event/authentication.py
```diff
@@ -91,3 +91,46 @@ def authenticate(self, request):
...
+    def authenticate(self, request):
+        """
+        Authenticate the request and return a two-tuple of (user, token).
+        """
+        required_token = settings.SUMMARY_SERVICE_WEBHOOK_API_TOKEN
+        if not required_token:
+            raise AuthenticationFailed("Webhook authentication is not configured.")
+
+        auth_header: str = request.headers.get("Authorization") or ""
+        if not auth_header.startswith("Bearer "):
```
lebaudantoine: RFC 7235 says the scheme is case-insensitive

src/backend/core/models.py
FloChehab: I think we can be relaxed here, to avoid self holster being forced to used a specific uuid format, etc.

src/backend/core/api/viewsets.py
FloChehab: I think we can be relaxed here, to avoid self holster being forced to used a specific uuid format, etc.

src/summary/summary/core/config.py
FloChehab: Yeah we should find something better.

src/summary/summary/core/config.py
FloChehab: Default should be handled at the caller level from my pov

src/backend/meet/settings.py
FloChehab: I agree with ease of configuration. However that wouldn't be ideal.
Maybe this could be adapted with a single SUMMARY_SERVICE_TOKEN, that we use to sign the payload with on the summary side, and on visio side we can check that signed payload is signed with something we know :shrug: Not sure though and would require changes in Transcripts too.

src/backend/core/utils.py
FloChehab: yeah, I need to delete one of those;

src/backend/core/recording/event/notification.py
FloChehab: True, maybe, I am not familiar enough with that. I'll check.

src/backend/core/recording/event/notification.py
FloChehab: They are moved in the "V2" part, so not breaking from my pov.

FloChehab: (they are shared between v2 and v1)

src/summary/summary/core/models.py
FloChehab: What else are you thinking ? (like a proper email check?)

src/summary/summary/core/models.py
FloChehab: True

src/backend/core/recording/event/notification.py
FloChehab: True I'll like that idea

src/backend/core/recording/event/authentication.py
```diff
@@ -91,3 +91,46 @@ def authenticate(self, request):
...
+
+    def authenticate(self, request):
+        """
+        Authenticate the request and return a two-tuple of (user, token).
+        """
+        required_token = settings.SUMMARY_SERVICE_WEBHOOK_API_TOKEN
+        if not required_token:
+            raise AuthenticationFailed("Webhook authentication is not configured.")
+
+        auth_header: str = request.headers.get("Authorization") or ""
```
lebaudantoine: wip: probably, use the class attribute, `AUTH_HEADER` and `TOKEN_TYPE`?

src/backend/core/recording/event/authentication.py
```diff
@@ -91,3 +91,46 @@ def authenticate(self, request):
...
+        if not auth_header.startswith("Bearer "):
+            logger.warning(
+                "Authentication failed: Invalid authorization header format (ip: %s)",
+                request.META.get("REMOTE_ADDR"),
+            )
+            raise AuthenticationFailed("Invalid authorization header format.")
+        token = auth_header[7:]  # len("Bearer ") == 7
```
lebaudantoine: it doesn't seem robust to me, 
Could smth like this be appropriate : 

```
scheme, _, token = auth_header.partition(" ")
if scheme.lower() != "bearer" or not token.strip():
    raise AuthenticationFailed("Invalid authorization header format.")
token = token.strip()
```

src/backend/core/recording/event/authentication.py
```diff
@@ -91,3 +91,46 @@ def authenticate(self, request):
     def authenticate_header(self, request):
         """Return the WWW-Authenticate header value."""
         return f"{self.TOKEN_TYPE} realm='Storage event API'"
+
+
+class RecordingProcessWebhookAuthentication(BaseAuthentication):
```
lebaudantoine: Probably we should follow the implementation from https://github.com/suitenumerique/docs/blob/3fe8bd5cc0aa250e2513297cdfda34f219b180af/src/backend/core/authentication/__init__.py#L9, and maybe backport it to the django-lasuite library

src/helm/env.d/common.yaml.gotmpl
```diff
@@ -28,7 +28,25 @@ _summaryEnvVars: &summaryEnvVars
...
+    [
+      {
+        "id": "dictaphone",
+        "api_key": "dictaphone_token",
+        "webhook_url": "http://dictaphone-backend.dictaphone.svc.cluster.local/api/v1.0/ai-jobs/webhook/",
+        "webhook_api_key": "token_summary",
+        "allowed_push_to_docs": false
+      },
+      {
+        "id": "visio",
```
lebaudantoine:
```suggestion
        "id": "meet",
```

src/summary/summary/core/models.py
lebaudantoine: yes

src/summary/summary/core/models.py
```diff
@@ -46,6 +108,9 @@ class TranscribeTaskV2Payload(TranscribeTaskV2Request):
     """Model for creating a transcribe and summarize task (used for actual task creation)."""  # noqa: E501
```
lebaudantoine:
```suggestion
    """Model for creating a transcribe task (used for actual task creation)."""  # noqa: E501
```

src/summary/summary/core/models.py
```diff
@@ -12,6 +14,56 @@ class SharedV2TaskCreation(BaseModel):
...
+        "for the transcription task and push it to docs.",
+        default=False,
+    )
+
+
+class PushToDocsSummaryConfig(PushToDocsBaseConfig):
+    """Model for push to docs information for summaries."""
 
 
 class TranscribeTaskV2Request(SharedV2TaskCreation):
```
lebaudantoine: `TranscribeTaskV2Request` and `TranscribeTaskV2Payload` don't make clear that one is the API contract while the other is the data contract passed to the Celery worker. Renaming to `TranscribeTaskApiRequest` and `TranscribeTaskJob` would surface that distinction.

CHANGELOG.md
```diff
@@ -13,10 +13,12 @@ and this project adheres to
 - ✨(backend) add command to clean pending and deleted files
 - 🧱(helm) run clean files command as cronjob
 - ✨(backend) add fallback to save recordings without S3/MinIO webhooks
+- ✨(visio) use compatible with summary v2 #1362
```
lebaudantoine: Avoid mentioning Visio; the project's name is LaSuite Meet

src/summary/summary/api/route/tasks_v2.py
```diff
@@ -43,9 +82,27 @@ async def create_summarize_task_v2(
...
+    analytics.capture(
+        settings.posthog_event_request,
+        request.user_sub,
+        properties={
+            "kind": "summarize",
+            "$set": {
+                "email": request.user_email,
+            },
+        },
+    )
```
lebaudantoine, line 105: I wonder whether this query might create a new user with a different unique ID but the same email address. That would make the email-based feature flag work, but could also result in duplicate users in posthog

src/summary/summary/api/route/tasks_v2.py
```diff
@@ -20,17 +24,52 @@
...
+    analytics.capture(
+        settings.posthog_event_request,
+        request.user_sub,
+        properties={
+            "kind": "transcribe",
+            "$set": {
+                "email": request.user_email,
+            },
+        },
     )
```
lebaudantoine, line 73: same

src/backend/core/api/viewsets.py
```diff
@@ -989,6 +992,47 @@ def on_storage_event_received(self, request, pk=None):  # pylint: disable=unused
...
+        try:
+            recording = models.Recording.objects.get(external_process_id=job_id)
+        except models.Recording.DoesNotExist as e:
+            logger.warning("No recording found for job_id %s: %s", job_id, e)
+            recording = None
```
lebaudantoine: I guess you should return early if recording is not found

src/backend/core/api/viewsets.py
```diff
@@ -989,6 +992,47 @@ def on_storage_event_received(self, request, pk=None):  # pylint: disable=unused
...
+
+        changed = False
+        if recording and data.get("type") == "transcript":
```
lebaudantoine: again, we would benefit from having data validation for the type and status. 
We should rely on drf serializer, which makes explicit what the endpoint accepts as a payload, and clarifies the API contract.

src/backend/core/api/viewsets.py
```diff
@@ -989,6 +992,47 @@ def on_storage_event_received(self, request, pk=None):  # pylint: disable=unused
...
+                )
+                changed = True
+            if data.get("status") == "failure":
+                recording.status = models.RecordingStatusChoices.EXTERNAL_PROCESS_FAILED
+                changed = True
+
+        if changed and recording:
+            recording.save()
+        else:
+            logger.info("No changes to save for external process id %s", job_id)
```
lebaudantoine: Actually, the changed variable is misleading here. If the recording was already set to that value and the webhook gets triggered again for any reason (even incorrectly), the code would still interpret changed as a state transition, while in reality, nothing changed.

I would probably remove the log info and changed variable, and consider every call to the viewset update the recording row

src/backend/core/api/viewsets.py
lebaudantoine: what are the other possible types?

src/summary/summary/core/webhook_service.py
lebaudantoine, line 48: Do we need a retry mechanism here?

It would be a shame for the transcription to fail entirely just because the final HTTP request failed once. Seems worth making this more resilient.

src/summary/summary/core/celery_worker.py
```diff
@@ -617,6 +483,53 @@ def process_audio_transcribe_v2_task(
...
+            )
+            locale = get_locale(payload.context_language, payload.language)
+            create_document_in_docs(
+                content=summary,
+                title=locale.summary_title_template.format(
+                    title=payload.push_to_docs_config.title
+                ),
+                email=payload.push_to_docs_config.user_email,
+                sub=payload.user_sub,
+            )
```
lebaudantoine, line 536: That’s one of the major changes here: waiting for the summary output synchronously. I don’t think this is the right pattern.

Right now, because of an issue in the Albert API, the summarization job is failing, and I haven’t fixed it yet. With this setup, that means all transcription jobs will fail from a Celery standpoint. To me, that clearly shows that coupling these two tasks is not the right approach.

As soon as we start working again on the summary feature, we need to be able to iterate on it freely without any risk of impacting the transcription pipeline, which is production-ready.

We’re also regressing operationally: with the current design, we lose the ability to scale workers independently. Today, we can dedicate workers to summaries (which involve external API calls and can block), while keeping transcription workers isolated and efficient.

For me, this is a blocking concern for merge.

src/summary/summary/core/celery_worker.py
lebaudantoine, line 463:
```suggestion
    task_id = self.request.id
```

src/summary/summary/core/celery_worker.py
```diff
@@ -526,32 +365,59 @@ def summarize_transcription_internals(
...
+        return False
+    if not settings.is_docs_integration_enabled:
+        logger.info("Docs integration is disabled, skipping push to docs.")
+        return False
+    if not settings.get_authorized_tenant(
+        tenant_id=payload.tenant_id
+    ).allowed_push_to_docs:
+        logger.info("Push to docs is not allowed for tenant, skipping push to docs.")
+        return False
+    return True
```
lebaudantoine: is the info the right level? 

I’d go with this approach, since all those if statements really only exist to compute the reason that gets logged. Sharing a common prefix would also make the logs easier to parse when this situation occurs.

```suggestion
def _should_push_to_docs() -> bool:
    """Wip."""
    if not payload.push_to_docs_config:
        reason = "not requested in the payload"
    elif not settings.is_docs_integration_enabled:
        reason = "docs integration is disabled"
    elif not settings.get_authorized_tenant(
        tenant_id=payload.tenant_id
    ).allowed_push_to_docs:
        reason = "not allowed for tenant"
    else:
        return True

    logger.info("Skipping push to docs: %s.", reason)
    return False
```

lebaudantoine:
@greptileai

src/backend/core/api/viewsets.py
FloChehab: Letting you close or bounce back @lebaudantoine

src/backend/core/models.py
FloChehab: Letting you close or bounce back @lebaudantoine

src/summary/summary/core/config.py
FloChehab: Letting you close or bounce back @lebaudantoine

src/backend/meet/settings.py
FloChehab: Letting your close or bounce back @lebaudantoine

src/backend/core/recording/event/notification.py
FloChehab: Letting you close or bounce back @lebaudantoine

src/summary/summary/api/route/tasks_v2.py
FloChehab: Dicussed IRL we are going to create a new posthog project for this

src/backend/core/api/viewsets.py
FloChehab: Summary in this case, hence why I didn't do a very strict serializer.

src/summary/summary/core/celery_worker.py
FloChehab: Fixed in 2b60ccb3

src/summary/summary/core/webhook_service.py
FloChehab: The retry is handled at the task level, which is "better" as it releases the worker for another task while waiting.

FloChehab: I'll you close this one @lebaudantoine

src/summary/summary/core/celery_worker.py
FloChehab: There is a bit of mix of job and task I am sorry, I suggest leaving it that way in that PR. The external API is "job_id".

src/backend/core/api/viewsets.py
FloChehab: ok

src/backend/core/api/viewsets.py
FloChehab: Fixed in d8d069e1

src/backend/core/api/viewsets.py
FloChehab: Fixed in d8d069e1

src/backend/core/api/viewsets.py
FloChehab: Fixed in d8d069e1

src/backend/core/api/viewsets.py
FloChehab: fixed in d8d069e1

src/summary/summary/api/route/tasks_v2.py
FloChehab: I'll let you close this one @lebaudantoine

src/summary/summary/api/route/tasks_v2.py
```diff
@@ -20,32 +24,85 @@
...
 
     return TranscribeWebhookPendingPayload(job_id=task.id).model_dump()
```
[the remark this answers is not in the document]
FloChehab: false

src/backend/core/recording/event/authentication.py
FloChehab: Our version was a bit more robust, I refactored it in cdd72912 happy to open a PR in DjangoLasuite later.

src/backend/core/recording/event/authentication.py
FloChehab: Refactored in cdd72912

src/backend/core/models.py
lebaudantoine: ok

src/backend/core/recording/event/authentication.py
lebaudantoine: let's go

lebaudantoine (approved):

PR #1377, opened 2026-06-01
🔒️(backend) prevent file change post checks

FloChehab:
Before this commit, after a file check, the policy could be reused to change the verified file.
Now, files are uploaded to a temporary location, then inside a transaction that prevents concurrent calls, the file is copied to its final destination and the checks are run on that one. A new file can still be updated with the policy but it will never be read, etc.

As part of this change, all files in the new tmp directory on s3 should have an expiration policy (to be configured at the infrastructure level)

Follow up to https://github.com/suitenumerique/meet/pull/1030#discussion_r2878416690


FloChehab:
@coderabbitai full review

src/backend/core/api/viewsets.py
```diff
@@ -1196,96 +1196,116 @@ def upload_ended(self, request, *args, **kwargs):
...
+                range_response = s3_client.get_object(
+                    Bucket=default_storage.bucket_name,
+                    Key=file.file_key,
+                    Range="bytes=0-2047",
+                )
+                file_head = range_response["Body"].read()
+            else:
+                file_head = s3_client.get_object(
+                    Bucket=default_storage.bucket_name, Key=file.file_key
+                )["Body"].read()
```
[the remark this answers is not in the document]
FloChehab: Hum, I think It's ok to 500 here

src/backend/core/api/viewsets.py
```diff
@@ -1196,96 +1196,116 @@ def upload_ended(self, request, *args, **kwargs):
...
+        self.get_object()
 
         s3_client = default_storage.connection.meta.client
+        validation_error = None
 
-        head_response = s3_client.head_object(
-            Bucket=default_storage.bucket_name, Key=file.file_key
-        )
-        file_size = head_response["ContentLength"]
+        with transaction.atomic():
```
kernicPanel: Having all those s3 calls in an atomic transaction would not cause problems ?
I'm thinking about multiple files upload, and concurrent upload by multiple users.

FloChehab: It's just the upload-ended that is run inside a transaction.
This transaction should "lock" only the specific file so I think it's ok perf wise.

FloChehab: The intent is to block concurrent call the the upload-ended on the same file.

kernicPanel: Yes, I got the intent, but I feared some side effects.

lunika: I agree with Nico. The `select_for_update` is creating a lock in the table for the row and the transaction will be very long due to all http requests with the object storage. We had a lot of trouble on Docs with locks... It can **maybe** scale in your case. I'm pretty sure it will not on Drive.

lunika: An other solution to create a lock is to use the django cache with a short expire.

FloChehab: Tweaked in [7f4ad1e](https://github.com/suitenumerique/meet/pull/1377/commits/7f4ad1e95bb3f3a60db7b3bc2de46121dcb88962)

src/backend/meet/settings.py
```diff
@@ -1049,6 +1052,11 @@ def post_setup(cls):
         """
         super().post_setup()
 
+        if cls.FILE_UPLOAD_TMP_PATH == cls.FILE_UPLOAD_PATH:
+            raise ValueError(
+                "FILE_UPLOAD_TMP_PATH cannot be the same as FILE_UPLOAD_PATH"
+            )
```
kernicPanel, line 1096: Instead of raising if env vars clashes, WDYT of using a prefix ?

```python
    FILE_UPLOAD_TMP_PREFIX = values.Value(
        "tmp", environ_name="FILE_UPLOAD_TMP_PREFIX", environ_prefix=None
    )
```    

```suggestion
    cls.FILE_UPLOAD_TMP_PATH = path.join(cls.FILE_UPLOAD_TMP_PREFIX, cls.FILE_UPLOAD_PATH)
```

kernicPanel (commented):
Just a small comment. Everything else looks good to me.

src/backend/meet/settings.py
FloChehab: I think it's better to be explicit, especially because s3 file conservation policies will have to be put in place.
Also, we would need to keep the check :
```
In [5]: path.join("/tmp/toto", "")
Out[5]: '/tmp/toto/'
```

lunika (approved):

src/backend/meet/settings.py
kernicPanel: Indeed :+1:

kernicPanel (approved):

PR #1383, opened 2026-06-02
✨(frontend) cap and paginate tiles in picture-in-picture

Ovgodd:
## Purpose

In big meetings, PiP was showing all camera tiles.
It was using too much render + WebRTC subscribe for nothing.

## Proposal

We limit PiP to 5 tiles per page.
If more people, user can paginate.
Active speaker stays priority and page stays stable.
Not visible tiles are unmounted, so subscriptions can drop.

https://github.com/user-attachments/assets/9a721b96-08c1-4afc-86d8-c0be46609f10

- [x] Add cap 5 with `usePagination` in `PipStage`
- [x] Add `PipPagination` (simple + accessible)
- [x] Keep screen share / active speaker in first page
- [x] Improve `PaginationControl` accessibility (nav + live count)
- [x] Add `pagination.label` in i18n (en, fr, de, nl)

### DIFF
--- a/src/frontend/src/features/pip/components/layout/PipPagination.tsx
+++ b/src/frontend/src/features/pip/components/layout/PipPagination.tsx
@@ -1,0 +1,96 @@
+import { RiArrowLeftSLine, RiArrowRightSLine } from '@remixicon/react'
+import { useTranslation } from 'react-i18next'
+import { styled } from '@/styled-system/jsx'
+
+interface PipPaginationProps {
+  totalPageCount: number
+  currentPage: number
+  nextPage: () => void
+  prevPage: () => void
+}
+
+export const PipPagination = ({
+  totalPageCount,
+  currentPage,
+  nextPage,
+  prevPage,
+}: PipPaginationProps) => {
+  const { t } = useTranslation('rooms', { keyPrefix: 'pagination' })
+
+  if (totalPageCount <= 1) return null
+
+  return (
+    <Nav aria-label={t('label')}>
+      <ArrowButton
+        type="button"
+        onClick={prevPage}
+        disabled={currentPage === 1}
+        aria-label={t('previous')}
+      >
+        <RiArrowLeftSLine size={18} />
+      </ArrowButton>
+      <Counter role="status">
+        {t('count', { currentPage, totalPageCount })}
+      </Counter>
+      <ArrowButton
+        type="button"
+        onClick={nextPage}
+        disabled={currentPage === totalPageCount}
+        aria-label={t('next')}
+      >
+        <RiArrowRightSLine size={18} />
+      </ArrowButton>
+    </Nav>
+  )
+}
+
+const Nav = styled('nav', {
+  base: {
+    display: 'flex',
+    alignItems: 'center',
+    justifyContent: 'center',
+    gap: '0.25rem',
+    marginTop: '1rem',
+    flexShrink: 0,
+  },
+})
+
+const ArrowButton = styled('button', {
+  base: {
+    display: 'inline-flex',
+    alignItems: 'center',
+    justifyContent: 'center',
+    width: '1.75rem',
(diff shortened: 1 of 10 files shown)


Ovgodd:
@coderabbitai full review

src/frontend/src/features/pip/components/layout/PipPagination.tsx
```diff
@@ -0,0 +1,96 @@
+import { RiArrowLeftSLine, RiArrowRightSLine } from '@remixicon/react'
```
lebaudantoine, line 1: After quickly reviewing the code, I have the impression that we may not properly handle subscription and unsubscription of hidden tracks. Is that correct?

Would it make sense to reuse `PaginationControl` and `PaginationIndicator` to manage the subscription logic, instead of duplicating this fairly complex logic across multiple places?

To clarify: the goal is not only to hide video elements in the DOM, but also to ensure that participants subscribe only to the tracks that are currently visible. This is important to avoid unnecessary media consumption and potential overload of the media server.

It might be useful to look into concepts such as adaptive streaming or dynacast, which highlight the optimizations available on the client side. These also explain why layout components need to be very careful when dealing with both video rendering and subscription management.

Ovgodd: Hey, after investigations, I looked into this carefully, both in the code and at runtime.

Off-page tracks are unmounted, not just hidden, and `adaptiveStream` disables them on the SFU (Selective Forwarding Unit media server of livekit). No subscription logic was duplicated.

PiP uses the same `usePagination` hook as `GridLayout` in the main window.
Only `pagination.tracks`, meaning the current page, is passed to `PipGridLayout`.

Off-page tracks have:
- no `<ParticipantTile>`;
- no `<VideoTrack>`;
- no `<video>` element in the DOM.

When the page changes, React unmounts tiles whose keys leave the `.map()`.

This triggers the following cleanup chain:
```txt
- React unmounts
-  <ParticipantTile> unmounts
-  <VideoTrack> unmounts
-  useMediaTrackBySourceOrName cleanup
-  track.detach(element)
-  RemoteVideoTrack.stopObservingElement()
-  updateVisibility()
-  visible = false
-  adaptiveStream sends disabled: true to the SFU
-  SFU stops forwarding video frames
```

This works because of two settings already configured in `Conference.tsx`:
- `adaptiveStream: true` : the client tells the SFU to stop sending video when no `<video>` element is attached.
- `dynacast: true` : the server stops encoding layers that no one consumes.

Why not reuse `PaginationControl` / `PaginationIndicator`?

Both are **UI-only** components: arrows, page counter, dots.

They contain no subscription logic.
The subscription logic lives in `usePagination`, the LiveKit hook shared between the main window and PiP.

`PipPagination` is a separate small component because:
- `PaginationControl` uses `position: absolute` and auto-hide on mouse movement, which does not work well in the PiP flex layout.
- PiP needs pagination to stay always visible.
- PiP needs dark-themed styling.

### Runtime tests

Test setup:

- 13 participants
- pages: `5 + 5 + 3`

### DOM check

Checked by inspecting the PiP window.

| Page | `<video>` elements |
| ---- | ------------------ |
| 1    | 5                  |
| 2    | 5                  |
| 3    | 3                  |

### Console logs

Filter used:

```txt
adaptivestream video visibility
```

On page change from page 2 to page 3:

- 5 tracks become `visible=false` when leaving the page.
- 3 tracks become `visible=true` when arriving on the page.
- Disabled tracks then show `enabled: false` and `dimensions 0x0`.
- A `send update settings` message is sent to the SFU.

The same pattern happens on every page change.

Let me know if you have questions mate !

lebaudantoine:
@greptileai

lebaudantoine (approved):
tested locally, lgtm, video tracks are indeed set to visible=false

PR #1386, opened 2026-06-02
✨(backend) add LiveKit egress_ended fallback for saving recordings

cameledev:
## Purpose & Proposal

https://github.com/orgs/suitenumerique/projects/11?pane=issue&itemId=187635669

### Variations on original spec:

- No new setting: Instead, allways fallback to newly implement egress_ended check when `RECORDING_STORAGE_EVENT_ENABLE` is not defined or false


src/backend/core/services/livekit_events.py
```diff
@@ -135,14 +135,14 @@ def receive(self, request):
...
-        # pylint: disable=not-callable
-        handler(data)
+        if webhook_type.value == "egress_updated":
+            self._handle_egress_updated(data)
+        if webhook_type.value == "egress_ended":
+            self._handle_egress_ended(data)
+        if webhook_type.value == "room_started":
+            self._handle_room_started(data)
+        if webhook_type.value == "room_finished":
+            self._handle_room_finished(data)
```
cameledev: changed for clarity

src/backend/core/tests/recording/test_api_recordings_storage_hook.py
```diff
@@ -224,3 +235,43 @@ def test_save_recording_success(recording_settings, mock_get_parser, client, sta
...
+
+
+@pytest.mark.parametrize(
+    "notification_succeeded,expected_status",
+    [
+        (True, RecordingStatusChoices.NOTIFICATION_SUCCEEDED),
+        (False, RecordingStatusChoices.SAVED),
+    ],
+)
+def test_save_recording_notifies_external_services(
```
cameledev: AFAICT call to `notify_external_services` was not being tested previously. Added here.

src/backend/core/api/viewsets.py
```diff
@@ -972,24 +972,15 @@ def on_storage_event_received(self, request, pk=None):  # pylint: disable=unused
         except models.Recording.DoesNotExist as e:
             raise drf_exceptions.NotFound("No recording found for this event.") from e
 
-        if not recording.is_savable():
```
cameledev, line 975: done in `handle_complete`

src/backend/core/recording/event/parsers.py
```diff
@@ -58,7 +58,7 @@ def __init__(self, bucket_name, allowed_filetypes=None):
     def parse(self, data: Dict) -> StorageEvent:
         """Extract storage event data from raw dictionary input."""
 
-    def validate(self, data: StorageEvent) -> None:
+    def validate(self, data: StorageEvent) -> str:
```
cameledev, line 61: fixing wrong signature

src/backend/core/services/livekit_events.py
lebaudantoine: I'm not convinced by this change. While it aligns with the explicit over implicit principle, it conflicts with the open/closed principle, every new webhook type will require modifying this method. With at least 10 event types, duplicating that logic 10 times doesn't strike me as cleaner or more readable.

I prefer the dynamic dispatch. Moreover, this change is not really related with the main goal of the PR.

src/backend/core/services/livekit_events.py
```diff
@@ -195,6 +196,21 @@ def _handle_egress_ended(self, data):
...
+            try:
+                self.recording_events.handle_complete(recording)
+            except RecordingNotSavableError:
+                logger.info(
+                    "Recording %s is not savable on egress complete "
+                    "(already saved or in an error state); ignoring.",
+                    recording.id,
+                )
+
+        # Silently ignoring EGRESS_ABORTED, EGRESS_FAILED
```
lebaudantoine, line 219: Should we address them in a dedicated PR?

src/backend/core/tests/services/test_livekit_events.py
```diff
@@ -326,6 +326,114 @@ def test_handle_egress_ended_does_not_call_metadata_collector_stop_when_conditio
...
+    mock_data = mock.MagicMock()
+    mock_data.egress_info.egress_id = recording.worker_id
+    mock_data.egress_info.status = EgressStatus.EGRESS_COMPLETE
+
+    service._handle_egress_ended(mock_data)
+
+    mock_notify_external_services.assert_called_once_with(recording)
+
+    recording.refresh_from_db()
+    assert recording.status == "notification_succeeded"
```
lebaudantoine: should we merge these tests?

src/backend/core/services/livekit_events.py
```diff
@@ -195,6 +196,21 @@ def _handle_egress_ended(self, data):
...
                 ) from e
 
+        # Fallback for completion when no MinIO/S3 webhooks are configured
+        if (
+            not settings.RECORDING_STORAGE_EVENT_ENABLE
+        ) and data.egress_info.status == api.EgressStatus.EGRESS_COMPLETE:
+            try:
+                self.recording_events.handle_complete(recording)
+            except RecordingNotSavableError:
+                logger.info(
```
lebaudantoine: I'm not sure, should it be a warning?

src/backend/core/services/livekit_events.py
```diff
@@ -195,6 +196,21 @@ def _handle_egress_ended(self, data):
                     f"Failed to process limit reached event for recording {recording}"
                 ) from e
 
+        # Fallback for completion when no MinIO/S3 webhooks are configured
+        if (
+            not settings.RECORDING_STORAGE_EVENT_ENABLE
```
lebaudantoine, line 205: let's discuss this settings name

cameledev: yes !

src/backend/core/tests/services/test_livekit_events.py
```diff
@@ -326,6 +326,114 @@ def test_handle_egress_ended_does_not_call_metadata_collector_stop_when_conditio
...
+
+
+@mock.patch("core.utils.update_room_metadata")
+def test_handle_egress_ended_does_not_save_when_feature_disabled(
```
cameledev: rename

src/backend/core/tests/services/test_livekit_events.py
cameledev: done, with parametrize.

src/backend/core/tests/services/test_livekit_events.py
cameledev: done

src/backend/core/services/livekit_events.py
cameledev: to be done in future PR

src/backend/core/services/livekit_events.py
cameledev: Changed for dict registry following discussions

src/backend/core/services/livekit_events.py
cameledev: changed for warning

src/backend/core/services/livekit_events.py
cameledev: https://github.com/suitenumerique/meet/issues/1391

lebaudantoine (commented):
Tested locally, it works great

src/backend/core/services/livekit_events.py
```diff
@@ -135,14 +136,16 @@ def receive(self, request):
...
 
-        handler_name = f"_handle_{webhook_type.value}"
-        handler = getattr(self, handler_name, None)
+        # Handle according to received webhook type
+        handler = {
+            "egress_updated": self._handle_egress_updated,
+            "egress_ended": self._handle_egress_ended,
+            "room_started": self._handle_room_started,
+            "room_finished": self._handle_room_finished,
+        }.get(webhook_type.value)
```
lebaudantoine: way better!

lebaudantoine: I moved it to the constructor to avoid redefining the dict every call.

src/backend/core/api/viewsets.py
```diff
@@ -972,24 +972,15 @@ def on_storage_event_received(self, request, pk=None):  # pylint: disable=unused
...
+        # Save recording
+        livekit_events_service = LiveKitEventsService()
+        try:
+            livekit_events_service.recording_events.handle_complete(recording)
```
lebaudantoine: Why not directly instantiate the RecordingEventsService? It would be more straightforward and easier to follow.

docs/features/recording.md
```diff
@@ -96,7 +96,7 @@ sequenceDiagram
...
-| **RECORDING_STORAGE_EVENT_ENABLE**      | Boolean     | `False`                                                                                                                                                            | Enable handling of storage events (must configure webhook in storage).                                                                                                                                                                                                                             |
+| **RECORDING_STORAGE_EVENT_ENABLE**      | Boolean     | `False`                                                                                                                                                            | Enable handling of storage events (must configure webhook in storage). If `False`, fallback to LiveKit egress complete webhook.                                                    |
```
lebaudantoine, line 99: your commit mentions you added "a configurable setting " but I guess, we did not really add any new settings.

src/backend/core/services/livekit_events.py
```diff
@@ -195,6 +198,21 @@ def _handle_egress_ended(self, data):
                     f"Failed to process limit reached event for recording {recording}"
                 ) from e
 
+        # Fallback for completion when no MinIO/S3 webhooks are configured
+        if (
+            not settings.RECORDING_STORAGE_EVENT_ENABLE
```
lebaudantoine, line 205: I don't remember your initial version, but I'm probably suggesting what you initially wrote. There is my suggestion : 

Let's introduce a new setting

```
# Save recordings on LiveKit 'egress_ended' (EGRESS_COMPLETE) webhook.
# Alternative to storage events for deployments without MinIO/S3 bucket
# notifications. Ignored when RECORDING_STORAGE_EVENT_ENABLE is True:
# the storage event hook remains the source of truth for saving.
RECORDING_SAVE_ON_EGRESS_ENDED = values.BooleanValue(
    False, environ_name="RECORDING_SAVE_ON_EGRESS_ENDED", environ_prefix=None
)
```

and then 

```
# Fallback for completion when MinIO/S3 storage events are not used.
# RECORDING_STORAGE_EVENT_ENABLE takes precedence: when storage events
# are enabled, they remain the only mechanism that marks recordings saved.
save_on_egress_ended = (
    settings.RECORDING_SAVE_ON_EGRESS_ENDED
    and not settings.RECORDING_STORAGE_EVENT_ENABLE
)

if (
    save_on_egress_ended
    and data.egress_info.status == api.EgressStatus.EGRESS_COMPLETE
):

```

and optionally, add a warning in [the settings post_setup](https://github.com/suitenumerique/meet/blob/90f95ab2a954ae4e2b4361a0fe2c1e90ec16d580/src/backend/meet/settings.py#L1086) as, 

```
if cls.RECORDING_STORAGE_EVENT_ENABLE and cls.RECORDING_SAVE_ON_EGRESS_ENDED:
    getLogger(__name__).warning(
        "RECORDING_SAVE_ON_EGRESS_ENDED is ignored because "
        "RECORDING_STORAGE_EVENT_ENABLE is True; recordings are saved "
        "via the storage event hook."
    )
```

src/backend/core/services/livekit_events.py
```diff
@@ -195,6 +200,21 @@ def _handle_egress_ended(self, data):
...
+        ) and data.egress_info.status == api.EgressStatus.EGRESS_COMPLETE:
+            try:
+                self.recording_events.handle_complete(recording)
+            except RecordingNotSavableError:
+                logger.warning(
+                    "Recording %s is not savable on egress complete "
+                    "(already saved or in an error state); ignoring.",
+                    recording.id,
+                )
```
lebaudantoine: I tried lowering the egress limit to 1 minute, and I think we have an issue. When the egress stops because the limit is reached, the `egress complete` event is not triggered. This seems to confirm my initial suspicion.

docs/features/recording.md
cameledev: My bad. Comment was that of original implem, then updated. Will update the comment.

src/backend/core/services/livekit_events.py
cameledev: This was indeed the original spec and implem. 

I believe we had discussed removing this extra variable, as the problem is actually binary: either we want the extra security of waiting for bucket confirmation, or we don't. In this case, we just need one bool variable.

I don't see the situation where we would not want to save the file at all.

What do you think ?

src/backend/core/services/livekit_events.py
cameledev: TODO: this should be fixed by adding a check on status `LIMIT_REACHED`

src/backend/core/tests/services/test_livekit_events.py
```diff
@@ -367,82 +376,38 @@ def test_handle_egress_ended_notifies_external_services_on_complete(  # noqa: PL
...
-    assert recording.status == recording_status
-
-
-@mock.patch(
-    "core.recording.services.recording_events.notification_service."
-    "notify_external_services"
-)
-@mock.patch("core.utils.notify_participants")
-@mock.patch("core.utils.update_room_metadata")
-def test_handle_egress_ended_does_not_finalize_on_limit_reached_when_webhooks_enabled(
```
cameledev: now in `test_handle_egress_ended_finalizes_recording`

src/backend/core/tests/services/test_livekit_events.py
```diff
@@ -367,82 +376,38 @@ def test_handle_egress_ended_notifies_external_services_on_complete(  # noqa: PL
...
     service._handle_egress_ended(mock_data)
 
     mock_notify_external_services.assert_not_called()
 
     recording.refresh_from_db()
-    assert recording.status == "stopped"
-
-
-@mock.patch("core.utils.update_room_metadata")
-def test_handle_egress_ended_does_not_save_when_webhooks_enabled(
```
cameledev: replaced by `test_handle_egress_ended_does_not_finalize_when_webhooks_enabled`

src/backend/core/services/livekit_events.py
cameledev: Fixed. added tests

src/backend/core/api/viewsets.py
cameledev: changed

lebaudantoine (approved):

PR #1395, opened 2026-06-03
🔒️(backend) prevent accessing files if they are not ready

FloChehab:
With the addition of the ANALYSING state
files could be accessed in the short time they
were in that state.
We now require files to be in ready.

Also adds missing frontend types (no impact).


lebaudantoine (approved):
LGTM, did not try locally, but it looks like a minor update.

PR #1396, opened 2026-06-03
⬆️(dependencies) update python dependencies

cameledev:
Update python dependencies.

Tested locally.

### DIFF
--- a/src/backend/pyproject.toml
+++ b/src/backend/pyproject.toml
@@ -2,7 +2,7 @@
 # Meet package
 #
 [build-system]
-requires = ["uv_build>=0.10.9,<0.11.0"]
+requires = ["uv_build>=0.11.16,<0.12.0"]
 build-backend = "uv_build"
 
 [project]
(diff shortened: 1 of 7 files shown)


FloChehab (commented):
The only major bump is for gunicorn and we should be good.
Maybe we can prefix this branch with `integration/...` so that we can deploy it on staging early and leave it there without impacting master.

i think that would require a thorough QA, so pinging @lebaudantoine

src/backend/pyproject.toml
```diff
@@ -24,45 +24,45 @@ keywords = ["Django", "Contacts", "Templates", "RBAC"]
...
     "django-pydantic-field==0.5.4",
     "django==5.2.14",
-    "djangorestframework==3.16.1",
+    "djangorestframework==3.17.1",
     "drf_spectacular==0.29.0",
     "dockerflow==2026.3.4",
     "easy_thumbnails==2.10.1",
     "factory_boy==3.3.3",
-    "gunicorn==25.1.0",
+    "gunicorn==26.0.0",
```
lebaudantoine, line 49: It was released a month ago. I’ll merge your PR, but for such a critical part of the application, we should probably be a bit more conservative when it comes to upgrades.

PR #1399, opened 2026-06-03
✨(backend) Purge deleted & pending files

FloChehab:
Commits are self explanatory :
1. Adds 2 commands for deleting pending files & purging soft deleted files,
2. Adds the ability to run cronjobs with the helm chart and by default those 2 commands are run.

Initially based on code from Docs, that was ported to Transcripts and then here :)

### DIFF
--- a/src/backend/core/api/viewsets.py
+++ b/src/backend/core/api/viewsets.py
@@ -1187,7 +1187,10 @@ class FileViewSet(
         serializer.save(creator=self.request.user)
 
     def perform_destroy(self, instance):
-        """Override to implement a soft delete instead of dumping the record in database."""
+        """Override to implement a soft delete instead of dumping the record in database.
+
+        Files are actually purged by commands that should run periodically.
+        """
         instance.soft_delete()
 
     @decorators.action(detail=True, methods=["post"], url_path="upload-ended")
--- a/src/backend/core/management/commands/clean_pending_files.py
+++ b/src/backend/core/management/commands/clean_pending_files.py
@@ -1,0 +1,47 @@
+"""Clean stale pending files that were never fully uploaded."""
+
+from datetime import timedelta
+
+from django.core.management.base import BaseCommand, CommandError
+from django.utils import timezone
+
+from core.models import File, FileUploadStateChoices
+from core.tasks.file import process_file_deletion
+
+
+class Command(BaseCommand):
+    """Remove pending files older than a given threshold."""
+
+    help = "Delete pending files that have been stuck for too long"
+
+    def add_arguments(self, parser):
+        parser.add_argument(
+            "--hours",
+            type=int,
+            default=24,
+            help="Age threshold in hours (default: 24)",
+        )
+
+    def handle(self, *args, **options):
+        hours = options["hours"]
+        if hours < 0:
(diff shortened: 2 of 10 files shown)


lebaudantoine:
@coderabbitai full review

src/backend/core/api/viewsets.py
```diff
@@ -1187,7 +1187,10 @@ def perform_create(self, serializer):
         serializer.save(creator=self.request.user)
 
     def perform_destroy(self, instance):
-        """Override to implement a soft delete instead of dumping the record in database."""
+        """Override to implement a soft delete instead of dumping the record in database.
+
+        Files are actually delete by commands that should run periodically.
+        """
```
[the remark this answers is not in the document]
lebaudantoine: agree

src/backend/core/management/commands/clean_pending_files.py
```diff
@@ -0,0 +1,41 @@
...
+
+    help = "Delete pending files that have been stuck for too long"
+
+    def add_arguments(self, parser):
+        parser.add_argument(
+            "--hours",
+            type=int,
+            default=24,
+            help="Age threshold in hours (default: 24)",
+        )
```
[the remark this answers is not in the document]
lebaudantoine: True, it’s usually worth enforcing that explicitly rather than relying on generic integer parsing.

src/backend/core/management/commands/clean_pending_files.py
```diff
@@ -0,0 +1,41 @@
...
+
+        count = 0
+        for file in files.iterator():
+            file.soft_delete()
+            file.hard_delete()
+            process_file_deletion(file.id)
+            count += 1
+
+        self.stdout.write(f"Cleaned {count} stale pending file(s).")
```
lebaudantoine, line 47: I’m not 100% sure this comment is relevant, so feel free to discard it if it’s not.

With this query, I’m wondering if some files could have a non-null `deleted_at` field. In that case, calling `soft_delete` on them might raise an exception and cause the command to fail.

Is this something we should account for? Are we comfortable assuming that filtering with `hard_deleted_at = null` implicitly means we’re only dealing with files where `deleted_at = null` as well?

src/backend/meet/settings.py
```diff
@@ -182,7 +182,9 @@ class Base(Configuration):
         environ_name="FILE_UPLOAD_ENABLED",
         environ_prefix=None,
     )
-
+    FILE_PURGE_GRACE_DAYS = values.IntegerValue(
```
lebaudantoine: It should be a PositiveIntegerValue

src/helm/meet/Chart.yaml
lebaudantoine, line 4: I've released a chart 0.0.23 since then. We should bump the version to 0.0.24

src/backend/core/management/commands/clean_pending_files.py
FloChehab: From an application standpoint this case shouldn't happen. But I am just going to guard against it, just in case.

lebaudantoine (approved):
Nice, you can merge it, and deploy you're new chart in staging, so we can test it

PR #1401, opened 2026-06-03
♿️(frontend) improve accessibilty of the Effects panel

FloChehab:
Since we have added the custom backgrounds we needed to update the accessibility on the effects panel.
I have fixed few aria-labels on the way (that I might have broken back when adding the feature), I also took the liberty to do further improvements to the structe.

### DIFF
--- a/src/frontend/src/features/rooms/livekit/components/effects/EffectsConfiguration.tsx
+++ b/src/frontend/src/features/rooms/livekit/components/effects/EffectsConfiguration.tsx
@@ -400,6 +400,8 @@ export const EffectsConfiguration = ({
       config: ProcessorConfig
       isSelected: boolean
       tooltip: string
+      ariaLabel: string
+      ariaDeleteLabel: string
       file: ApiFileItem
     }[]
   }>(() => {
@@ -445,9 +447,11 @@ export const EffectsConfiguration = ({
         }
         const id = deriveIdFromProcessorConfig(config)
         const isSelected = selectedId === id
-        const prefix = isSelected ? 'selectedLabel' : 'apply'
         const backgroundName = t(`virtual.presets.descriptions.${index}`)
-        const ariaLabel = `${t(`virtual.presets.${prefix}`)} ${backgroundName}`
+        const ariaLabelPrefix = t(
+          isSelected ? `virtual.selectedLabel` : `virtual.apply`
+        )
+        const ariaLabel = `${ariaLabelPrefix} ${backgroundName}`
 
         return {
           tooltip: backgroundName,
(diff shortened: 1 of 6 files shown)


src/frontend/src/features/rooms/livekit/components/effects/EffectsConfiguration.tsx
```diff
@@ -552,6 +566,7 @@ export const EffectsConfiguration = ({
               [layout === 'vertical' ? 'height' : 'minHeight']: '175px',
               borderRadius: '8px',
             }}
+            aria-hidden={true}
```
lebaudantoine, line 570: why?

lebaudantoine (commented):
LGTM. One question regarding the accessibility tree.

From the screen reader's perspective, the current structure appears to be:

* Apply background name
* Remove background name

Since a custom background exposes two actions associated with the same item, would it make more sense for the hierarchy to be:

* Background name

  * Apply
  * Remove

rather than exposing each action as a separate top-level item?

I'm wondering whether grouping the actions under the background itself would provide a clearer navigation experience for screen reader users.

Image

?  

Wdyt @Ovgodd ?

Except from this question, good to merge.

lebaudantoine:
@coderabbitai full review

src/frontend/src/features/rooms/livekit/components/effects/EffectsConfiguration.tsx
FloChehab: I feel people with a screen reader shouldn't care about the presence of the video feedback. But open to revert.

FloChehab:
> LGTM. One question regarding the accessibility tree.
> 
> From the screen reader's perspective, the current structure appears to be:
> 
>     * Apply background name
> 
>     * Remove background name
> 
> 
> Since a custom background exposes two actions associated with the same item, would it make more sense for the hierarchy to be:
> 
>     * Background name
>       
>       * Apply
>       * Remove
> 
> 
> rather than exposing each action as a separate top-level item?
> 
> I'm wondering whether grouping the actions under the background itself would provide a clearer navigation experience for screen reader users.
> Image
> 
> ?
> 
> Wdyt @Ovgodd ?
> 
> Except from this question, good to merge.

I think from a screen reader perspective the apply / remove doesn't work : remove should be present only when the the background image is selected.
Curious to hear the expert perspective on this :)

lebaudantoine:
An image can be removed anytime, selected or not, did i miss something? @FloChehab 

Capture d’écran 2026-06-04 à 20 37 25

FloChehab:
I have tweaked things a bit to get closer to what you suggested @lebaudantoine 

image

lebaudantoine (approved):

FloChehab:
Merged, @Ovgodd feel free to come back to this :) I have merged to move and I think it's way better than what we currently have haha

PR #1409, opened 2026-06-04
feat(auth): OAuth2 + PKCE token endpoint for native apps

mmaudet:
## Summary

Adds an RFC 8252 / RFC 7636 PKCE flow to Meet so native mobile and desktop apps
can authenticate against a Meet instance without sharing a session cookie.

**Why:** native clients (Android, iOS, Tauri) can't keep a Django session
cookie alive cleanly — and the previous session-exchange path of #1170 isn't
a standard the OAuth ecosystem recognises. PKCE is the recommended flow for
public/native clients (OAuth 2.1 §4.1, IETF BCP 212).

**How:** the implementation mirrors what
[`suitenumerique/dictaphone`](https://github.com/suitenumerique/dictaphone)
already ships in production. We port its `core/authentication` module almost
verbatim: same schemas, same view classes, same JWT settings. Web SSO is
untouched — PKCE only activates when `response_type=code` is on
`/api/v1.0/authenticate/`.

### What changes

- `djangorestframework-simplejwt` 5.5.1 + `token_blacklist` for rotating refresh tokens.
- Pydantic schemas validating `code_challenge` / `code_challenge_method` / `state` per RFC 7636.
- Three views in `core/authentication/views.py`:
  - request: caches the PKCE challenge in the session before SSO,
  - callback: issues a single-use auth code, redirects to `MOBILE_DEEP_LINK_SCHEME?code=…&state=…`,
  - exchange: verifies the verifier, deletes the code first (anti-oracle), returns a JWT pair.
- New routes: `POST /api/v1.0/oauth/token/`, `POST /api/v1.0/oauth/token/refresh/`, `GET /mobile-login`.
- Settings: swap `OIDC_AUTHENTICATE_CLASS` / `OIDC_CALLBACK_CLASS`, add `MOBILE_DEEP_LINK_SCHEME` and `AUTH_PKCE_CACHE_TTL_SECONDS`, add `SIMPLE_JWT` (10 min access, 7 day refresh, rotate + blacklist).
- 19 tests ported from dictaphone covering parameter validation, code single-use, verifier mismatch, deep-link redirect, refresh rotation, and non-PKCE fall-through.

## Test plan

- [ ] `pytest src/backend/core/tests/authentication/test_pkce_views.py` green (19 tests).
- [ ] Full backend suite still green (no regression on web SSO).
- [ ] End-to-end against a local Meet + Keycloak + visio-mobile dev build: SSO login → deep-link → `/oauth/token/` → Bearer call → `/oauth/token/refresh/`.
- [ ] Replay: second exchange of the same code is rejected.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,20 @@ and this project adheres to
 
 ## [Unreleased]
 
+### Added
+
+- ➕(backend) add djangorestframework-simplejwt 5.5.1 #1409
+- ✨(backend) add update_url_query_params helper #1409
+- ✨(backend) add Pydantic schemas for PKCE OAuth2 flow #1409
+- ✨(backend) add OAuth2 + PKCE views for native mobile/desktop login #1409
+- ✨(backend) wire PKCE auth + JWT for native app login #1409
+- ✨(backend) expose PKCE token endpoints + /mobile-login URL #1409
+- ✅(backend) port dictaphone PKCE view test suite (19 tests) #1409
+
+### Fixed
+
+- 🔒️(backend) restrict /mobile-login to GET only #1409
+
 ## [1.19.0] - 2026-06-04
 
 ### Added
--- a/src/backend/core/authentication/schemas.py
+++ b/src/backend/core/authentication/schemas.py
@@ -1,0 +1,26 @@
+"""Pydantic models for authentication-related operations."""
+
+from typing import Literal
+
+from pydantic import BaseModel, Field
+
+
+class PKCEAuthenticationRequestModel(BaseModel):
+    """Model for generating PKCE authentication requests."""
+
+    code_challenge: str = Field(
+        min_length=43, max_length=128, pattern=r"^[A-Za-z0-9\-_]+$"
+    )
+    code_challenge_method: Literal["S256"] = Field(
+        default="S256", description="Code challenge method for PKCE authentication"
+    )
+    state: str = Field(min_length=43, max_length=128, pattern=r"^[A-Za-z0-9\-_]+$")
+
+
+class PKCETokenExchangeModel(BaseModel):
+    """Model for exchanging PKCE tokens."""
+
+    code: str = Field(min_length=43, max_length=128, pattern=r"^[A-Za-z0-9\-_]+$")
+    code_verifier: str = Field(
+        min_length=43, max_length=128, pattern=r"^[A-Za-z0-9\-_]+$"
+    )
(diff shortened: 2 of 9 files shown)


FloChehab:
Hi, as quick notes :
* Logout is currently not super clean from Assistant Transcripts mobile apps.
* We are likely going to move this logic into Django LaSuite if we start to have more than 1 app that relies on it https://github.com/suitenumerique/django-lasuite

mmaudet:
Hi @FloChehab , thanks for your review!

On logout: could you share what specifically isn't clean on the
Assistant Transcripts side? We'd like to avoid reproducing the same
gap on Visio Mobile. Probably :
- `POST /oauth/logout/` that blacklists the refresh token
- redirect to Keycloak `end_session_endpoint`
- clear the Django session created by the SSO callback

Happy to add this in a follow-up PR rather than block this one or
in this PR if you'd prefer.

On django-lasuite: makes total sense. With Meet becoming the second
consumer, the extraction criterion is met. Happy to help on the
move. Let me know how you'd like to coordinate...

**Context on my side**: this PR is currently the blocker for the
official v1 release of Visio Mobile (Android + iOS + Desktop). The
native apps need a Meet instance that exposes the PKCE endpoints to
authenticate without a Django session cookie without this merged
upstream, I can't ship because users want to test with the official visio.numerique.gouv.fr meet

**How can I help to unblock the release?** Concretely I can:
- address any blocker you identify on this PR (logout, tests, code
  organization). Just point me at what needs to change
- own the django-lasuite extraction PR if that's the path you'd
  prefer before merging into Meet
- pair on a call to align on scope and timeline

What would help you most?

FloChehab:
Hello @mmaudet,

Sorry I am mostly oow until Wednesday.

Thanks for your detailed comment.
Regarding the logout, I don't have precise elements in mind but what your are saying seems right to me. 

About the process to move forward, I think @lebaudantoine should own the decision, given the impact on visio.numerique.gouv.fr you expect.

My understanding is that shipping this directly into Django la suite **might** be the easiest path forward. I'll ping Antoine internally to quickly get a strategy about this.

mickenordin:
I am VERY interested in this. I am one of the developers of an email client called Chithi: https://github.com/SUNET/chithi/ we have support for creating meetings in Zoom, Nextcloud Talk and Matrix currently, and since Sunet also has a Visio instance, it would be great to add the capability to create rooms for calendar events, but we would need PKCE for that to make sense.

Is there anything I can do to move this forward? Patches? Cheerleading? Anything?

lebaudantoine:
Hi @mickenordin,

We already support several calendar integrations (e.g., Outlook, with Thunderbird on the way) and provide a dedicated external API for custom integrations, supporting multiple authentication methods (Resource Server, API Key, JWT for add-ons, etc.).

Feel free to join our Matrix channel at https://matrix.to/#/#meet-official:matrix.org and tag me, I'd be happy to help!

Issue #1363, opened 2026-05-28
Saved language is overwritten by the browser locale on every page load


rezk2ll:
## Bug Report

**Problematic behavior**

Every page load pushes the browser's detected language back to the user endpoint, overwriting whatever the server had. When the user query resolves the app reads the current UI language, compares it against the server value, and PUTs the browser value when they differ.

A server-stored language is silently reverted whenever it disagrees with the browser.

**Expected behavior/code**

A saved language should persist across sessions. Signing in from a different browser locale should not overwrite it.

**Steps to Reproduce**

1. Pick French as the interface language while logged in. Server stores `fr-fr`.
2. In an incognito window with the browser locale set to German, sign in to the same account.
3. On load the incognito session PUTs `de-de`, overwriting the saved `fr-fr`.
4. From then on the saved language is whichever session refreshed last: refresh the original session and it counter-PUTs `fr-fr`, refresh the incognito one and it goes back to `de-de`.

**Environment**
- Meet version: 1.16.0 ( and main )
- Platform: any browser

**Possible Solution**

On load, adopt the server's language into i18n instead of pushing the browser value. Only PUT when the user picks a language in the settings UI.

PR #1414, opened 2026-06-08
🐛(frontend) stop overwriting saved language on page load

rezk2ll:
## Problem

Every page load pushes the browser's detected language back to the user endpoint, overwriting whatever the server had. A language saved in settings gets reverted whenever it disagrees with the browser locale, so signing in from a different browser locale silently changes the saved preference.

https://github.com/suitenumerique/meet/issues/1363

## Solution

Treat the saved language as the source of truth: adopt it into the UI on load, and PUT only when the user picks a language in the settings panel. Timezone keeps following the browser as before.

### DIFF
--- a/src/frontend/src/features/auth/api/useSyncUserPreferencesWithBackend.tsx
+++ b/src/frontend/src/features/auth/api/useSyncUserPreferencesWithBackend.tsx
@@ -1,47 +1,66 @@
 import { useMutation } from '@tanstack/react-query'
 import { keys } from '@/api/queryKeys'
-import { useEffect } from 'react'
+import { useEffect, useRef } from 'react'
 import { useTranslation } from 'react-i18next'
 import { queryClient } from '@/api/queryClient'
 import { updateUserPreferences } from './updateUserPreferences'
-import { convertToBackendLanguage } from '@/utils/languages'
+import {
+  convertToBackendLanguage,
+  convertToFrontendLanguage,
+  type BackendLanguage,
+} from '@/utils/languages'
 import { useUser } from './useUser'
 
 /**
- * Hook that synchronizes user browser preferences (language, timezone) with backend user settings.
- * Automatically updates backend when browser settings change for logged-in users.
+ * Synchronizes user preferences with the backend. The saved language is the
+ * source of truth: it is adopted into the UI on load and only pushed back when
+ * the user changes it in the settings panel. Timezone follows the browser.
  */
 export const useSyncUserPreferencesWithBackend = () => {
   const { i18n } = useTranslation()
   const { user, isLoggedIn } = useUser()
 
-  const { mutateAsync } = useMutation({
+  const lastServerLanguage = useRef<BackendLanguage | null>(null)
+
+  const { mutate } = useMutation({
     mutationFn: updateUserPreferences,
     onSuccess: (updatedUser) => {
       queryClient.setQueryData([keys.user], updatedUser)
+      lastServerLanguage.current = updatedUser.language
+    },
+    onError: (error) => {
+      console.error(error)
     },
   })
 
   useEffect(() => {
-    if (!user || !isLoggedIn) return
+    if (!user || !isLoggedIn || !user.language) return
 
-    const syncBrowserPreferencesToBackend = async () => {
-      const currentTimezone = Intl.DateTimeFormat().resolvedOptions().timeZone
-      const currentLanguage = convertToBackendLanguage(i18n.language)
-      if (
-        currentLanguage !== user.language ||
-        currentTimezone !== user.timezone
-      ) {
-        await mutateAsync({
-          user: {
-            id: user.id,
-            timezone: currentTimezone,
-            language: currentLanguage,
-          },
-        })
+    const serverLang = user.language
+    const frontendServerLang = convertToFrontendLanguage(serverLang)
+
+    if (lastServerLanguage.current !== serverLang) {
+      lastServerLanguage.current = serverLang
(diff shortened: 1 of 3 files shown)


src/frontend/src/features/auth/api/useSyncUserPreferencesWithBackend.tsx
```diff
@@ -1,47 +1,66 @@
...
+    const serverLang = user.language
+    const frontendServerLang = convertToFrontendLanguage(serverLang)
+
+    if (lastServerLanguage.current !== serverLang) {
+      lastServerLanguage.current = serverLang
+      if (frontendServerLang && frontendServerLang !== i18n.language) {
+        i18n.changeLanguage(frontendServerLang)
+        return
       }
     }
```
davd-gzl, line 48: Critical: a new account carries [`en-us`](https://github.com/rezk2ll/meet/blob/fix/stop-overwriting-saved-language/src/backend/meet/settings.py#L214), [the column default](https://github.com/rezk2ll/meet/blob/fix/stop-overwriting-saved-language/src/backend/core/models.py#L172-L178), so a first sign-in comes up in English whatever the browser's language. The column cannot record that nobody chose a language, so the app cannot separate that default from a deliberate choice of English.

a French browser signing in for the first time, with the browser's language, the language on the account and the language on screen at each step

src/frontend/src/features/auth/api/useSyncUserPreferencesWithBackend.tsx
```diff
@@ -1,47 +1,66 @@
...
+
+  const { mutate } = useMutation({
     mutationFn: updateUserPreferences,
     onSuccess: (updatedUser) => {
       queryClient.setQueryData([keys.user], updatedUser)
+      lastServerLanguage.current = updatedUser.language
+    },
+    onError: (error) => {
+      console.error(error)
     },
```
davd-gzl, line 33: Nothing retries a save the server refused, so the next page load puts the old language back and the only trace is one line in the console.

src/frontend/src/features/auth/api/useSyncUserPreferencesWithBackend.tsx
```diff
@@ -1,47 +1,66 @@
...
-        currentTimezone !== user.timezone
-      ) {
-        await mutateAsync({
-          user: {
-            id: user.id,
-            timezone: currentTimezone,
-            language: currentLanguage,
-          },
-        })
+    if (!user || !isLoggedIn || !user.language) return
```
davd-gzl, line 37: Nit: `!user.language` can never be true, since the [column is non-null](https://github.com/rezk2ll/meet/blob/fix/stop-overwriting-saved-language/src/backend/core/models.py#L172-L178) and [`ApiUser`](https://github.com/rezk2ll/meet/blob/fix/stop-overwriting-saved-language/src/frontend/src/features/auth/api/ApiUser.ts#L8) types the field as a language code. An account arriving without a language would also lose the timezone sync, since this line returns before it.

src/frontend/src/features/auth/api/useSyncUserPreferencesWithBackend.tsx
```diff
@@ -1,47 +1,66 @@
...
+    if (!languageChanged && !timezoneChanged) return
+
+    mutate({
+      user: {
+        id: user.id,
+        timezone: currentTimezone,
+        language: languageChanged ? uiLang : serverLang,
+      },
+    })
+  }, [i18n, i18n.language, isLoggedIn, user, mutate])
```
davd-gzl, line 65: Nit: `i18n` is the [instance held in context](https://github.com/i18next/react-i18next/blob/v15.1.1/src/useTranslation.js#L35) rather than a new object per render, so naming it in the dependency list beside `i18n.language` changes nothing.

src/frontend/src/utils/languages.ts
```diff
@@ -12,6 +12,14 @@ const frontendToBackendMap: Record<FrontendLanguage, BackendLanguage> = {
 
 export const convertToBackendLanguage = (
   frontendLang: string = 'fr'
-): BackendLanguage => {
+): BackendLanguage | undefined => {
   return frontendToBackendMap[frontendLang as FrontendLanguage]
 }
```
davd-gzl, line 17: Suggestion: the `'fr'` default is a second copy of [`fallbackLng`](https://github.com/rezk2ll/meet/blob/fix/stop-overwriting-saved-language/src/frontend/src/i18n/init.ts#L6), which the i18n setup declares and does not export, so the fallback language can be changed in one place and left wrong in the other.

davd-gzl (changes requested):

src/frontend/src/features/auth/api/useSyncUserPreferencesWithBackend.tsx
rezk2ll: You're right that i18n is the context instance and listing it changes nothing at runtime, but it still can't come out: the effect calls i18n.changeLanguage, so `react-hooks/exhaustive-deps` reports missing dependency: 'i18n' and the CI goes red.

I tried it to be sure. i18n.language has to stay as well; otherwise, the effect stops re-running when the user picks a language, and the save never fires. So I'm keeping both.

src/frontend/src/utils/languages.ts
rezk2ll: Agreed, it's two copies of the same constant. The problem is the import direction.

The clean version is the reverse, with the util owning the constant and init.ts importing it, but that is a refactor unrelated to this bug. Better as its own PR.

PR #1415, opened 2026-06-10
✨(backend) Audit logging [WIP]

cameledev:
## Purpose

[Description and CSIRT recommendations](https://github.com/orgs/suitenumerique/projects/11/views/1?filterQuery=assignee%3Acameledev&pane=issue&itemId=168650269)

## Proposal

- create `src/backend/core/audit.py` meant to be ported to `django-lasuite` when approved
- `src/backend/core/audit.py` from which you can import a logger (`getLogger`) and a json formatter (`AuditJsonFormatter`)
- The audit logger requires a mandatory `request` variable (may be None). When passed, multiple information are derived and added to logs (`source_ip`, `source_port` etc.)
- additional kwargs can be passed
-  getLogger may be overriden with project specific serializers (see `src/backend/core/audit_meet.py`). For example, we can pass `logger.info(..., original_room=room)`, and `room.slug`, `room.name` are automatically added to the log (prefixed with the variable name, e.g. `original_room.slug`)

TODO: 
- complete audit logging coverage, when architecture is approved

### DIFF
--- a/src/backend/core/audit.py
+++ b/src/backend/core/audit.py
@@ -1,0 +1,194 @@
+"""Structured audit logging."""
+
+# Audit helpers intentionally expose many optional keyword fields.
+# pylint: disable=R0913,R0917
+# ruff: noqa: PLR0913
+
+import json
+import logging
+from datetime import datetime, timezone
+from functools import partialmethod
+from typing import TYPE_CHECKING, Any, Protocol
+
+from django.http import HttpRequest
+
+AUDIT_LOGGER_NAME = "audit"
+
+
+def resolve_source_ip(request: HttpRequest):
+    """Return the best-effort client IP for ``request``.
+
+    Reads the original client from ``X-Forwarded-For`` when present,
+    falling back to ``REMOTE_ADDR``.
+
+    NB: behind a proxy/load-balancer chain, correctness depends on the ingress
+    being configured to set and trust ``X-Forwarded-For``. Confirm the
+    trusted-proxy chain before relying on this value for security decisions.
+    """
+    forwarded = request.META.get("HTTP_X_FORWARDED_FOR")
+    if forwarded:
+        return forwarded.split(",")[0].strip()
+    return request.META.get("REMOTE_ADDR")
+
+
+def extract_request_fields(request: HttpRequest) -> dict[str, Any]:
+    """Return the audit fields derivable from ``request``."""
+    meta = request.META
+    return {
+        "source_ip": resolve_source_ip(request),
+        "source_port": meta.get("REMOTE_PORT"),
+        "request_path": request.path,
+        "request_url": request.build_absolute_uri(),
+        "request_method": request.method,
+        "request_body_bytes": meta.get("CONTENT_LENGTH"),
+        "http_version": meta.get("SERVER_PROTOCOL"),
+        "user_agent": meta.get("HTTP_USER_AGENT"),
+        "referer": meta.get("HTTP_REFERER"),
+        "server_user": meta.get("REMOTE_USER"),
+    }
+
+
+class AuditJsonFormatter(logging.Formatter):
+    """Render audit records as single-line JSON.
+
+    Read the structured payload attached to the record under ``audit`` and
+    wrap it in a small envelope.
+    """
+
+    def format(self, record):
+        payload = {"log_type": "audit"}
+
+        audit = getattr(record, "audit", None)
+        if isinstance(audit, dict):
+            payload.update(audit)
+        else:
+            payload["event_type"] = record.getMessage()
+
+        payload.setdefault(
+            "timestamp",
+            datetime.fromtimestamp(record.created, tz=timezone.utc).isoformat(),
+        )
+        payload["level"] = record.levelname
+        payload["logger"] = record.name
+
+        # ``default=str`` serialises UUIDs, datetimes, etc.; ``ensure_ascii``
+        # off keeps emails and non-ASCII identifiers readable.
+        return json.dumps(payload, default=str, ensure_ascii=False)
+
+
+class _AuditEmit(Protocol):
+    """Public signature shared by the per-level audit methods.
+
+    Declared so editors and type checkers see the real keyword fields despite using `partialwrapper`.
+    """
+
+    def __call__(
+        self,
+        event_type: str,
+        *,
+        auth_type: str | None = ...,
+        actor: dict[str, Any] | None = ...,
+        source_ip: str | None = ...,
+        target: dict[str, Any] | None = ...,
+        request: HttpRequest | None = ...,
+        **extra: Any,
+    ) -> None:
+        pass
+
+
+class AuditLogger:
+    """Wrapper around the named ``audit`` logger."""
+
+    def __init__(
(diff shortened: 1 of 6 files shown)


src/backend/core/audit.py
```diff
@@ -0,0 +1,194 @@
...
+        debug = partialmethod(_emit, logging.DEBUG)
+        info = partialmethod(_emit, logging.INFO)
+        warning = partialmethod(_emit, logging.WARNING)
+        error = partialmethod(_emit, logging.ERROR)
+        critical = partialmethod(_emit, logging.CRITICAL)
+        exception = partialmethod(_emit, logging.ERROR, exc_info=True)
+
+
+def getLogger(
+    name: str | None = None, custom_serializers: list[tuple | None] = None
```
lebaudantoine, line 174: ?
```suggestion
    name: str | None = None, custom_serializers: list[tuple] | None = None
```

src/backend/core/audit.py
```diff
@@ -0,0 +1,194 @@
...
+        self._logger = logging.getLogger(logger_name)
+        self._serializers = []
+        if custom_serializers is not None:
+            self._serializers = custom_serializers
+
+    def _emit(
+        self,
+        level,
+        event_type,
+        request: HttpRequest | None,  # Intentionnaly mandatory
```
lebaudantoine, line 116: The comment is contradicting the typing, why?
*intentionaly

src/backend/core/audit.py
```diff
@@ -0,0 +1,194 @@
...
+        critical: _AuditEmit
+        # ``exception`` mirrors stdlib: ERROR level with the active traceback.
+        exception: _AuditEmit
+    else:
+        debug = partialmethod(_emit, logging.DEBUG)
+        info = partialmethod(_emit, logging.INFO)
+        warning = partialmethod(_emit, logging.WARNING)
+        error = partialmethod(_emit, logging.ERROR)
+        critical = partialmethod(_emit, logging.CRITICAL)
+        exception = partialmethod(_emit, logging.ERROR, exc_info=True)
```
lebaudantoine, line 170: What would you think about just spelling these out as plain methods, like CPython does for [Logger.debug/info/warning/...](https://github.com/python/cpython/blob/912a68b639882aa11c10636117ff0dda3bd6b1fc/Lib/logging/__init__.py#L1463), thin wrappers that each delegate to self._log(LEVEL, ...)?

My worry is that the partialmethod + Protocol + TYPE_CHECKING approach doesn't really remove the duplication, just moves it into the typing layer where it's harder to follow, the method names still appear twice, and only the body is actually deduplicated (which a plain delegating method does too, since _emit holds the real work).

Since this is headed for django-lasuite, I'd lean toward keeping it stupid-simple: plain defs give honest signatures and tracebacks for free (the thing the Protocol is there to recover) and sidestep stub/runtime drift. The levels are fixed by stdlib, so it's about as stable as duplication gets. Curious what drove the current choice though, is there a constraint I'm missing?

src/backend/core/external_api/viewsets.py
```diff
@@ -86,10 +85,11 @@ def generate_jwt_access_token(self, request, *args, **kwargs):
             )
 
         if not application.can_delegate_email(email):
-            logger.warning(
-                "Application %s denied delegation for %s",
-                application.client_id,
-                email,
+            audit_logger.warning(
+                "Application denied delegation",
```
lebaudantoine, line 89: Looks like we're mixing up event message and event type here. Bit confused tbh, is passing just the event type (snake_case) what's expected? And should we list allowed event types in a data structure somewhere, or is that overkill?

src/backend/core/audit.py
```diff
@@ -0,0 +1,194 @@
...
+        pass
+
+
+class AuditLogger:
+    """Wrapper around the named ``audit`` logger."""
+
+    def __init__(
+        self,
+        logger_name=AUDIT_LOGGER_NAME,
+        custom_serializers: list[tuple] | None = None,
```
FloChehab, line 105: Could this be typed better or documented ? list[tuple] ?

src/backend/core/audit.py
```diff
@@ -0,0 +1,194 @@
...
+    else:
+        debug = partialmethod(_emit, logging.DEBUG)
+        info = partialmethod(_emit, logging.INFO)
+        warning = partialmethod(_emit, logging.WARNING)
+        error = partialmethod(_emit, logging.ERROR)
+        critical = partialmethod(_emit, logging.CRITICAL)
+        exception = partialmethod(_emit, logging.ERROR, exc_info=True)
+
+
+def getLogger(
```
FloChehab, line 173: I'd rename this to get_audit_logger, and maybe make the AuditLogger class private ?

src/backend/core/audit.py
```diff
@@ -0,0 +1,194 @@
...
+
+    Names are nested under ``AUDIT_LOGGER_NAME`` (e.g. ``audit.core.foo``) so
+    they inherit its handlers through the standard logging hierarchy, keeping
+    the module visible in the ``logger`` field of the emitted JSON.
+    """
+    if not name or name == AUDIT_LOGGER_NAME:
+        return AuditLogger(AUDIT_LOGGER_NAME, custom_serializers=custom_serializers)
+    return AuditLogger(
+        f"{AUDIT_LOGGER_NAME}.{name}", custom_serializers=custom_serializers
+    )
```
FloChehab, line 194: You should add an `__ALL__ = ...` to this module

src/backend/core/audit.py
FloChehab: There is no  = None so it's valid from my pov

src/backend/core/audit.py
```diff
@@ -0,0 +1,194 @@
...
+
+    def __init__(
+        self,
+        logger_name=AUDIT_LOGGER_NAME,
+        custom_serializers: list[tuple] | None = None,
+    ):
+        self._logger = logging.getLogger(logger_name)
+        self._serializers = []
+        if custom_serializers is not None:
+            self._serializers = custom_serializers
```
FloChehab, line 110: You should build a mapping of class -> serializer method, so that later you can make direct lookups in 0(1) based on the type of the item. (and fallback to 0(n) if it's a subclass of a class you have handling for).

It might also be possible to type this in a generic way, so that if we try to pass something that is not supported we get typehint feedback.

You should crash early if you have multiple sirializers for the same kind of object.

src/backend/core/audit.py
```diff
@@ -0,0 +1,194 @@
...
+        for key, value in extra.items():
+            for cls, serializer in self._serializers:
+                if isinstance(value, cls):
+                    new_extra = new_extra | {
+                        f"{key}.{ser_key}": ser_field
+                        for ser_key, ser_field in serializer(value).items()
+                    }
+                    break
+            else:
+                new_extra[key] = value
```
FloChehab, line 141: This is not very optimal perf wise, see my comment above.

FloChehab (commented):
Few things by reading this :)

At first I was like, why not just configure specific handle for audit loggers whose name start with `audit` but I do understand the usefullness of a dedicated class for this.

PR #1416, opened 2026-06-10
🐛(frontend) fix noise reduction left-channel-only audio

cameledev:
Fix bug with RNNoise noise reduction which interprets mono input as left channel with some browsers.

https://github.com/suitenumerique/meet/issues/764

Thanks to @chrnin for the insight.


lebaudantoine (approved):
tested locally. It did not raise any error in the console.
Let's deploy it in staging and try it.

PR #1420, opened 2026-06-12
✨(summary) generalized stt api call

FloChehab:
Commits are self explanatory.

Main changes:
* CI : push images on integration branches to ease tests
* Relaxed WhisperX pydantic model to allow the use of AlbertAPI, which doesn't return word_segments now
* Make sure to make the expected API future API call (meet whisperX doesn't care about it, albert / does and will).
* Had to change the way we call the STT API to avoid openai lib casting the result to an unwanted format.

**:warning: On staging we will loose word level segments which impacts transcripts and may impact code by @Leobouloc regarding speaker assignment. This will also introduce a difference between staging and prod. But there is a GPU at stake.**


src/summary/summary/core/celery_worker.py
```diff
@@ -130,8 +128,47 @@ def transcribe_audio(
...
+                    "timestamp_granularities": ["word", "segment"],
+                    "response_format": "diarized_json",
+                },
+                files={"file": audio_file},
+                headers={"Authorization": f"Bearer {api_key}"},
+                # Mimic OpenAI's timeout settings
+                timeout=(60, 10 * 60),
+            )
+            if 200 < res.status_code < 300:
+                raise RuntimeError("WhisperX transcription failed, %s", res.text)
```
lebaudantoine: I might not have understood everything, but isn't weird to raise a runtime error for error codes that are considered successful ? https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Status#successful_responses

src/summary/summary/core/celery_worker.py
```diff
@@ -101,11 +104,6 @@ def transcribe_audio(
         )
 
     logger.info("Initiating WhisperX client")
-    whisperx_client = openai.OpenAI(
-        api_key=settings.whisperx_api_key.get_secret_value(),
-        base_url=settings.whisperx_base_url,
-        max_retries=settings.whisperx_max_retries,
```
lebaudantoine, line 107: we should cleanup this setting in docs/features/transcription

src/summary/summary/core/celery_worker.py
```diff
@@ -130,8 +128,47 @@ def transcribe_audio(
...
-                model=settings.whisperx_asr_model, file=audio_file, language=language
+
+            api_key = settings.whisperx_api_key.get_secret_value()
+            base_url = settings.whisperx_base_url
+
+            # We use a manual call to the transcripion endpoint to avoid
+            # the OpenAI lib striping elements from the response, since
+            # we expect a non-standard format, and setting response_format
+            # to "diarized_json" makes the lib cast the payload to their
+            # standard.
```
lebaudantoine: Perhaps it would help if we could understand from this comment which elements were stripped, what the expected standard is, and how exactly it differs from our WhisperX API.

src/summary/summary/core/celery_worker.py
lebaudantoine: why not relying on `res.raise_for_status()` ?

src/summary/summary/core/celery_worker.py
```diff
@@ -130,8 +128,47 @@ def transcribe_audio(
...
+            # We remove the "usage" key from the transcription_json dictionary
+            # as it may cause issues with parsing inside the Transcription model
+            transcription_json.pop("usage", None)
+
+            # We force the use of the Transcription model here
+            # to avoid changing too much code for now.
+            # Note that it should be WhisperXResponse instead.
+            transcription = Transcription.model_validate(
+                # We add a dummy "text" to make the model validate
+                {**transcription_json, "text": ""},
```
lebaudantoine: Why doesn't it validate without this dummy text?

src/summary/summary/core/celery_worker.py
```diff
@@ -130,8 +128,47 @@ def transcribe_audio(
...
+                # Mimic OpenAI's timeout settings
+                timeout=(60, 10 * 60),
+            )
+            if 200 < res.status_code < 300:
+                raise RuntimeError("WhisperX transcription failed, %s", res.text)
+
+            transcription_json: dict[str, Any] = res.json()
+            # We remove the "usage" key from the transcription_json dictionary
+            # as it may cause issues with parsing inside the Transcription model
+            transcription_json.pop("usage", None)
```
lebaudantoine, line 168: I don't understand why, I'm not familiar with pydantic v2, but isn't usage supposed to be part of the Transcription model? Plus, the validate method seems to allow extra parameters.

src/summary/summary/core/celery_worker.py
```diff
@@ -130,8 +128,47 @@ def transcribe_audio(
...
+                    "model": settings.whisperx_asr_model,
+                    "language": language,
+                    "timestamp_granularities": ["word", "segment"],
+                    "response_format": "diarized_json",
+                },
+                files={"file": audio_file},
+                headers={"Authorization": f"Bearer {api_key}"},
+                # Mimic OpenAI's timeout settings
+                timeout=(60, 10 * 60),
+            )
```
lebaudantoine, line 158: I guess we just dropped the retry behavior that could have been used by people self-hosting the service, but I ack and accept it.

src/summary/summary/core/celery_worker.py
FloChehab: I should have put 400, I'll double check that

src/summary/summary/core/celery_worker.py
FloChehab: i'll clarify the comment

src/summary/summary/core/celery_worker.py
FloChehab: Yup

lebaudantoine (approved):

PR #1426, opened 2026-06-15
🐛(frontend) fix metadata agent collector enabled check

lebaudantoine:
Fix an incorrect check that caused issues in production when determining whether the metadata agent collector is enabled.


lebaudantoine:
@coderabbitai full review

PR #1435, opened 2026-06-22
🐛(frontend) prefill prejoin name from the user profile

maartendra:
## Purpose

On the prejoin screen, the display-name field is sourced **only** from `userChoicesStore` (persisted to `localStorage` via LiveKit's `loadUserChoices`). Nothing seeds it from the authenticated user's profile, so a logged-in user who hasn't manually typed a name is shown an **empty, required** field and asked to enter their name on **every join / refresh** — even though the backend already knows them and `useUser()` (`/users/me` → `full_name`) exposes the name client-side.

Reported by a self-hosting deployment using OIDC (Authentik): users authenticate, `meet_user.full_name` is populated, yet the prejoin still demands a name on each refresh.

## Changes

In `src/features/rooms/components/Join.tsx`:

- Seed `username` from `user.full_name` when the user **is logged in** and **hasn't set one yet**, via `saveUsername(...)`. Writing it to the store (not just the field) means the room/lobby requests (`fetchRoom`, `useLobby`) and LiveKit connect use it too.
- Seed **once** (guarded by a ref) so it never fights a manual edit or a deliberate clear.
- Done in an effect because `useUser()` can resolve **after** the prejoin mounts (e.g. a cold load straight to a room URL).
- Switch the name `Field` from `defaultValue` to a controlled `value={username}` so a profile that resolves after mount is actually reflected (a `defaultValue` only applies on first mount).

**Guests are unaffected** — `isLoggedIn` is false, so nothing is seeded and the field behaves exactly as before.

## Checks

- `npm run lint` (eslint, `--max-warnings 0`) ✅
- `panda codegen && tsc -b` (typecheck) ✅
- Changelog entry added under `[Unreleased] / Fixed`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,10 @@ and this project adheres to
 
 - ✨(summary) generalized stt api call #1420
 
+### Fixed
+
+- 🐛(frontend) prefill prejoin name from the user profile #1435
+
 ## [1.21.0] - 2026-06-15
 
 ### Added
--- a/src/frontend/src/features/rooms/components/Join.tsx
+++ b/src/frontend/src/features/rooms/components/Join.tsx
@@ -33,6 +33,7 @@ import { ApiLobbyStatus, type ApiRequestEntry } from '../api/requestEntry'
 import { Spinner } from '@/primitives/Spinner'
 import { ApiAccessLevel } from '../api/ApiRoom'
 import { useLoginHint } from '@/hooks/useLoginHint'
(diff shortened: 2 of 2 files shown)


lebaudantoine:
Thanks for your contribution, I've addressed this fix in #1499, with deeper changes.

maartendra:
Thanks @lebaudantoine, that's great — #1499 looks like a more complete take (nice to have it as a config toggle). Happy to see it handled upstream; I'll follow #1499 and drop our local workaround once it lands. 🙏

PR #1443, opened 2026-06-23
📝(docs) document rebranding the favicon via a volume mount

maartendra:
## Purpose

The favicon is baked into the frontend image, so rebranding Meet (logo, name, theme) appears to require forking and rebuilding the frontend just to change the tab icon.

Originally this PR added a runtime `FRONTEND_FAVICON_URL` config knob (mirroring `custom_css_url`). Per review feedback from @lebaudantoine, that approach is the wrong fit for favicons: the bundled `<link rel="icon">` lives in the static `index.html`, so the browser fetches it during HTML parse — before the JS swap can run — giving a visible favicon flash on every load, and the swap only covered the `rel~="icon"` PNGs (not `apple-touch-icon`, the `android-chrome`/PWA icons, `favicon.ico`, or the manifest).

## Changes

Reworked to **docs-only**. Since the icons are served from stable, unhashed paths under `/usr/share/nginx/html`, they can simply be overlaid with a volume — no code, no rebuild, no flash, and every variant is covered.

`docs/installation/kubernetes.md` now has a **"Rebranding the favicon"** section showing the recipe: a `ConfigMap` (`binaryData`) of replacement icons mounted over the bundled ones via the chart's `frontend.extraVolumes` / `frontend.extraVolumeMounts` (per-file `subPath`).

No code changes; backwards-compatible; no i18n impact.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ and this project adheres to
 
 ### Added
 
+- 📝(docs) document rebranding the favicon via a volume mount #1443
 - ✨(backend) add command to clean pending and deleted files
 - 🧱(helm) run clean files command as cronjob
 
--- a/docs/installation/kubernetes.md
+++ b/docs/installation/kubernetes.md
@@ -244,6 +244,69 @@ meet-admin               <none>   meet.127.0.0.1.nip.io       localhost   80, 44
 
 You can use LaSuite Meet on https://meet.127.0.0.1.nip.io from the local device. The provisioning user in keycloak is meet/meet.
 
+## Rebranding the favicon
+
+The favicon is bundled into the frontend image and served as a set of static
+files from `/usr/share/nginx/html` (`favicon.ico`, `favicon-16x16.png`,
+`favicon-32x32.png`, `apple-touch-icon.png`, `android-chrome-192x192.png`,
+`android-chrome-512x512.png`, `icon.png`). To rebrand without forking and
+rebuilding the image, overlay your own icons onto those paths with a volume —
+this serves the right icon from the first byte (no rebuild, no flash) and
+covers every variant, including the iOS home-screen and Android/PWA icons.
+
+Put your icons in a `ConfigMap` (`binaryData` keeps the PNGs intact)…
+
+```yaml
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: meet-favicon
+binaryData:
+  # base64 of each replacement icon
+  favicon.ico: <base64…>
+  favicon-16x16.png: <base64…>
+  favicon-32x32.png: <base64…>
+  apple-touch-icon.png: <base64…>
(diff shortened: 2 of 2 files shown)


lebaudantoine:
Thanks for this, the rebranding pain point is real and worth solving.

One question on the approach, though: why surface the favicon through the frontend config and swap it at runtime, rather than just mounting a volume over the bundled favicon when running the image? A bind mount keeps the icon fully static and served correctly from the first byte, with no extra setting to surface and no new code path to maintain.

The runtime swap also has a visible side effect: because the `<link rel="icon">` is replaced in `AppInitialization` only after the frontend has rendered, the browser fetches and shows the bundled icon first, then flips to the configured one, a noticeable favicon blink on every load.

I understand the motivation for matching the existing `custom_css_url` pattern for consistency. If we do want to keep the dynamic approach, could the swap happen earlier (before first paint) so we avoid the flash? Otherwise, mounting the favicon as a volume seems like the simpler and flicker-free option.

maartendra:
Thank you — this is exactly the right call, and you convinced me on both counts.

On the flash: you're right that it's inherent to my approach, not just a matter of swapping earlier. The bundled `<link rel="icon">` is in the static `index.html`, so the browser requests it during HTML parse, before any JS runs — and `favicon_url` only arrives via the async frontend-config fetch, so there's no "before first paint" moment to hook into. The blink can't really be designed away here.

Two things you said also made me realise the runtime swap was *worse* than I'd thought, not just flashier: it only touched the `rel~="icon"` PNGs, so a rebranded instance would still ship the stock `apple-touch-icon`, the `android-chrome` PWA icons, `favicon.ico`, and the manifest icons. A volume mount covers all of them, and serves the right icon from the first byte.

And the clincher is that the icons are referenced by stable, unhashed paths (`/favicon-32x32.png`, `/apple-touch-icon.png`, …) served from `/usr/share/nginx/html` — so the mount needs no code at all.

So I've dropped the backend/frontend changes and reworked this into a docs-only PR: a **"Rebranding the favicon"** section in `kubernetes.md` showing the `ConfigMap` + `frontend.extraVolumes` / `extraVolumeMounts` recipe. Same rebranding-without-forking outcome, flicker-free, nothing new to maintain. Thanks for the thoughtful steer!

---

_Update: merged as the docs-only change — the **“Rebranding the favicon”** section in `kubernetes.md` (`ConfigMap` + `frontend.extraVolumes`/`extraVolumeMounts`). No backend/frontend code in the end. Thanks again @lebaudantoine for steering it there._

lebaudantoine:
I'm curious are you working on self hosting LaSuite Meet, let's connect, over Matrix or linkedin.
best

maartendra:
> I'm curious are you working on self hosting LaSuite Meet, let's connect, over Matrix or linkedin. best

Yes, self-hosted. We have added LaSuite Meet to https://email.eu. I have added you on LinkedIn!

PR #1444, opened 2026-06-24
✨ Restrict settable room access levels

### LINKED ISSUE #1076: Ability to restrict access levels
## Feature Request

phanky1:
## Purpose

Instance operators can set the default room access level, but room owners and admins can still create or update a room with any access level, including `public`, so operators can't enforce an instance-wide policy such as forbidding fully public rooms.

## Proposal

Add a `RESOURCE_ALLOWED_ACCESS_LEVELS` allow-list, defaulting to all current levels, enforced through the room API on create and update. The allow-list is exposed via the config endpoint, and the room access settings hide the options that aren't allowed.

Closes #1076

### DIFF
--- a/src/backend/core/api/__init__.py
+++ b/src/backend/core/api/__init__.py
@@ -62,6 +62,9 @@ def get_frontend_configuration(request):
         },
         "telephony": build_telephony_config(),
         "subtitle": {"enabled": settings.ROOM_SUBTITLE_ENABLED},
+        "room": {
+            "allowed_access_levels": settings.RESOURCE_ALLOWED_ACCESS_LEVELS,
+        },
         "livekit": {
             "url": settings.LIVEKIT_CONFIGURATION["url"],
             "force_wss_protocol": settings.LIVEKIT_FORCE_WSS_PROTOCOL,
--- a/src/backend/core/api/serializers.py
+++ b/src/backend/core/api/serializers.py
@@ -142,6 +142,14 @@ class RoomSerializer(serializers.ModelSerializer):
             raise serializers.ValidationError(e.errors()) from e
         return value
 
+    def validate_access_level(self, access_level):
+        """Validate the access level against the instance allow-list."""
+        if access_level not in settings.RESOURCE_ALLOWED_ACCESS_LEVELS:
+            raise serializers.ValidationError(
+                "This access level is not allowed on this instance."
+            )
+        return access_level
+
     def to_representation(self, instance):
         """
         Add users only for administrator users.
--- a/src/backend/core/tests/rooms/test_api_rooms_create.py
+++ b/src/backend/core/tests/rooms/test_api_rooms_create.py
@@ -9,7 +9,7 @@ import pytest
 from rest_framework.test import APIClient
 
 from ...factories import RoomFactory, UserFactory
-from ...models import Room
+from ...models import Room, RoomAccessLevel
 
 pytestmark = pytest.mark.django_db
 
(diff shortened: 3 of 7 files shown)


src/backend/core/api/serializers.py
```diff
@@ -142,6 +142,14 @@ def validate_configuration(self, value):
             raise serializers.ValidationError(e.errors()) from e
         return value
 
+    def validate_access_level(self, access_level):
+        """Validate the access level against the instance allow-list."""
+        if access_level not in settings.RESOURCE_ALLOWED_ACCESS_LEVELS:
+            raise serializers.ValidationError(
+                "This access level is not allowed on this instance."
+            )
+        return access_level
```
davd-gzl, line 151: The external API for third-party integrations (e.g. an Outlook add-on) never reaches this check. It creates rooms at its own default, [`EXTERNAL_API_DEFAULT_ACCESS_LEVEL`](https://github.com/phanky1/meet/blob/feat/restrict-access-levels/src/backend/core/external_api/serializers.py#L55-L68), and ignores `RESOURCE_ALLOWED_ACCESS_LEVELS`. Set the allowlist to invited-only and its rooms stay open to anyone signed in. Check the allowlist here too.

src/backend/meet/settings.py
```diff
@@ -662,6 +662,11 @@ class Base(Configuration):
     RESOURCE_DEFAULT_ACCESS_LEVEL = values.Value(
         "public", environ_name="RESOURCE_DEFAULT_ACCESS_LEVEL", environ_prefix=None
     )
+    RESOURCE_ALLOWED_ACCESS_LEVELS = values.ListValue(
+        ["public", "trusted", "restricted"],
+        environ_name="RESOURCE_ALLOWED_ACCESS_LEVELS",
+        environ_prefix=None,
+    )
```
davd-gzl, line 669: Nothing checks that this list holds real access levels. A typo like `trusted,restrcited` passes the boot guard, which only asks that the default is in the list, and drops invited-people-only from what an administrator can choose.

src/backend/core/api/__init__.py
```diff
@@ -62,6 +62,9 @@ def get_frontend_configuration(request):
         },
         "telephony": build_telephony_config(),
         "subtitle": {"enabled": settings.ROOM_SUBTITLE_ENABLED},
+        "room": {
+            "allowed_access_levels": settings.RESOURCE_ALLOWED_ACCESS_LEVELS,
+        },
```
davd-gzl, line 67: Missing test: nothing asserts that `/api/v1.0/config/` carries `room.allowed_access_levels`. That key can go away unnoticed, and the picker then falls back to offering all three levels.

src/backend/core/tests/rooms/test_api_rooms_create.py
```diff
@@ -65,6 +65,30 @@ def test_api_rooms_create_authenticated(reset_cache):
...
+        "/api/v1.0/rooms/",
+        {
+            "name": "my room",
+            "access_level": RoomAccessLevel.PUBLIC,
+        },
+        format="json",
+    )
+
+    assert response.status_code == 400
+    assert not Room.objects.exists()
```
davd-gzl, line 89: Missing test: `test_api_rooms_create_authenticated` above omits `access_level` but never asserts the level the room lands on. Nothing checks that an omitted level takes `RESOURCE_DEFAULT_ACCESS_LEVEL`, the value the boot-time guard keeps inside the allowlist.

src/frontend/src/features/rooms/livekit/components/Admin.tsx
```diff
@@ -201,23 +231,7 @@ export const Admin = () => {
               })
               .catch((e) => console.error(e))
```
davd-gzl, line 232: Suggestion: the picker reads its list once per page load, so a tab open across a change to `RESOURCE_ALLOWED_ACCESS_LEVELS` keeps offering a level the server now refuses. The 400 reaches `console.error` alone: the radio never moves and nothing on screen says the meeting is unchanged. Show the refusal here, and refetch the list on it.

a room administrator picking a level the server no longer allows: the picker offers three while the server allows two, the write comes back 400, the radio stays where it was, and a reload drops the level from the list

davd-gzl (commented):
The new environment variable has no row in [`docs/installation/kubernetes.md`](https://github.com/phanky1/meet/blob/feat/restrict-access-levels/docs/installation/kubernetes.md?plain=1#L340), while the two settings beside it do.

davd-gzl:
Hi @phanky1, do you plan to finish this PR? If you don't, can I take the ownership of it?
Outside of my remarks, your work LGTM

phanky1:
> Hi @phanky1, do you plan to finish this PR? If you don't, can I take the ownership of it? Outside of my remarks, your work LGTM

go for it, thank you.

PR #1453, opened 2026-06-29
🚸(frontend) use "Advanced" instead of "Premium" in the recording sidepanel

lebaudantoine:
Following internal feedback, rename the "Premium" wording to "Advanced" across the transcription and recording sidepanels, so the label no longer implies a paid tier.


lebaudantoine:
@coderabbitai full review

PR #1459, opened 2026-06-30
♿️(frontend) make fullscreen share warning keyboard accessible

Ovgodd:
## Purpose

The fullscreen screen share warning was not accessible: keyboard users could not reach its actions and screen readers were not notified when it appeared.

## Proposal

- [x] Improve `FullScreenShareWarning` with `role="alert"`, autoFocus on "Stop presenting", and `Escape` to dismiss.


lebaudantoine (commented):
Instead of a hook I would autofocus the button, this should work similarly.

Ovgodd:
> Instead of a hook I would autofocus the button, this should work similarly.

indeed ! that works well

lebaudantoine (approved):

PR #1460, opened 2026-06-30
(agents) Add sentry monitoring

cameledev:
## Purpose

`agents` section had no Sentry observability. Add sentry.

## Proposal

- [x] slightly refactor tasks closing to uniformize and allow better handling of errors by Sentry (create new  `tasks.py`). Otherwise, errors in callback would be raised as `exception in callback`
- [x] add Sentry observability

## TODO:

- [x] Test that errors surface in sentry from within agent


src/agents/multi_user_transcriber.py
```diff
@@ -99,24 +102,29 @@ def on_participant_connected(self, participant: rtc.RemoteParticipant):
         logger.info(f"starting session for {participant.identity}")
         task = asyncio.create_task(self._start_session(participant))
         self._tasks.add(task)
-
-        def on_task_done(task: asyncio.Task):
-            try:
-                self._sessions[participant.identity] = task.result()
```
cameledev, line 105: moved into `_start_session`

src/agents/observability.py
```diff
@@ -0,0 +1,96 @@
...
+
+    Args:
+        agent_name: Identifier of the agent, attached as a tag to Sentry issues
+    """
+    # Read the DSN at call time so it picks up variables that load_dotenv()
+    # populated after this module was first imported.
+    sentry_dsn = os.getenv("SENTRY_DSN")
+    if not sentry_dsn:
+        logger.warning("SENTRY_DSN not defined for agent '%s'", agent_name)
+        return
```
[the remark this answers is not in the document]
lebaudantoine: Agree

src/agents/observability.py
```diff
@@ -0,0 +1,96 @@
...
+        environment=os.getenv("SENTRY_ENVIRONMENT"),
+        release=get_release(),
+        debug=False,
+        integrations=[
+            # Capture log records emitted at ERROR and above as Sentry events.
+            # This covers the agents' explicit logger.exception(...) calls as
+            # well as asyncio's "Exception in callback" / "Task exception was
+            # never retrieved" records, so unhandled task failures surface too.
+            LoggingIntegration(level=logging.INFO, event_level=logging.ERROR),
+        ],
```
lebaudantoine, line 64: can't you just pass `enable_logs`? Do you need this explicit `LoggingIntegration`?

src/agents/observability.py
```diff
@@ -0,0 +1,96 @@
...
+
+BASE_DIR = path.dirname(path.abspath(__file__))
+
+# PIDs in which Sentry has already been initialized. LiveKit runs each agent
+# job in its own process. A forked child inherits the parent's initialized
+# Sentry client but not its background thread (threads do not survive fork),
+# so events would be queued and never sent. We therefore key the the current
+# PID and re-initialize whenever we run in a new process, rather than relying on
+# sentry_sdk.is_initialized() (which the child inherits as True).
+_initialized_pids: set[int] = set()
```
lebaudantoine: It feels a bit too bulky to manage PIDs this way. Also, won’t this set keep growing indefinitely as new processes start, without ever removing old IDs?

src/agents/observability.py
```diff
@@ -0,0 +1,96 @@
...
+    if not sentry_dsn:
+        logger.warning("SENTRY_DSN not defined for agent '%s'", agent_name)
+        return
+
+    pid = os.getpid()
+    if pid in _initialized_pids:
+        logger.info(
+            "Sentry already initialized for agent '%s' (pid %d)", agent_name, pid
+        )
+        return
```
lebaudantoine: Is it harmful if the init is called twice for the same process?

src/agents/multi_user_transcriber.py
```diff
@@ -139,6 +147,7 @@ async def _start_session(self, participant: rtc.RemoteParticipant) -> AgentSessi
                 participant_identity=participant.identity,
             )
         )
+        self._sessions[participant.identity] = session
```
lebaudantoine, line 150: was it a bug?

src/agents/observability.py
cameledev: This corresponds to the default, but makes it explicit.

src/agents/observability.py
cameledev: If i am not mistaken, there should be a max of 2 pids, as all agents are processes started from main. 

I will check to confirm.

src/agents/observability.py
cameledev: tbh not sure how this would behave if we call `sentry_sdk.init` twice

src/agents/multi_user_transcriber.py
cameledev: not that I know of. It was moved from on_task_done

src/agents/multi_user_transcriber.py
lebaudantoine: re-reading it, I don’t think this is exactly the same, `start_session` and `on_participant_connected` aren’t triggered at the same moment. I’d advocate for avoiding a silent change in logic here and, if feasible, keeping the session-saving step in the same place in the code.

cameledev: I don't think the logic/behavior here is changed: 

- `_start_session` is called inside of `on_participant_connected` in both cases. 
- The former code:  `self._sessions[participant.identity] = task.result()` runs at the same moment, just in a different function
- I think it makes more sense that `_start_session` actually adds the session object to the participant.

cameledev:
NB: removing PID guard, which was an unnecessary complexity

lebaudantoine (approved):
LGTM

PR #1463, opened 2026-06-30
🛂(backend) reject user access tokens on the API

lebaudantoine:
The original implementation, introduced two years ago, was incorrect and exposed the API to an undesired authentication mode: any user access token obtained for a given user was being accepted as valid credentials on the external API.

Restrict authentication to the intended mode so that user access tokens are no longer accepted on this API.

Thanks @lunika spotting this.


lebaudantoine:
@greptileai

lunika (approved):

PR #1464, opened 2026-06-30
✨(backend) Add per recording encoding config

cameledev:
## Purpose

https://github.com/suitenumerique/meet/issues/1344

NB: Work based on https://github.com/suitenumerique/meet/pull/1357 by @sarthakbahal

## Proposal

Changes on the original PR:

- encoding.py goes into settings
- profile settings are validated
- profile encoding serializers are validated and no longer use hardcoded variables
- profiles are video only and do not apply to audio
- persist query and resolved encoding settings: 
```
recording.options["encoding"] == {
        "resolution": "720p",
        "profile": "talking_heads",
        "resolved": {
            "key_frame_interval": settings.RECORDING_ENCODING_KEY_FRAME_INTERVAL_S,
            "width": 1280,
            "height": 720,
            "framerate": 15,
            "video_bitrate": 700,
        },
    }
    ```

### DIFF
--- a/src/backend/core/api/serializers.py
+++ b/src/backend/core/api/serializers.py
@@ -13,7 +13,12 @@ from django.core.exceptions import SuspiciousOperation
 from django.utils.translation import gettext_lazy as _
 
 from django_pydantic_field.rest_framework import SchemaField
-from pydantic import BaseModel, Field, field_serializer
+from pydantic import (
+    BaseModel,
+    Field,
+    field_serializer,
+    field_validator,
+)
 from pydantic import ValidationError as PydanticValidationError
 from rest_framework import serializers
 from rest_framework.exceptions import PermissionDenied
@@ -227,6 +232,49 @@ class BaseValidationOnlySerializer(serializers.Serializer):
         raise NotImplementedError(f"{self.__class__.__name__} is validation-only")
 
 
+class EncodingConfig(BaseModel):
+    """Configuration options for recording encoding.
+
+    The allowed `resolution` and `profile` values are derived at validation time
+    from ``settings.RECORDING_ENCODING_AVAILABLE_RESOLUTIONS`` and
+    ``settings.RECORDING_ENCODING_AVAILABLE_PROFILES``, so adding a resolution or profile
+    to those maps is enough to make it accepted here.
+
+    Attributes:
+        resolution: Target video resolution.
+        profile: Encoding profile to balance quality and CPU usage. When `None`,
+        LiveKit default framerate/bitrate are used for the resolution.
+    """
+
+    resolution: str
+    profile: str | None = None
+    model_config = {"extra": "forbid"}
+
+    @field_validator("resolution")
+    @classmethod
+    def _validate_resolution(cls, value):
+        """Reject resolutions absent from RECORDING_ENCODING_AVAILABLE_RESOLUTIONS."""
+        allowed = set(settings.RECORDING_ENCODING_AVAILABLE_RESOLUTIONS)
+        if value not in allowed:
+            raise ValueError(
+                f"Invalid resolution '{value}'. Choose from {sorted(allowed)}."
+            )
+        return value
+
+    @field_validator("profile")
+    @classmethod
+    def _validate_profile(cls, value):
+        """Reject profiles absent from RECORDING_ENCODING_AVAILABLE_PROFILES."""
+        if value is None:
+            return None
+        allowed = set(settings.RECORDING_ENCODING_AVAILABLE_PROFILES)
+        if value not in allowed:
+            raise ValueError(
+                f"Invalid profile '{value}'. Choose from {sorted(allowed)}."
+            )
+        return value
+
+
 class RecordingOptions(BaseModel):
     """Configuration options for recording.
 
@@ -247,7 +295,7 @@ class RecordingOptions(BaseModel):
     transcribe: bool | None = None
     collect_metadata: bool | None = None
     original_mode: Literal["screen_recording", "transcript"] | None = None
-
+    encoding: EncodingConfig | None = None
     model_config = {"extra": "forbid"}
 
 
--- a/src/backend/core/api/serializers.py
+++ b/src/backend/core/api/serializers.py
@@ -13,7 +13,12 @@ from django.core.exceptions import SuspiciousOperation
 from django.utils.translation import gettext_lazy as _
 
 from django_pydantic_field.rest_framework import SchemaField
-from pydantic import BaseModel, Field, field_serializer
+from pydantic import (
+    BaseModel,
+    Field,
+    field_serializer,
+    field_validator,
+)
 from pydantic import ValidationError as PydanticValidationError
 from rest_framework import serializers
 from rest_framework.exceptions import PermissionDenied
@@ -244,6 +249,49 @@ class BaseValidationOnlySerializer(serializers.Serializer):
         raise NotImplementedError(f"{self.__class__.__name__} is validation-only")
 
 
+class EncodingConfig(BaseModel):
+    """Configuration options for recording encoding.
+
+    The allowed `resolution` and `profile` values are derived at validation time
+    from ``settings.RECORDING_ENCODING_AVAILABLE_RESOLUTIONS`` and
+    ``settings.RECORDING_ENCODING_AVAILABLE_PROFILES``, so adding a resolution or profile
+    to those maps is enough to make it accepted here.
+
+    Attributes:
+        resolution: Target video resolution.
+        profile: Encoding profile to balance quality and CPU usage. When `None`,
+        LiveKit default framerate/bitrate are used for the resolution.
+    """
+
+    resolution: str
+    profile: str | None = None
+    model_config = {"extra": "forbid"}
+
+    @field_validator("resolution")
+    @classmethod
+    def _validate_resolution(cls, value):
+        """Reject resolutions absent from RECORDING_ENCODING_AVAILABLE_RESOLUTIONS."""
+        allowed = set(settings.RECORDING_ENCODING_AVAILABLE_RESOLUTIONS)
+        if value not in allowed:
+            raise ValueError(
+                f"Invalid resolution '{value}'. Choose from {sorted(allowed)}."
+            )
+        return value
+
+    @field_validator("profile")
+    @classmethod
+    def _validate_profile(cls, value):
+        """Reject profiles absent from RECORDING_ENCODING_AVAILABLE_PROFILES."""
+        if value is None:
+            return None
+        allowed = set(settings.RECORDING_ENCODING_AVAILABLE_PROFILES)
+        if value not in allowed:
+            raise ValueError(
+                f"Invalid profile '{value}'. Choose from {sorted(allowed)}."
+            )
+        return value
+
+
 class RecordingOptions(BaseModel):
     """Configuration options for recording.
 
@@ -264,7 +312,7 @@ class RecordingOptions(BaseModel):
     transcribe: bool | None = None
     collect_metadata: bool | None = None
     original_mode: Literal["screen_recording", "transcript"] | None = None
-
+    encoding: EncodingConfig | None = None
     model_config = {"extra": "forbid"}
 
 
--- a/src/backend/core/api/viewsets.py
+++ b/src/backend/core/api/viewsets.py
@@ -384,6 +384,20 @@ class RoomViewSet(
         options = serializer.validated_data.get("options")
         room = self.get_object()
 
+        if (
+            options is not None
+            and options.encoding is not None
+            and not settings.RECORDING_CUSTOM_ENCODING_ENABLED
+        ):
+            # Per-recording encoding selection is gated by
+            # RECORDING_CUSTOM_ENCODING_ENABLED. When disabled, recordings use
+            # encoding defined by RECORDING_ENCODING_DEFAULT_RESOLUTION
+            # and RECORDING_ENCODING_DEFAULT_PROFILE.
+            return drf_response.Response(
+                {"detail": "Per-recording encoding selection is disabled."},
+                status=drf_status.HTTP_400_BAD_REQUEST,
+            )
+
         options_data = options.model_dump(exclude_none=True) if options else {}
         if options is not None and options.encoding is not None:
             # Persist the resolved encoding (concrete width/height/framerate/
--- a/src/backend/core/api/viewsets.py
+++ b/src/backend/core/api/viewsets.py
@@ -63,12 +63,12 @@ from core.recording.worker.exceptions import (
     RecordingStopError,
 )
 from core.recording.worker.factories import (
+    build_encoding_options,
     get_worker_service,
 )
 from core.recording.worker.mediator import (
     WorkerServiceMediator,
 )
-from core.recording.worker.services import resolve_encoding_config
 from core.services.invitation import InvitationService
 from core.services.livekit_events import (
     LiveKitEventsService,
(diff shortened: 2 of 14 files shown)

### CODE REVIEW

File: src/backend/meet/settings.py [hunk 1]
```diff
@@ -1090,6 +1118,35 @@ def PARLER_LANGUAGES(self):
...
+                "RECORDING_ENCODING_PROFILE_MAP: "
+                f"{resolutions ^ profile_resolutions}"
+            )
+
+        profile_sets = set(map(frozenset, profiles_by_resolution.values()))
+        if len(profile_sets) > 1:
+            raise ValueError(
+                "Every resolution in RECORDING_ENCODING_PROFILE_MAP must expose "
+                f"the same set of profiles, got: {profiles_by_resolution}"
+            )
```
lebaudantoine, line 1251: dx: highlighting what's missing would be more appropriate in this error message.

src/backend/meet/settings.py
```diff
@@ -749,6 +749,34 @@ class Base(Configuration):
...
+        ("540p", "talking_heads"): (15, 400),
+        ("540p", "text"): (15, 600),
+        ("540p", "mixed"): (20, 900),
+        ("720p", "talking_heads"): (15, 700),
+        ("720p", "text"): (15, 1000),
+        ("720p", "mixed"): (20, 1500),
+        ("1080p", "talking_heads"): (15, 1200),
+        ("1080p", "text"): (15, 1800),
+        ("1080p", "mixed"): (20, 2500),
+    }
```
lebaudantoine: I would go for a simpler setup, with a nest profile settings : 

```
RECORDING_ENCODING_MAP = {
    "540p": {
        "size": (960, 540),
        "profiles": {
            "talking_heads": (15, 400),
            "text": (15, 600),
            "mixed": (20, 900),
        },
    },
    "720p": {
        "size": (1280, 720),
        "profiles": {
            "talking_heads": (15, 700),
            "text": (15, 1000),
            "mixed": (20, 1500),
        },
    },
    "1080p": {
        "size": (1920, 1080),
        "profiles": {
            "talking_heads": (15, 1200),
            "text": (15, 1800),
            "mixed": (20, 2500),
        },
    },
}
```

lebaudantoine: The FPS seems to depend only from the profile and not the resolution.

We could also go with : 

```
RECORDING_ENCODING_RESOLUTIONS = {
    "540p": (960, 540),
    "720p": (1280, 720),
    "1080p": (1920, 1080),
}

RECORDING_ENCODING_PROFILES = {
    # profile: (fps, kbps per resolution)
    "talking_heads": (15, {"540p": 400,  "720p": 700,  "1080p": 1200}),
    "text":          (15, {"540p": 600,  "720p": 1000, "1080p": 1800}),
    "mixed":         (20, {"540p": 900,  "720p": 1500, "1080p": 2500}),
}
```

src/backend/core/api/serializers.py
```diff
@@ -227,6 +227,53 @@ def update(self, instance, validated_data):
...
+    profile: str | None = None
+    model_config = {"extra": "forbid"}
+
+    @field_validator("resolution")
+    @classmethod
+    def _validate_resolution(cls, value):
+        """Reject resolutions absent from RECORDING_ENCODING_PROFILE_MAP."""
+        if value is None:
+            return value
+        allowed = {res for res, _ in settings.RECORDING_ENCODING_PROFILE_MAP}
```
lebaudantoine: why don't you validate it against `RECORDING_ENCODING_RESOLUTION_MAP`?

src/backend/core/recording/worker/services.py
```diff
@@ -99,13 +101,46 @@ def _build_encoding_options(self):
...
+            if resolution
+            else (None, None)
+        )
+
+        if resolution and profile:
+            fps, video_kbps = settings.RECORDING_ENCODING_PROFILE_MAP[
+                (resolution, profile)
+            ]
+        else:
+            fps, video_kbps = None, None
```
lebaudantoine: If the resolution isn’t explicitly passed through the API, should it fall back to the default resolution (720p) when determining the profile values?

src/backend/core/tests/recording/worker/test_encoding_resolver.py
```diff
@@ -0,0 +1,117 @@
...
+
+
+def test_resolve_text_1080p(service):
+    """text profile at 1080p should map to high bitrate."""
+    cfg = make_encoding_config(resolution="1080p", profile="text")
+    result = service._resolve_encoding_options(cfg)
+    assert result.width == 1920
+    assert result.height == 1080
+    assert result.framerate == 15
+    assert result.video_bitrate == 1800
```
lebaudantoine: These are snapshots of configurations; I'm not sure they really add value.
You could make sure the logic of resolution works by mocking the settings with the fixture setting.

src/backend/core/tests/recording/worker/test_encoding_resolver.py
```diff
@@ -0,0 +1,117 @@
...
+    """Build a mock EncodingConfig-like object for resolver tests."""
+
+    class FakeEncodingConfig:
+        """Minimal stand-in for EncodingConfig used by the resolver."""
+
+        def __init__(self, resolution_value, profile_value):
+            self.resolution = resolution_value
+            self.profile = profile_value
+
+    return FakeEncodingConfig(resolution, profile)
```
lebaudantoine: Can't you reuse the Pydantic model?

lebaudantoine (commented):
Great, love it!

Just one small concern: could we also test the API without passing any parameters, to confirm that the default resolution and profile are applied? That would help ensure we’re not introducing any unintended changes

src/backend/meet/settings.py
cameledev: Added this information.

src/backend/meet/settings.py
cameledev: I agree with the second format you suggest

cameledev: Making the change

src/backend/core/api/serializers.py
cameledev: changed data structure.

src/backend/core/tests/recording/worker/test_encoding_resolver.py
cameledev: agreed. forgot to remove this.

src/backend/core/recording/worker/services.py
cameledev: Since profile and resolution are inherently linked, I would actually enforce providing both values. I believe this would also be clearer for future devs.

src/backend/core/recording/worker/factories.py
```diff
@@ -78,7 +78,12 @@ class WorkerService(Protocol):
     def __init__(self, config: WorkerServiceConfig):
         """Initialize the service with the given configuration."""
 
-    def start(self, room_id: str, recording_id: str) -> str:
+    def start(
+        self,
+        room_id: str,
+        recording_id: str,
+        encoding_options: Optional[Dict[str, Any]] = None,
```
cameledev, line 108: This is new to latest commit

src/backend/core/recording/worker/services.py
```diff
@@ -11,6 +13,41 @@
 from .factories import WorkerServiceConfig
 
 
+def resolve_encoding_config(encoding_config):
```
lebaudantoine, line 16: wip add type

src/backend/meet/settings.py
```diff
@@ -751,6 +751,37 @@ class Base(Configuration):
...
+    RECORDING_ENCODING_PROFILES = values.DictValue(
+        {
+            # profile: (fps, kbps per resolution)
+            "talking_heads": (15, {"540p": 400, "720p": 700, "1080p": 1200}),
+            "text": (15, {"540p": 600, "720p": 1000, "1080p": 1800}),
+            "mixed": (20, {"540p": 900, "720p": 1500, "1080p": 2500}),
+        },
+        environ_name="RECORDING_ENCODING_PROFILES",
+        environ_prefix=None,
+    )
```
lebaudantoine: WIP: Nitpicking, but I think the naming doesn't make it clear that these settings represent configuration values for the API's accepted parameters, as opposed to the initial settings, which are simply the default values used when no parameter is provided.

src/backend/core/api/viewsets.py
```diff
@@ -379,12 +380,20 @@ def start_room_recording(self, request, pk=None):  # pylint: disable=unused-argu
         options = serializer.validated_data.get("options")
         room = self.get_object()
 
+        options_data = options.model_dump(exclude_none=True) if options else {}
+        if options is not None and options.encoding is not None:
+            # Persist the resolved encoding (concrete width/height/framerate/
+            # bitrate) alongside the requested resolution/profile for traceability.
+            options_data["encoding"]["resolved"] = resolve_encoding_config(
+                options.encoding
+            )
```
lebaudantoine, line 410: I'm usually a bit conservative about what we persist in the database, but I think your call makes sense. Keeping the value could definitely help when debugging scenarios where it's needed.

src/backend/core/recording/worker/factories.py
lebaudantoine, line 97: As discussed during the call, the Protocol static typing mechanism would not properly work with a dynamic `import_string`, let's address this technical debt in a dedicated PR.
Spotted by @FloChehab in another pr.

src/backend/core/recording/worker/services.py
```diff
@@ -99,13 +136,26 @@ def _build_encoding_options(self):
 
         return livekit_api.EncodingOptions(**opts)
 
+    def _resolve_encoding_options(self, encoding_options):
+        """Build LiveKit EncodingOptions from a resolved per-recording dict, or None.
+
+        ``encoding_options`` is the dict persisted by the API in
+        ``recording.options["encoding"]["resolved"]``. Returns None when empty
```
lebaudantoine: I think this comment simply restates the implementation. I'd probably remove it unless it provides additional context that's not obvious from the code.

src/backend/core/recording/worker/services.py
```diff
@@ -145,8 +198,13 @@ class AudioCompositeEgressService(BaseEgressService):
...
     hrid = "audio-recording-composite-livekit-egress"
 
-    def start(self, room_name, recording_id):
-        """Start the audio composite egress process for a recording."""
+    def start(self, room_name, recording_id, encoding_options=None):
+        """Start the audio composite egress process for a recording.
+
+        ``encoding_options`` is accepted for signature compatibility with the
+        WorkerService protocol but ignored: audio-only egress has no video
+        encoding to configure.
```
lebaudantoine:
```suggestion
        WorkerService protocol but ignored: audio-only egress has no encoding
        to configure.
```

lebaudantoine (approved):
LGTM! I think a deeper refactoring of the `WorkerService` protocol and `BaseEgressService` would be beneficial.

Personally, I would probably introduce a `BaseVideoEgressService` to override the `start` signature and encapsulate all the encoding-related logic. That said, it's more of a design preference than a blocker for this PR.

Thanks for your work!

src/backend/meet/settings.py
cameledev: changed for RECORDING_ENCODING_AVAILABLE_RESOLUTIONS

cameledev:
# Claude (opus 4.8) review

# Review: per-recording encoding config (since `ab707d8`)

## Overall

A well-executed, well-tested change. It replaces the old boolean-gated
`RECORDING_ENCODING_*` flat settings with a **profile × resolution** model: a
default encoding always applied, plus an opt-in per-recording `encoding`
override on the start-recording API. Startup validation, docs, and a `💥`
changelog entry for the breaking settings change are all present. **No
correctness bugs found** — the logic is sound.

## What I verified

- **Resolution-only encoding relies on LiveKit treating `framerate=0` /
  `video_bitrate=0` as "use server default"** — confirmed correct against
  `livekit_egress.proto` (`framerate` → `default 30`, `video_bitrate` →
  `default 4500` for the unset/zero case). The docs' claim that a
  resolution-only request "leaves LiveKit's default framerate/bitrate" holds.
- **New DB-persistence path** (`options["encoding"]["resolved"]`): the LiveKit
  `VideoCodec` / `AudioCodec` enums are plain `int`s, so the resolved dict is
  JSON-serializable, round-trips through the `JSONField`, and reconstructs into
  `EncodingOptions(**dict)` cleanly. Verified in-process.
- **`_resolve_encoding_options(...) or _build_encoding_options()`** in
  `services.py`: protobuf `EncodingOptions` are always truthy (even empty), and
  the resolver returns `None` only for an empty/missing dict — so per-recording
  correctly wins over the default, never the reverse. No falsy-message trap.
- **Env override of the new `DictValue` maps**: `DictValue.to_python` uses
  `ast.literal_eval`, which handles the nested-dict literals fine — so
  `RECORDING_ENCODING_AVAILABLE_PROFILES` etc. remain operator-overridable.
- **Gating** in `viewsets.py`: with `RECORDING_CUSTOM_ENCODING_ENABLED=False`, a
  client-supplied `encoding` is rejected (400) and can't influence output;
  without `encoding`, the default applies in both toggle states. `extra:
  "forbid"` on both `RecordingOptions` and `EncodingConfig` blocks unknown keys.
- **`warnings` import** for the new startup check exists; the
  fall-back-to-preset path (`missing default → warn + return`) matches
  `from_settings`' `if resolution and profile` guard.
- **Tests**: ran the DB-independent suites (`test_encoding_resolver.py`,
  `test_factories.py`) — 29 passed.

## Minor observations (all low severity — no action required)

1. **Encoding accepted for audio-only / transcript recordings.** The API
   validates and persists an `encoding` object regardless of `mode`, but
   `AudioCompositeEgressService.start` ignores it (documented). Harmless and
   consistent with the docs' "silently ignored," but a `transcript` recording
   with `encoding` still gets a `201` and a stored-but-inert `resolved` blob.
   Consider rejecting `encoding` when the mode has no video egress — otherwise
   fine as-is.

2. **`_check_recording_encoding_maps` robustness** (`settings.py`): it indexes
   `profile_config["kbps"]` and (in `build_encoding_options`)
   `resolution_config["width"]` without guarding shape. A malformed
   operator-provided map (via env `ast.literal_eval`) surfaces as a raw
   `KeyError` / `TypeError` rather than the friendly `ValueError` the rest of
   the method raises. Still fails fast at startup, so low impact — just a less
   clear message.

3. **Default now always emits explicit `advanced` EncodingOptions.** Previously
   the default (`RECORDING_ENCODING_ENABLED=False`) sent none and let LiveKit
   apply its implicit preset; now `full` / `720p` is sent on every recording.
   Values are chosen to match `H264_720P_30`, and this is intentional +
   changelog-flagged. Note a couple of fields are now pinned explicitly
   (`audio_frequency=48000`, `key_frame_interval=4.0`) where the bare preset may
   have used protobuf defaults — negligible, but a behavior delta beyond the
   settings rename.

Leobouloc:
NB: added these changes to UPGRADE.md

src/backend/core/recording/worker/services.py
```diff
@@ -99,13 +99,24 @@ def _build_encoding_options(self):
...
+    def _resolve_encoding_options(self, encoding_options):
+        """Build LiveKit EncodingOptions from a resolved per-recording dict, or None.
+
+        ``encoding_options`` is the dict persisted by the API in
+        ``recording.options["encoding"]["resolved"]``.
+        """
+        if not encoding_options:
+            return None
+
+        return livekit_api.EncodingOptions(**encoding_options)
```
lebaudantoine: should we use a data class to declare the expected structure of an encoding options?

cameledev:
Replaced by new PR/branch: https://github.com/suitenumerique/meet/pull/1668

Following faulty rebasing.

PR #1467, opened 2026-07-01
✨(frontend) prioritize screen share in picture-in-picture layout

Ovgodd:
## Purpose

When someone shares their screen in PiP, the shared screen should be the main focus,  large at the bottom, with camera tiles in a small row on top. Without screen share, existing layouts stay as they are.

https://github.com/user-attachments/assets/2c8cf9af-22d1-47bc-8d20-61a4c7406f40

## Proposal

- Add `PipScreenShareLayout`: camera row on top, screen share below
- `PipStage` switches to this layout as soon as a screen share is active
- Extract shared pagination wrapper into `PaginatedStage`
- Remove `gridTracks` , screen share and cameras are handled separately now
-----------
- [x] 2 participants + screen share > cameras on top, screen share large below
- [x] I share my screen > same layout, full-screen warning still works
- [x] 4+ participants + screen share > pagination on cameras only, screen share always visible
- [x] 3+ participants, no screen share > grid unchanged
- [x] 1–2 participants, no screen share > focus layout unchanged
- [x] Stop sharing > back to normal layout, no visual glitch


lebaudantoine:
A few product remarks before I review the code:

1. First, I really love the pagination, thanks a lot for this, it’s great. I would keep the same aspect ratio whenever there is a second page, regardless of the number of videos. As soon as the header row is displayed, all videos should maintain a consistent ratio. Otherwise, the image becomes overly stretched, and it’s difficult to focus on the center of the video or the most important content.

Capture d’écran 2026-07-03 à 19 01 53


2. Second, I didn’t notice this before, but we should probably remove the **Track** and **Pin** buttons that appear in the PiP context, since they currently have no effect. I’m curious about your thoughts on this, but as it stands, it could be misleading for users. Another option would be to keep them, but give them an actual purpose. For example, while the screenshot layout may currently be pinned manually, the **Pin** button could be useful for pinning a track directly within the PiP context.

Capture d’écran 2026-07-03 à 19 02 23

3. Finally, one last remark: when there are multiple screen shares, only the most recent one is displayed to the user. I think we should improve this behavior. A possible solution could be the pinning workflow mentioned above, for example, showing additional screen shares in a smaller view, automatically pinning the newest screen share when it appears, and allowing users to pin whichever screen share they want to focus on.

src/frontend/src/features/pip/components/layout/PipStage.tsx
```diff
@@ -13,9 +14,12 @@ import {
 } from '@livekit/components-core'
 
 /**
- * PipStage picks between two layouts based on track count:
- *   - Focus mode (≤ 2 tracks): one main track + one thumbnail overlay.
- *   - Grid mode  (3+ tracks):  adaptive tiling.
+ * PipStage picks between three layouts:
+ *   - Screen share mode (any screen share active):
+ *     small camera tiles in a row at the top, large screen share below.
```
lebaudantoine, line 19: yet, this is the first one that is shown.

lebaudantoine (approved):
LGTM

src/frontend/src/features/pip/components/layout/PipStage.tsx
Ovgodd: indeed, as talk together it will be handled in another PR

PR #1474, opened 2026-07-04
🩹(backend) fix case-insensitive email deduplication in merge command

lebaudantoine:
The management command that merges users with duplicate emails was comparing emails in a case-insensitive manner, which left some duplicates in the database when their emails only differed by case.


lebaudantoine:
@greptileai

lebaudantoine:
@greptileai

cameledev (approved):
Looks good to me

Issue #528, opened 2025-05-12
Pin for everyone


lebaudantoine:
## Feature Request

**Is your feature request related to a problem or unsupported use case? Please describe.**
Room admins currently have no way to control which participant's video is highlighted for everyone. This limits coordination during meetings, presentations, or events where a single speaker should be the focus.

**Describe the solution you'd like**
Add a "Pin for everyone" action to participant video tiles, available only to room admins. When used, it forces all participants to focus on the selected video tile. This state should persist until unpinned by the admin.

**Discovery, Documentation, Adoption, Migration Strategy**
The UI should display the "Pin for everyone" action in the admin context menu on each video tile. A banner or indicator can show that a global pin is active. Documentation should explain that only admins can use this feature and how it affects participant views.

PR #1476, opened 2026-07-05
✨(frontend) pin multiple screen shares and tiles at once

### LINKED ISSUE #1447: Ability to pin multiple shared screens (and tiles) at once
## Feature Request

When several participants share their screen at the same time, only one tile can be pinned (in focus) at a time. Pinning a new tile replaces the previous one, so it's impossible to keep multiple shared screens visible side by side.

This is limiting in several common scenarios:
- **Comparison / collaboration:** following two shared screens at once during a code review, a design comparison, or a multi-document walkthrough.
- **Accessibility:** keeping a presenter's shared screen **and** a sign-language interpreter's video in focus simultaneously — a key use case for deaf and hard-of-hearing participants.

Allow a participant to pin more than one tile (screen share and/or camera) at the same time:
- The pin/unpin action toggles each tile independently, instead of replacing the current focus.
- Pinned tiles are displayed together in a responsive grid (focus area), while non-pinned tiles move to the carousel.
- Unpinning a tile returns it to the grid without affecting the other pinned tiles.
- A reasonable upper bound should be enforced so the focus grid stays readable (Zoom allows up to 9, Google Meet up to 3 — a small cap such as 3–4 seems sensible to start).

This pin state remains **local** to each participant (no impact on others), consistent with the current pin behavior.

- Keeping the current single-pin behavior and relying on the gallery/grid view to see other screens — but it doesn't let the user choose *which* tiles stay in focus, and shared screens get small among camera tiles.
- A global "pin for everyone" (see #528) — useful, but solves a different problem: it's host-controlled and forces a single shared focus, whereas this request is about each participant choosing several tiles for their own view.

**Prior art**
- **Zoom** — "Multi-Pin": pin up to 9 tiles locally; the global equivalent ("Spotlight") is a separate feature.
- **Google Meet** — pin multiple tiles locally; hosts can also pin up to 3 tiles for everyone, introduced largely for accessibility (presenter + sign-language interpreter).

**Relationship to existing issues**
This is the **local / multi-pin** counterpart of #528 ("Pin for everyone"), which covers the **global** admin-controlled pin. The two are complementary and not duplicates: #528 forces a single focus for all participants, whereas this request lets each participant choose several tiles for their own view.

- The existing "Pin" action in the tile context menu keeps the same entry point; only its behavior changes from single-focus to multi-focus.
- The focus layout needs to handle an arbitrary (capped) number of pinned tiles in a responsive grid.
- No migration needed — current single-pin usage is just the n=1 case of the new behavior.

Yes, I'd like to work on this. Happy to coordinate on the approach before starting the implementation.

**Open questions**
Is there already any ongoing reflection or discussion on this topic? Or even early ideas on how it should be implemented?

Nur-Adnan:
## Purpose

Closes #1447.

When several participants share their screen at the same time, only one tile
could be pinned (in focus) at a time: pinning a new tile replaced the previous
one. This made it impossible to keep multiple shared screens visible side by
side, or to keep a presenter's screen and a sign-language interpreter in focus
together (an accessibility use case raised in the issue).

**Root cause.** Focus is LiveKit's `LayoutContext.pin`, whose state is already a
`TrackReferenceOrPlaceholder[]`. LiveKit's built-in reducer only implements
`set_pin` (replace the array) and `clear_pin` (empty it), and
`VideoConference` only ever read `usePinnedTracks(...)?.[0]`. So the single-pin
limit lived entirely in the reducer and the render, not in the data model.

Pin state stays local to each participant, consistent with the current behavior;
nothing is shared with other participants (this is the local counterpart of the
host-controlled #528).

## Proposal

Let a participant pin several tiles (screen shares and/or cameras) at once:

- The pin/unpin action toggles each tile independently instead of replacing the
  current focus.
- Pinned tiles are displayed together in a responsive grid (focus area) while
  non-pinned tiles move to the carousel.
- A small cap (`MAX_PINNED_TILES = 4`) keeps the focus grid readable; pin
  controls are disabled once it is reached.

- [x] Replace LiveKit's single-tile pin reducer with a track-aware multi-pin
      reducer (`toggle` / `add` / `remove` / `replace`, capped, de-duplicated,
      referentially stable on no-ops) in a new `useMultiPin` module, exposed
      through the layout context so LiveKit consumers (`usePinnedTracks`,
      `isTrackReferencePinned`) keep working unchanged.
- [x] Render the pinned set in the existing responsive `GridLayout`, and exclude
      pinned tiles from the carousel.
- [x] Update every pin entry point (tile pin button, participant menu, join
      toast) to toggle independently; unpinning or unsubscribing one tile no
      longer clears the others.
- [x] Generalize the screen-reader pin/unpin announcements and the screen-share
      auto-focus / placeholder handling to the multi-pin set.

### Design decisions

- **Reuse LiveKit's pin state, swap only the reducer.** Overriding just the
  `pin` slice of the layout context (keeping the `widget` slice) means the pin
  state stays a plain array and every LiveKit read-path keeps working. No fork
  of unrelated LiveKit internals.
- **Conservative auto-focus.** Screen-share auto-focus is unchanged (only the
  first screen share auto-focuses), so single-share meetings behave exactly as
  before; additional tiles are pinned manually.
- **Cap = 4.** Small enough to stay readable (Zoom allows 9, Meet 3); one named
  constant, trivially tunable.
- **Disable pin controls at the cap** rather than silently ignoring the click,
  consistent with how the app already disables controls (e.g. the mute button).

### Files changed

- `hooks/useMultiPin.ts` (new) — reducer + hooks + cap constant.
- `prefabs/VideoConference.tsx` — context wiring, focus grid, carousel
  exclusion, auto-focus + placeholder + announcement effects.
- `components/ParticipantTileFocus.tsx`, `hooks/useFocusToggleParticipant.ts`,
  `components/ParticipantMenu/PinMenuItem.tsx`,
  `components/ParticipantTile.tsx`, `notifications/ToastJoined.tsx` — track-aware
  pin entry points.
- `components/FocusLayout.tsx` — removed (dead after the change).
- `CHANGELOG.md` — one line.

No new dependencies. No i18n changes (reuses the existing pin/unpin strings).

### Testing performed

- `npm run lint` (eslint `--max-warnings 0`), `npm run check` (prettier),
  `npm run build` (`tsc -b` + vite) all pass.
- Manual reasoning + code review over the reducer and the two effects
  (convergence, no dispatch loop, StrictMode double-invoke safety).
- Remaining: a live click-through in a running room (pin/unpin independence, the
  4-tile grid, cap disabling, screen-share auto-focus). Screenshots to be added
  from that session.

### Performance impact

Negligible. The layout-context value is rebuilt per render exactly as the
original `useCreateLayoutContext()` was, so there is no new re-render surface;
the reducer now returns the same array reference on no-op dispatches to avoid
spurious updates. The `Room` bundle chunk is unchanged (slightly smaller).

### Accessibility

Pin/unpin changes are announced to screen readers per tile via the existing
`useScreenReaderAnnounce` channel. Pin controls at the cap use the app's
standard disabled-control pattern. Keyboard operability of the tile controls is
unchanged.

### Security

None. This is local, client-side UI state: no API, auth, RBAC, token, or input
handling is touched.

### Breaking changes / deployment

None. Single-pin usage is simply the `n = 1` case of the new behavior; no
migration, config, or deployment step required.

### Reviewer guidance

Start at `useMultiPin.ts` (the reducer + the layout-context override and the
`as unknown as` cast at the LiveKit boundary), then `VideoConference.tsx` (focus
grid render + the auto-focus/placeholder/announcement effects). The rest are
mechanical `useFocusToggle` -> track-aware toggle swaps.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ and this project adheres to
 
 ## [Unreleased]
 
+- ✨(frontend) pin multiple screen shares and tiles at once #1447
 - 🚀 (front) fix frontend build failure
 
 ## [1.22.0] - 2026-07-03
--- a/src/frontend/src/features/rooms/livekit/components/ParticipantTileFocus.tsx
+++ b/src/frontend/src/features/rooms/livekit/components/ParticipantTileFocus.tsx
@@ -9,15 +9,13 @@ import {
   RiPushpin2Line,
   RiUnpinLine,
 } from '@remixicon/react'
-import {
-  useFocusToggle,
-  useTrackMutedIndicator,
-} from '@livekit/components-react'
+import { useTrackMutedIndicator } from '@livekit/components-react'
 import { useTranslation } from 'react-i18next'
 import { TrackReferenceOrPlaceholder } from '@livekit/components-core'
 import { useEffect, useRef, useState } from 'react'
 import { useSidePanel } from '../hooks/useSidePanel'
 import { useFullScreen } from '../hooks/useFullScreen'
+import { useTogglePin } from '../hooks/useMultiPin'
 import { type Participant, Track } from 'livekit-client'
 import { MuteAlertDialog } from './MuteAlertDialog'
 import { useMuteParticipant } from '@/features/rooms/api/muteParticipant'
@@ -56,20 +54,17 @@ const FocusButton = ({
   trackRef: TrackReferenceOrPlaceholder
 }) => {
   const { t } = useTranslation('rooms', { keyPrefix: 'participantTileFocus' })
-  const { mergedProps, inFocus } = useFocusToggle({
-    trackRef,
-    props: {},
-  })
+  const { toggle, isPinned, canPin } = useTogglePin(trackRef)
   return (
     <Button
       size="sm"
       variant="primaryTextDark"
       square
-      tooltip={inFocus ? t('pin.disable') : t('pin.enable')}
-      // eslint-disable-next-line @typescript-eslint/no-explicit-any
-      onPress={(event) => mergedProps?.onClick?.(event as any)}
+      isDisabled={!canPin}
+      tooltip={isPinned ? t('pin.disable') : t('pin.enable')}
+      onPress={() => toggle()}
     >
-      {inFocus ? <RiUnpinLine /> : <RiPushpin2Line />}
+      {isPinned ? <RiUnpinLine /> : <RiPushpin2Line />}
     </Button>
   )
 }
--- a/src/frontend/src/features/rooms/livekit/hooks/useMultiPin.ts
+++ b/src/frontend/src/features/rooms/livekit/hooks/useMultiPin.ts
@@ -1,0 +1,182 @@
+import {
+  isTrackReference,
+  isTrackReferencePinned,
+  PIN_DEFAULT_STATE,
+} from '@livekit/components-core'
+import type {
+  PinState,
+  TrackReferenceOrPlaceholder,
+} from '@livekit/components-core'
+import {
+  useCreateLayoutContext,
+  useMaybeLayoutContext,
+} from '@livekit/components-react'
+import type { LayoutContextType } from '@livekit/components-react'
+import { useCallback, useReducer, type Dispatch } from 'react'
+
+/**
+ * Maximum number of tiles a single participant can pin (keep in focus) at once.
+ *
+ * The pin state stays local to each participant, so this only bounds one user's
+ * own focus grid. It is kept intentionally small so the grid stays readable when
+ * several screen shares and cameras are pinned together (Zoom allows 9, Google
+ * Meet 3 — 4 is a comfortable middle ground for side-by-side screen shares plus
+ * an accessibility tile such as a sign-language interpreter).
+ */
+export const MAX_PINNED_TILES = 4
+
+/**
+ * Pin actions that operate on individual tiles, superseding LiveKit's built-in
+ * `set_pin`/`clear_pin` reducer (which only ever holds a single tile). Every
+ * action carries the track it targets so toggling one tile never affects the
+ * others.
+ */
+export type MultiPinAction =
+  | { msg: 'toggle_pin'; trackReference: TrackReferenceOrPlaceholder }
+  | { msg: 'add_pin'; trackReference: TrackReferenceOrPlaceholder }
+  | { msg: 'remove_pin'; trackReference: TrackReferenceOrPlaceholder }
+  /**
+   * Replace a pinned placeholder with its now-published track reference, in
+   * place, without changing the pin order or count.
+   */
+  | { msg: 'replace_pin'; trackReference: TrackReferenceOrPlaceholder }
+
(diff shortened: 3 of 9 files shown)


src/frontend/src/features/rooms/livekit/prefabs/VideoConference.tsx
```diff
@@ -110,89 +102,68 @@ export function VideoConference({ ...props }: VideoConferenceProps) {
...
     ) {
       log.debug('Auto set screen share focus:', {
         newScreenShareTrack: screenShareTracks[0],
       })
-      layoutContext.pin.dispatch?.({
-        msg: 'set_pin',
-        trackReference: screenShareTracks[0],
-      })
+      dispatchPin({ msg: 'add_pin', trackReference: screenShareTracks[0] })
       lastAutoFocusedScreenShareTrack.current = screenShareTracks[0]
```
davd-gzl, line 167: A screen share is meant to come to the front on its own. With four tiles pinned there is no slot, so the first line does nothing, and the second records it as done anyway. That record is what stops it ever trying again: freeing a slot does not bring the share forward, only stopping and restarting it does.

src/frontend/src/features/rooms/livekit/components/ParticipantTileFocus.tsx
```diff
@@ -56,20 +54,17 @@ const FocusButton = ({
...
   return (
     <Button
       size="sm"
       variant="primaryTextDark"
       square
-      tooltip={inFocus ? t('pin.disable') : t('pin.enable')}
-      // eslint-disable-next-line @typescript-eslint/no-explicit-any
-      onPress={(event) => mergedProps?.onClick?.(event as any)}
+      isDisabled={!canPin}
+      tooltip={isPinned ? t('pin.disable') : t('pin.enable')}
```
davd-gzl, line 64: Nit: at the cap the pin greys out and hovering it explains nothing on why it is disabled.

src/frontend/src/features/rooms/livekit/hooks/useMultiPin.ts
```diff
@@ -0,0 +1,182 @@
...
+ * because the pin state remains a plain `TrackReferenceOrPlaceholder[]`.
+ */
+export function useCreateMultiPinLayoutContext(): MultiPinLayoutContext {
+  const base = useCreateLayoutContext()
+  const [pinnedTracks, dispatchPin] = useReducer(
+    multiPinReducer,
+    PIN_DEFAULT_STATE
+  )
+
+  const layoutContext: LayoutContextType = {
```
davd-gzl: Suggestion: this object is built fresh on every render, so every video square in the meeting re-renders with it; a `useMemo` on `pinnedTracks` and `base.widget` holds it steady.

CHANGELOG.md
```diff
@@ -8,6 +8,7 @@ and this project adheres to
 
 ## [Unreleased]
 
+- ✨(frontend) pin multiple screen shares and tiles at once #1447
```
davd-gzl, line 11: Nit: this repo appends an entry at the bottom of its section, dates ascending.

PR #1477, opened 2026-07-06
🐛(makefile) fix passing args to tests in makefile

cameledev:
When using Makefile to launch tests, passing flags as well as specific classes (using "::") was broken. This PR fixes this issue, by adding an `ARGS` argument allowing  to do `make test ARGS="core/tests/foo.py::Test::x -vv"`.


lebaudantoine (approved):
LGTM but I have not tested it locally

PR #1479, opened 2026-07-06
Feat/summary feature flag in visio

FloChehab:
Commits are nicely split :)

I am leaving the `summary-tracing-consent` flag in the Summary project for now. From my PoV it makes sense for this one to be centralized to the scope of summary anyway.

### DIFF
--- a/src/backend/core/analytics/base.py
+++ b/src/backend/core/analytics/base.py
@@ -1,10 +1,11 @@
 """Analytics backend protocol and default no-op implementation."""
 
 from abc import ABC, abstractmethod
-from typing import Any
+from typing import Any, Mapping
 
 from ..models import User
 from .events import AnalyticsEvent
+from .user_feature_flags import UserFeatureFlag
 
 
 class AnalyticsBackend(ABC):
@@ -34,6 +35,21 @@ class AnalyticsBackend(ABC):
     @abstractmethod
     def shutdown(self) -> None:
         """Flush pending events. Called on process exit."""
+
+    def get_user_feature_flags(
+        self,
+        user: User,  # pylint: disable=unused-argument
+    ) -> Mapping[UserFeatureFlag, bool | str | None]:
+        """Return a dict of feature flags for the given user."""
+        # We return an empty dict here by default to avoid a breaking change
+        # By making this method abstract.
+        return {}
+
+    def is_user_feature_enabled(
+        self, user: User, feature_name: UserFeatureFlag
+    ) -> bool:
+        """Check if a feature is enabled at the user level."""
+        return self.get_user_feature_flags(user).get(feature_name, False) is True
 
 
 class NoOpAnalytics(AnalyticsBackend):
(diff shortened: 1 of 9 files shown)


src/backend/core/analytics/base.py
```diff
@@ -33,8 +34,21 @@ def capture(
...
+        """Return a dict of feature flags for the given user."""
+        # We return an empty dict here by default to avoid a breaking change
+        return {}
 
-class NoOpAnalytics:
+    def is_user_feature_enabled(
+        self, user: User, feature_name: UserFeatureFlag
+    ) -> bool:
+        """Check if a feature is enabled at the user level."""
+        return self.get_user_feature_flags(user).get(feature_name, False) is True
```
lebaudantoine, line 52: protocol are there to describe an interface, not an implementation

src/backend/core/analytics/posthog.py
```diff
@@ -72,3 +80,37 @@ def capture(
     def shutdown(self) -> None:
         """Flush pending events. Called on process exit."""
         self._client.shutdown()
+
+    def _compute_user_feature_flags(
```
cameledev:
```suggestion
    def _fetch_user_feature_flags(
```
?

cameledev (commented):
Code looks good to me !

(Not tested)

src/backend/core/analytics/base.py
FloChehab: Talked IRL we are moving to abstract class.

src/backend/core/analytics/posthog.py
FloChehab: Changed :)

src/backend/core/analytics/posthog.py
```diff
@@ -20,6 +23,8 @@ def __init__(
         *,
         api_key: str,
         host: str = "https://eu.i.posthog.com",
+        feature_flags_cache_ttl: int = 60,
+        feature_flags_cache_prefix: str = "user_feature_flags:",
```
lebaudantoine, line 27: It is not easily updatable without releasing a new version of the software. At this time, I don't have a better alternative to propose.

src/backend/core/analytics/posthog.py
```diff
@@ -73,3 +80,37 @@ def capture(
...
+        distinct_id = self._distinct_id(user)
+        if distinct_id is None:
+            return {}
+
+        try:
+            return cache.get_or_set(
+                f"{self._feature_flags_cache_prefix}{distinct_id}",
+                default=lambda: self._fetch_user_feature_flags(user),
+                timeout=self._feature_flags_cache_ttl,
+            )
```
lebaudantoine, line 113: The cache is intended to reduce network requests, as I understand it. Caching an empty dictionary when distinct_id is None doesn't seem particularly useful. I'd be cautious about what we cache and try to keep it to the minimum amount of data necessary.

lebaudantoine (approved):
Minor comments overall, LGTM, I'ven't tested it locally.

As we discussed in PM, I believe PostHog already provides built-in local evaluation and caching for feature flags through its SDK.

I haven't looked into the details to determine whether it fully applies to our use case, but from a quick read, it seems like it could, especially if we start proxying feature flag requests through the backend.

This might be worth considering as a future refactoring.

PR #1480, opened 2026-07-07
♻️(backend) refactor analytics backend from Protocol to abstract class

lebaudantoine:
Switch the analytics backend interface from a typing Protocol to an abstract base class, so the contract that backends must implement is explicit and enforced at instantiation time.


FloChehab (approved):
Approved, might be interesting to raise a NotImplementedError in the abstract method.

PR #1498, opened 2026-07-10
✨(frontend) add screen share zoom controls

Ovgodd:
## Purpose

This PR adds zoom and pan controls for remote screen shares, with keyboard support and screen reader announcements for accessibility compliance (RGAA/WCAG).

https://github.com/user-attachments/assets/2b93a420-5234-46d5-8633-bdb337404ce7

## Proposal

Add a bottom-right floating toolbar on remote screen share tiles with zoom in/out, fit-to-window reset, and fullscreen. 

- [x] Zoom in/out on remote screen shares (100% to 400%, 10% steps)
- [x] Pan with mouse drag when zoomed
- [x] Pan with arrow keys when zoomed
- [x] Ctrl/Cmd + scroll wheel zoom
- [x] Bottom-right toolbar (zoom in, zoom out, percentage, fit-to-window, fullscreen)
- [x] Fullscreen targets the tile container so controls stay visible
- [x] Screen reader supported
- [x] i18n: EN, FR, NL, DE


src/frontend/src/features/rooms/livekit/components/ScreenShareZoomControls.tsx
```diff
@@ -0,0 +1,175 @@
...
+  canZoomOut,
+  onZoomIn,
+  onZoomOut,
+  onResetZoom,
+}: ScreenShareZoomControlsProps) => {
+  const { t } = useTranslation('rooms', { keyPrefix: 'screenShareZoom' })
+  const announce = useScreenReaderAnnounce()
+
+  const [isFullscreen, setIsFullscreen] = useState(false)
+  const wasOwnFullscreen = useRef(false)
```
lebaudantoine: nit: I'm not sure to understand what "own" designates here.

src/frontend/src/features/rooms/livekit/components/ScreenShareZoomableVideo.tsx
```diff
@@ -0,0 +1,108 @@
...
+            transform: `scale(${zoom.zoomLevel}) translate(${zoom.panOffset.x}%, ${zoom.panOffset.y}%)`,
+            transformOrigin: 'center center',
+            transition: zoom.isDragging ? 'none' : 'transform 150ms ease-out',
+          }}
+        >
+          <VideoTrack
+            trackRef={trackRef}
+            onSubscriptionStatusChanged={onSubscriptionStatusChanged}
+            manageSubscription={manageSubscription}
+          />
```
lebaudantoine: Could we try to memoize this component to avoid it to re-render on every zoom change ? wdyt?

src/frontend/src/features/rooms/livekit/hooks/useScreenShareZoom.ts
```diff
@@ -0,0 +1,227 @@
...
+      const delta = -e.deltaY * WHEEL_ZOOM_SPEED
+      setZoomLevel((prev) => {
+        const next = clampZoom(prev + delta)
+
+        if (next <= MIN_ZOOM) {
+          setPanOffset({ x: 0, y: 0 })
+          return MIN_ZOOM
+        }
+
+        const rect = e.currentTarget.getBoundingClientRect()
```
lebaudantoine: updaters don't execute synchronously inside the event handler, currentTarget is null in my test

src/frontend/src/features/rooms/livekit/components/ScreenShareZoomableVideo.tsx
```diff
@@ -0,0 +1,108 @@
...
+        })}
+        style={{
+          cursor: panCursor,
+        }}
+        onWheel={zoom.handleWheel}
+        onMouseDown={zoom.handlePanStart}
+        onMouseMove={zoom.handlePanMove}
+        onMouseUp={zoom.handlePanEnd}
+        onMouseLeave={zoom.handlePanEnd}
+      >
```
lebaudantoine, line 77: have you checked https://react-aria.adobe.com/useMove?

lebaudantoine: I guess we could drop the onMouseDown/Move/Up/Leave

src/frontend/src/features/rooms/livekit/hooks/useScreenShareZoom.ts
```diff
@@ -0,0 +1,227 @@
...
+const WHEEL_ZOOM_FOCUS_BLEND = 0.3
+const PAN_STEP = 5
+const PAN_CLAMP_HALF = 50
+
+interface PanOffset {
+  x: number
+  y: number
+}
+
+export function useScreenShareZoom() {
```
lebaudantoine: using the useMove hook would simplify the math here

src/frontend/src/features/rooms/livekit/hooks/useScreenShareZoom.ts
```diff
@@ -0,0 +1,227 @@
...
+          const zoomRatio = next / prev
+          return clampPan(
+            {
+              x:
+                pan.x +
+                (cursorXPercent - pan.x) *
+                  (1 - 1 / zoomRatio) *
+                  WHEEL_ZOOM_FOCUS_BLEND,
+              y:
+                pan.y +
```
lebaudantoine: Updater must stay pure: nesting setPanOffset inside makes it impure, and StrictMode's double-invoke then applies the relative pan nudge twice (drifts in dev only). Compute zoom + pan in one transition instead.

src/frontend/src/features/rooms/livekit/components/ScreenShareZoomableVideo.tsx
```diff
@@ -0,0 +1,108 @@
...
+        <div
+          style={{
+            width: '100%',
+            height: '100%',
+            pointerEvents: 'none',
+            transform: `scale(${zoom.zoomLevel}) translate(${zoom.panOffset.x}%, ${zoom.panOffset.y}%)`,
+            transformOrigin: 'center center',
+            transition: zoom.isDragging ? 'none' : 'transform 150ms ease-out',
+          }}
```
lebaudantoine, line 85: can we use an imperative approach to exclude react from the loop. Using a ref, and calling transform directly from the dom to avoid triggering a react re-rendering everytime panOffset changes

lebaudantoine (commented):
mostly performances concerns

src/frontend/src/features/rooms/livekit/components/ScreenShareZoomControls.tsx
Ovgodd: it was for "identified this tile among multiple instances" but yes I will change the name

src/frontend/src/features/rooms/livekit/components/ScreenShareZoomableVideo.tsx
Ovgodd: yes of course

src/frontend/src/features/rooms/livekit/components/ScreenShareZoomableVideo.tsx
Ovgodd: indeed :)

src/frontend/src/features/rooms/livekit/utils/screenShareZoom.ts
```diff
@@ -0,0 +1,115 @@
...
+  return `scale(${zoom}) translate(${pan.x}%, ${pan.y}%)`
+}
+
+export const getCursorFromZoomState = (zoom: number, dragging: boolean) => {
+  if (zoom <= MIN_ZOOM) return 'default'
+  return dragging ? 'grabbing' : 'grab'
+}
+
+// Blend the current pan toward the cursor position so the zoom "tracks" the pointer.
+export const getWheelPanOffset = ({
```
davd-gzl, line 103: Critical: ctrl-scroll should zoom where the mouse is instead of the middle of the shared screen.

the spot under the mouse leaving the screen

src/frontend/src/features/rooms/livekit/utils/screenShareZoom.ts
```diff
@@ -0,0 +1,115 @@
...
+    nextZoom
+  )
+}
+
+// Convert cursor pixel position to a % offset from the surface center.
+export const getCursorPercentsFromWheelEvent = (
+  e: WheelEvent,
+  target: HTMLElement
+) => {
+  const rect = target.getBoundingClientRect()
```
davd-gzl, line 134: Critical: the mouse is measured against the whole video box, empty bars included, and not against the picture inside it: at the top of the picture that reads 9.9% down instead of 0.3%.

the two boxes and the two readings

src/frontend/src/features/rooms/livekit/components/ScreenShareZoomControls.tsx
```diff
@@ -0,0 +1,203 @@
...
+  const announce = useScreenReaderAnnounce()
+
+  const [isFullscreen, setIsFullscreen] = useState(false)
+  // Tracks whether this tile's container triggered fullscreen (vs another share's).
+  const wasThisTileFullscreen = useRef(false)
+  const [isFullscreenAvailable] = useState(
+    () => typeof document !== 'undefined' && document.fullscreenEnabled
+  )
+
+  // Covers Esc and browser UI exits, not just the toolbar button.
```
davd-gzl, line 49: Escape should leave this fullscreen and does not. Related: [#1507](https://github.com/suitenumerique/meet/pull/1507) binds Escape to close the side panel.

src/frontend/src/features/rooms/livekit/components/ScreenShareZoomableVideo.tsx
```diff
@@ -0,0 +1,104 @@
...
+    <>
+      <div
+        ref={zoom.surfaceElRef}
+        className={css({
+          width: '100%',
+          height: '100%',
+          overflow: 'hidden',
+          position: 'relative',
+          userSelect: 'none',
+          touchAction: 'none',
```
davd-gzl: A phone cannot pinch to zoom: this switches off the browser's own pinch and nothing replaces it.

a pinch on a phone viewport doing nothing

src/frontend/src/features/rooms/livekit/hooks/useScreenShareZoom.ts
```diff
@@ -0,0 +1,260 @@
...
+  const flush = useCallback(() => {
+    snapshotRef.current = buildZoomSnapshot(
+      zoomRef.current,
+      panRef.current,
+      draggingRef.current
+    )
+    listenersRef.current.forEach((cb) => cb())
+  }, [])
+
+  const applyTransform = useCallback((transition: boolean) => {
```
davd-gzl: The buttons and the keys animate the zoom over 150ms and the wheel does not. I think instant on every path is better.

src/frontend/src/features/rooms/livekit/hooks/useScreenShareZoom.ts
```diff
@@ -0,0 +1,260 @@
...
+    applyTransform(true)
+    applyCursor()
+    flush()
+  }, [applyTransform, applyCursor, flush])
+
+  // Must be attached with { passive: false } so preventDefault() blocks
+  // the browser's native Ctrl+scroll page zoom.
+  const handleWheel = useCallback(
+    (e: WheelEvent) => {
+      if (!e.ctrlKey && !e.metaKey) return
```
davd-gzl, line 136: On a trackpad, two-finger scroll does nothing over a zoomed share: only ctrl or cmd with the wheel reaches this handler.

an unmodified wheel over a zoomed share

src/frontend/src/features/rooms/livekit/hooks/useScreenShareZoom.ts
```diff
@@ -0,0 +1,260 @@
...
+    flush()
+  }, [applyTransform, applyCursor, flush])
+
+  const resetZoom = useCallback(() => {
+    zoomRef.current = MIN_ZOOM
+    panRef.current = { x: 0, y: 0 }
+    applyTransform(true)
+    applyCursor()
+    flush()
+  }, [applyTransform, applyCursor, flush])
```
davd-gzl, line 130: Refactor: folding these onto one setter also settles the asymmetry where `zoomOut` recentres the pan at 1.0x and `zoomIn` does not.

```suggestion
  const setZoom = useCallback(
    (next: number) => {
      zoomRef.current = next
      panRef.current =
        next <= MIN_ZOOM ? { x: 0, y: 0 } : clampPan(panRef.current, next)
      applyTransform(true)
      applyCursor()
      flush()
    },
    [applyTransform, applyCursor, flush]
  )

  const zoomIn = useCallback(
    () => setZoom(clampZoom(zoomRef.current + ZOOM_STEP)),
    [setZoom]
  )

  const zoomOut = useCallback(
    () => setZoom(clampZoom(zoomRef.current - ZOOM_STEP)),
    [setZoom]
  )

  const resetZoom = useCallback(() => setZoom(MIN_ZOOM), [setZoom])
```

src/frontend/src/features/rooms/livekit/components/ScreenShareZoomControls.tsx
```diff
@@ -0,0 +1,203 @@
...
+}: ScreenShareZoomControlsProps) => {
+  const { t } = useTranslation('rooms', { keyPrefix: 'screenShareZoom' })
+  const announce = useScreenReaderAnnounce()
+
+  const [isFullscreen, setIsFullscreen] = useState(false)
+  // Tracks whether this tile's container triggered fullscreen (vs another share's).
+  const wasThisTileFullscreen = useRef(false)
+  const [isFullscreenAvailable] = useState(
+    () => typeof document !== 'undefined' && document.fullscreenEnabled
+  )
```
davd-gzl: Refactor: the setter is discarded, nothing writes this state again, and the frontend has no server rendering.

```suggestion
  const isFullscreenAvailable = document.fullscreenEnabled
```

src/frontend/src/features/rooms/livekit/components/ScreenShareZoomableVideo.tsx
```diff
@@ -0,0 +1,104 @@
...
+    }
+
+    if (!zoom.isZoomed) hasAnnouncedPanHint.current = false
+  }, [zoom.zoomPercentage, zoom.isZoomed, announce, t])
+
+  // Attach keyboard listener on the tile container (has tabIndex=0).
+  useEffect(() => {
+    const el = tileRef.current
+    if (!el) return
+    el.addEventListener('keydown', zoom.handleKeyDown)
```
davd-gzl, line 50: These keys are bound here instead of in [`shortcutCatalog`](https://github.com/suitenumerique/meet/blob/feat/screen-share-zoom/src/frontend/src/features/shortcuts/catalog.ts#L32), which is the list [`ShortcutTab`](https://github.com/suitenumerique/meet/blob/feat/screen-share-zoom/src/frontend/src/features/settings/components/tabs/ShortcutTab.tsx#L48) shows the user, so plus, minus, zero and the arrows appear in no shortcut list.

src/frontend/src/features/rooms/livekit/components/controls/Options/FullScreenMenuItem.tsx
```diff
@@ -20,12 +20,12 @@ export const FullScreenMenuItem = () => {
...
       {isCurrentlyFullscreen ? (
         <>
-          <RiFullscreenExitLine size={20} />
+          <RiCollapseDiagonalLine size={20} />
           {t('fullscreen.exit')}
         </>
       ) : (
         <>
-          <RiFullscreenLine size={20} />
+          <RiExpandDiagonalLine size={20} />
```
davd-gzl: Question: the room's own fullscreen menu item changes icon here, which is not part of the zoom controls. Is that intended?

corner brackets against diagonal arrows

src/frontend/src/features/rooms/livekit/utils/screenShareZoom.ts
```diff
@@ -0,0 +1,115 @@
...
+  return {
+    cursorXPercent:
+      ((e.clientX - rect.left) / rect.width) * 100 - PAN_CLAMP_HALF,
```
davd-gzl: Nit: `PAN_CLAMP_HALF` is the middle of the picture here and the pan limit at [line 30](https://github.com/suitenumerique/meet/blob/feat/screen-share-zoom/src/frontend/src/features/rooms/livekit/utils/screenShareZoom.ts#L30), so tuning the limit would silently move the middle.

src/frontend/src/features/rooms/livekit/utils/screenShareZoom.ts
```diff
@@ -0,0 +1,115 @@
...
+  isDragging: boolean
+  canZoomIn: boolean
+  canZoomOut: boolean
+}
+
+export const clampZoom = (value: number) => {
+  return Math.max(MIN_ZOOM, Math.min(MAX_ZOOM, value))
+}
+
+// Restrict pan so the content edge never passes the viewport center.
```
davd-gzl: Nit: dragging stops when you run out of picture, not when half the view is empty as this comment says.

```suggestion
// Restrict pan so the picture always covers the view.
```

the picture stopping while the drag continues

davd-gzl (changes requested):

src/frontend/src/features/rooms/livekit/components/controls/Options/FullScreenMenuItem.tsx
lebaudantoine: Agree, I would keep the existing icons (RiFullscreenLine) for the room fullscreen option, and use the RiExpand/CollapseDiagonalLine for the tile zoom.

src/frontend/src/features/rooms/livekit/utils/screenShareZoom.ts
Ovgodd: https://github.com/user-attachments/assets/25a0bf64-091a-4f4f-b7fa-b1eb9d68568b

src/frontend/src/features/rooms/livekit/components/ScreenShareZoomControls.tsx
Ovgodd: Escape leave fullscreen on my side 

https://github.com/user-attachments/assets/88e8f29f-b81f-45fa-8a50-f8b8b746a915

src/frontend/src/features/rooms/livekit/components/ScreenShareZoomControls.tsx
Ovgodd: Exact ! nice catch

src/frontend/src/features/rooms/livekit/components/ScreenShareZoomControls.tsx
davd-gzl: I just retried and me too, I remember testing that one that's weird.
Thank you for the video! :slightly_smiling_face:

Ovgodd: Hey ! My pleasure :) !

src/frontend/src/features/rooms/livekit/utils/screenShareZoom.ts
Ovgodd: nice catch  thanks !

src/frontend/src/features/rooms/livekit/components/ScreenShareZoomControls.tsx
```diff
@@ -0,0 +1,223 @@
...
+  const { t } = useTranslation('rooms', { keyPrefix: 'screenShareZoom' })
+  const announce = useScreenReaderAnnounce()
+
+  const zoomInButtonRef = useRef<HTMLButtonElement>(null)
+  const hadFocusInCollapsibleRef = useRef(false)
+
+  const [isFullscreen, setIsFullscreen] = useState(false)
+  // Tracks whether this tile's container triggered fullscreen (vs another share's).
+  const wasThisTileFullscreen = useRef(false)
+  const isFullscreenAvailable = document.fullscreenEnabled
```
davd-gzl, line 47: This button is dead inside the picture-in-picture window. It asks the opener whether fullscreen is available, and the tile it would enlarge lives in that other window, where the API is off.

<details><summary>repro</summary>

Share a screen, open picture-in-picture from More options, then in the opener's console:

```js
const pip = window.documentPictureInPicture.window
document.fullscreenEnabled                      // the opener
pip.document.fullscreenEnabled                  // that window
await pip.document.querySelector('[data-lk-source="screen_share"]').requestFullscreen()
```

```console
true
false
TypeError: Fullscreen is not supported
```

</details>

The panel reads both documents live, and the last line is what [`toggleFullScreen`](https://github.com/suitenumerique/meet/blob/feat/screen-share-zoom/src/frontend/src/features/rooms/livekit/components/ScreenShareZoomControls.tsx#L78) catches and logs, three presses in:

the fullscreen button pressed three times inside the picture-in-picture window

src/frontend/src/features/rooms/livekit/components/ScreenShareZoomControls.tsx
```diff
@@ -0,0 +1,223 @@
...
+      } else {
+        // Tile container so zoom controls stay visible in fullscreen.
+        await containerRef.current?.requestFullscreen()
+      }
+    } catch (error) {
+      console.error('Error toggling fullscreen:', error)
+    }
+  }, [containerRef])
+
+  const wheelShortcutVisual = isMacintosh() ? '⌘+scroll' : 'Ctrl+scroll'
```
davd-gzl, line 91: Nit: The zoom tooltips read `Ctrl+scroll` in every language, while [`wheelShortcut`](https://github.com/suitenumerique/meet/blob/feat/screen-share-zoom/src/frontend/src/locales/fr/rooms.json#L765) and `wheelShortcutMac` sit translated in all four locale files and are read nowhere, so French shows this beside a screen-reader hint that does say `Contrôle plus molette`:

the zoom in tooltip in a French interface reading Zoomer (Ctrl+scroll)

davd-gzl (changes requested):
Other then this points it should be good, I will do a last manual pass! :)

---

Ctrl-scroll holds the point under the pointer across the picture, and stops short only where panning further would pull a letterbox bar into view.

Blue dot on the picture, white ring on the mouse, and the bars shaded:

the aimed point holding at the middle and clamping near the top

These two threads are still open: [a trackpad can zoom but not pan](https://github.com/suitenumerique/meet/pull/1498#discussion_r3766935785), and [three callbacks that differ only in the next zoom value](https://github.com/suitenumerique/meet/pull/1498#discussion_r3766935796).

src/frontend/src/features/rooms/livekit/components/ScreenShareZoomControls.tsx
Ovgodd: good catch, I will mask it for now in the PiP and do it in another PR !

Ovgodd:
> Other then this points it should be good, I will do a last manual pass! :)
> 
> Ctrl-scroll holds the point under the pointer across the picture, and stops short only where panning further would pull a letterbox bar into view.
> 
> Blue dot on the picture, white ring on the mouse, and the bars shaded:
> 
> the aimed point holding at the middle and clamping near the top [ the aimed point holding at the middle and clamping near the top ](https://raw.githubusercontent.com/davd-gzl/agent-artifact/main/meet/1498/where-the-clamp-stops-the-pan.gif) [ ](https://raw.githubusercontent.com/davd-gzl/agent-artifact/main/meet/1498/where-the-clamp-stops-the-pan.gif)
> 
> These two threads are still open: [a trackpad can zoom but not pan](https://github.com/suitenumerique/meet/pull/1498#discussion_r3766935785), and [three callbacks that differ only in the next zoom value](https://github.com/suitenumerique/meet/pull/1498#discussion_r3766935796).

Thank you for the reviews !

src/frontend/src/locales/de/rooms.json
```diff
@@ -756,6 +756,24 @@
     "muteParticipant": "{{name}} stummschalten",
```
lebaudantoine, line 756: please note that Spanish has been introduced since your work. Could you rebase your branch and add the spanish support to this pr?

CHANGELOG.md
```diff
@@ -259,6 +259,10 @@ and this project adheres to
 - ♿️(frontend) focus side panel container on open #1452
 - 🐛(summary) whisper call error handling
 
+### Added
+
+- ✨(frontend) add screen share zoom controls #1498
+
```
lebaudantoine, line 265: to be rebased

src/frontend/src/features/participantTile/components/participantTileFocus/ParticipantTileFocus.tsx
```diff
@@ -1,44 +1,22 @@
...
-const MOUSE_IDLE_TIME = 3000
 
 type FadeOverlayProps = {
   children: ReactNode
-  hasKeyboardFocus: boolean
+  isVisible: boolean
 }
 
-const FadeOverlay = ({ children, hasKeyboardFocus }: FadeOverlayProps) => {
-  const [active, setActive] = useState(false)
```
lebaudantoine, line 20: Have you moved this state higher up the React component tree? It looks like ParticipantTile now owns it, which causes all its children to re-render whenever it changes. That feels a bit suboptimal to me, could we keep the state closer to where it’s used?

https://github.com/user-attachments/assets/24fe98f6-3193-492f-a403-33716a33f9ee

src/frontend/src/features/rooms/livekit/components/ScreenShareZoomControls.tsx
```diff
@@ -0,0 +1,223 @@
...
+      className={css({
+        position: 'absolute',
+        bottom: '12px',
+        right: '12px',
+        zIndex: 2,
+        pointerEvents: 'auto',
+      })}
+    >
+      {/* react-aria Toolbar: left/right arrows move between the controls and
+          Tab leaves the group as a whole, as the toolbar role implies. */}
```
lebaudantoine, line 104: un-necessary comment IMO

src/frontend/src/features/rooms/livekit/components/ScreenShareZoomControls.tsx
```diff
@@ -0,0 +1,223 @@
...
+  onZoomOut,
+  onResetZoom,
+}: ScreenShareZoomControlsProps) => {
+  const { t } = useTranslation('rooms', { keyPrefix: 'screenShareZoom' })
+  const announce = useScreenReaderAnnounce()
+
+  const zoomInButtonRef = useRef<HTMLButtonElement>(null)
+  const hadFocusInCollapsibleRef = useRef(false)
+
+  const [isFullscreen, setIsFullscreen] = useState(false)
```
lebaudantoine, line 44: I’d move this state further down the component tree by extracting the fullscreen button into its own component. That component could own the fullscreen state and the useEffect you added to listen for the fullscreenchange event and the toggleFullScreen callback.

we could also memoize the component then.

src/frontend/src/features/participantTile/components/ParticipantTile.tsx
```diff
@@ -98,11 +103,38 @@ export const ParticipantTile: (
...
+  const idleTimerRef = React.useRef<number | null>(null)
+
+  const handleTileMouseMove = React.useCallback(() => {
+    if (idleTimerRef.current) window.clearTimeout(idleTimerRef.current)
+    idleTimerRef.current = window.setTimeout(
+      () => setIsIdle(true),
+      MOUSE_IDLE_TIME
+    )
+    setIsIdle(false)
+  }, [])
```
lebaudantoine, line 118: As mentioned, I would pass the state down.

src/frontend/src/features/participantTile/components/ParticipantTile.tsx
```diff
@@ -120,32 +152,68 @@ export const ParticipantTile: (
     },
   }
 
+  const isVideoTrack =
+    isTrackReference(trackReference) &&
+    (trackReference.publication?.kind === 'video' ||
+      trackReference.source === Track.Source.Camera ||
+      trackReference.source === Track.Source.ScreenShare)
```
lebaudantoine, line 159: Isn't clear from reading this code why my suggestion isn't enough :


```suggestion
  const isVideoTrack = trackReference.publication?.kind === 'video'
```

src/frontend/src/features/rooms/livekit/components/ScreenShareZoomableVideo.tsx
```diff
@@ -0,0 +1,106 @@
...
+import { ScreenShareVideoTrack } from './ScreenShareVideoTrack'
+
+interface ScreenShareZoomableVideoProps {
+  trackRef: TrackReference
+  tileRef: React.RefObject<HTMLDivElement | null>
+  onSubscriptionStatusChanged: (subscribed: boolean) => void
+  manageSubscription?: boolean
+}
+
+export const ScreenShareZoomableVideo = ({
```
lebaudantoine, line 17: As I originally suggested for FadeOverlay, I’d keep the zoom state inside a dedicated wrapper and pass the video as children from its parent. This would isolate zoom updates from the video subtree without needing the memoized component. The key is moving state ownership into the wrapper; simply passing the state down as a prop wouldn’t prevent those re-renders.

src/frontend/src/features/rooms/livekit/hooks/useScreenShareZoom.ts
```diff
@@ -0,0 +1,308 @@
...
+  const subscribe = useCallback((cb: () => void) => {
+    listenersRef.current.add(cb)
+    return () => {
+      listenersRef.current.delete(cb)
+    }
+  }, [])
+
+  const getSnapshot = useCallback(() => snapshotRef.current, [])
+
+  const snapshot = useSyncExternalStore(subscribe, getSnapshot, getSnapshot)
```
lebaudantoine, line 61: I might be missing something, but it looks like we're creating a local store scoped to a single hook instance, with only one consumer. Is that right?

Could we simplify this hook by keeping refs for the imperative transform and React state for the toolbar values? The local subscription store seems unnecessary with a single consumer.

src/frontend/src/features/rooms/livekit/utils/screenShareZoom.ts
```diff
@@ -0,0 +1,154 @@
...
+    isZoomed: zoom > MIN_ZOOM,
+    isDragging: dragging,
+    canZoomIn: zoom < MAX_ZOOM,
+    canZoomOut: zoom > MIN_ZOOM,
+  }
+}
+
+export const getZoomTransform = (zoom: number, pan: PanOffset) => {
+  return `scale(${zoom}) translate(${pan.x}%, ${pan.y}%)`
+}
```
lebaudantoine, line 92: Could we divide the pan delta by the current zoom? With `scale(zoom) translate(...)`, the translation is scaled too, so the image moves faster than the pointer.

For example, in a 1,000px-wide container, dragging 10px produces a `1%` translation. At 4× zoom:

```css
transform: scale(4) translate(1%, 0);
```

This moves the image **40px**, although the pointer only moved **10px**. Dividing the delta by the zoom gives `0.25%`, resulting in the expected 10px of visible movement:

```css
transform: scale(4) translate(0.25%, 0);
```

src/frontend/src/locales/fr/rooms.json
```diff
@@ -756,6 +756,24 @@
...
     "fullScreen": "Plein écran"
   },
+  "screenShareZoom": {
+    "toolbarLabel": "Contrôles de zoom du partage d'écran",
+    "zoomIn": "Zoomer",
+    "zoomInWithShortcut": "Zoomer ({{shortcut}})",
+    "zoomOut": "Dézoomer",
+    "zoomOutWithShortcut": "Dézoomer ({{shortcut}})",
+    "wheelShortcut": "Contrôle plus molette",
+    "wheelShortcutMac": "Commande plus molette",
```
lebaudantoine, line 766: nit: as on other shortcuts, you should use the command symbol, as `⌘+molette`

src/frontend/src/locales/de/rooms.json
Ovgodd: Yes of course !

src/frontend/src/features/participantTile/components/participantTileFocus/ParticipantTileFocus.tsx
Ovgodd: nice catch, fixed : 

https://github.com/user-attachments/assets/c92160d8-5604-4183-95e9-f52c065d7edb

src/frontend/src/features/rooms/livekit/hooks/useScreenShareZoom.ts
Ovgodd: You're right, there's only one consumer. I dropped the local store and kept refs for the transform, with plain state for the toolbar values. Pan and drag don't touch that state at all, since the toolbar doesn't display the position.

src/frontend/src/features/rooms/livekit/utils/screenShareZoom.ts
Ovgodd: yes of course

lebaudantoine (approved):

PR #1499, opened 2026-07-10
✨(all) allow forcing SSO display name for authenticated users

lebaudantoine:
Add a new setting that controls whether an authenticated user can
rename or modify their display name, or if they must use the one
returned by the SSO.

When the setting is disabled, only anonymous users can set a display
name freely; authenticated users always use their SSO display name.

Requested by many self-hosters.


lebaudantoine:
@greptileai

PR #1507, opened 2026-07-13
♿️(frontend) close side panel with Escape key

Ovgodd:
## Purpose

Side panels can be closed with the ✕ button, but keyboard users have no equivalent shortcut. Escape should close the panel and return focus to the trigger button.

## Proposal

- [x] Close side panel with Escape when focus is inside the panel
- [x] Restore focus on the trigger button that opened the panel
- [x] If a second panel is opened without closing the first, restore focus on the last opened panel trigger (not the first one)
- [x] Allow Escape from the chat textarea to close the panel

### DIFF
--- a/src/frontend/src/hooks/useEscapeToClose.ts
+++ b/src/frontend/src/hooks/useEscapeToClose.ts
@@ -1,0 +1,26 @@
+import { useEffect, useRef, type RefObject } from 'react'
+
+export const useEscapeToClose = (
+  isActive: boolean,
+  containerRef: RefObject<HTMLElement | null>,
+  onClose: () => void
+) => {
+  const onCloseRef = useRef(onClose)
+  useEffect(() => {
+    onCloseRef.current = onClose
+  })
+
+  useEffect(() => {
+    if (!isActive) return
+
+    const handleKeyDown = (e: KeyboardEvent) => {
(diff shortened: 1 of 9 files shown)


src/frontend/src/hooks/useEscapeToClose.ts
```diff
@@ -0,0 +1,26 @@
...
+  useEffect(() => {
+    onCloseRef.current = onClose
+  })
+
+  useEffect(() => {
+    if (!isActive) return
+
+    const handleKeyDown = (e: KeyboardEvent) => {
+      if (e.key !== 'Escape') return
+      if (!containerRef.current?.contains(document.activeElement)) return
```
davd-gzl, line 18: The panel only closes while focus is still inside it, so any click elsewhere in the room stops Escape working, and it never fires from a [subpanel](https://github.com/suitenumerique/meet/pull/1507#issuecomment-4957737101).

escape-click-outside

lebaudantoine (commented):
LGTM! Could you please rebase your branch (and the Changelog) against main?

Only closing on ESC if the side panel is focused is fine, though I'm not entirely sure if there are accessibility implications to consider here.

lebaudantoine (approved):

PR #1436, opened 2026-06-22
📝(doc) create complete documentation website for Meet

unteem:
Build a structured documentation website that centralizes guidance for users, administrators, and contributors.
This documentation replaces the previous documentation structure

Highlights include:

- Product overview and feature documentation
- Self-hosting guides (Docker Compose and Kubernetes) Configuration and integration references
- API, security, and troubleshooting documentation
- Development workflows and contribution guides
- FAQ and operational best practices
- Remove of outdated and duplicate documentation sources

Documentation is rendered on this temp URL:  https://documentation-meet.beta.numerique.gouv.fr/self-hosting/compose/quick-start/

In order to test the guides you will need replace var RAW to point to my branch:
```
RAW="https://raw.githubusercontent.com/suitenumerique/meet/main"
to 
RAW="https://raw.githubusercontent.com/unteem/meet/feat/documentation"
```

PR #1509, opened 2026-07-16
(docs) Fix CSS theming env var name

dawagner:
While here, also mention *where* this env var should be set.


lebaudantoine:
@unteem we should take it into account for #1436

lebaudantoine:
Thank you for your contribution @dawagner. Let's connect on matrix, https://matrix.to/#/#meet-official:matrix.org

PR #1516, opened 2026-07-20
✨(frontend) share OneToOneFocusLayout between PiP and main room

Ovgodd:
## Purpose

Unify the 1-to-1 call layout between PiP and the main room so both show the remote participant fullscreen with a local thumbnail overlay, instead of equal side-by-side tiles in the room.

Capture d&#39;écran 2026-07-20 111749

## Proposal

* [x] Extract `getTrackKey` into shared `layout/utils/trackSelection`
* [x] Add shared `OneToOneFocusLayout` (main + bottom-right thumbnail)
* [x] Replace `PipFocusLayout` with `OneToOneFocusLayout` in PiP stage
* [x] Use `OneToOneFocusLayout` in the main room when ≤2 cameras and no pin
* [x] Ensure participant tiles fill the main/thumbnail slots (camera off included)


src/frontend/src/features/pip/components/layout/PipStage.tsx
```diff
@@ -59,9 +59,11 @@ export const PipStage = () => {
     if (cameraTracks.length <= 1) {
       return (
         <StageFrame>
-          <PipFocusLayout
+          <OneToOneFocusLayout
```
davd-gzl, line 62: `OneToOneFocusLayout` is incorrectly named, as it is also used for solo camera.
`FocusThumbnailLayout`, `MainWithThumbnailLayout`, ...  would better suit the general idea.

src/frontend/src/features/rooms/livekit/prefabs/VideoConference.tsx
```diff
@@ -123,6 +124,26 @@ export function VideoConference({ ...props }: VideoConferenceProps) {
...
+    const remote = cameraTracks.find((t) => !t.participant?.isLocal)
+    const local = cameraTracks.find((t) => t.participant?.isLocal)
+    return remote ?? local
+  }, [isOneToOne, cameraTracks])
+
+  const oneToOneThumbnailTrack = useMemo(() => {
+    if (!isOneToOne) return undefined
+    const local = cameraTracks.find((t) => t.participant?.isLocal)
+    return oneToOneMainTrack === local ? undefined : local
+  }, [isOneToOne, cameraTracks, oneToOneMainTrack])
```
davd-gzl: Refactor: `find` hands back an element of `cameraTracks`, so these two hooks cache nothing the array does not already hold.

```suggestion
  const local = cameraTracks.find((t) => t.participant.isLocal)
  const remote = cameraTracks.find((t) => !t.participant.isLocal)
  const oneToOneMainTrack = remote ?? local
  const oneToOneThumbnailTrack = remote ? local : undefined
```

davd-gzl (commented):
Rebase onto main before this lands, otherwise this lgtm!

image-1786378002357

PR #1522, opened 2026-07-21
🐛(transcription) fix silent fail of speaker assignment

cameledev:
Fix bug introduced by #1362  which caused speaker assignment to silently fail. Bug was due to change in input variable type.

NB: Manual testing shows working speaker assignement

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,10 @@ and this project adheres to
 
 ## [Unreleased]
 
+## Fixed
+
+- 🐛(transcription) fix silent bug in speaker assignment
+
 ## [1.24.0] - 2026-07-21
 
 ### Added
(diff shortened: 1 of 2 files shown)


src/backend/core/recording/event/notification.py
```diff
@@ -386,7 +386,7 @@ def _notify_summary_service_v2(recording: models.Recording):
 
         try:
             response = requests.post(
-                settings.SUMMARY_SERVICE_ENDPOINT,
+                settings.SUMMARY_SERVICE_ENDPOINT,  # create_transcribe_task_v2
```
cameledev: added for devex

lebaudantoine: I think this couples the API call a bit too tightly with what's happening under the hood in the summary service.

The method name already makes it clear that we're calling the v2 endpoint, so I don't think the extra comment adds much value.

Comments like this can also be a bit hard to maintain and may end up getting out of sync with the implementation over time.

lebaudantoine (approved):
LGTM

lebaudantoine (approved):

PR #1524, opened 2026-07-21
Fix tests for fix speaker assignment

cameledev:
## Purpose

Fix broken tests following fix of speaker assignment (https://github.com/suitenumerique/meet/pull/1522)

### DIFF
--- a/src/summary/tests/unit/test_user_assign.py
+++ b/src/summary/tests/unit/test_user_assign.py
@@ -1,7 +1,6 @@
 """Tests for the speaker-to-user assignment service."""
 
 import math
-from dataclasses import dataclass, field
 from datetime import datetime
 
 from summary.core import user_assign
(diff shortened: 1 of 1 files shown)


src/summary/tests/unit/test_user_assign.py
```diff
@@ -17,11 +16,9 @@
 )
 
 
-@dataclass
-class FakeTranscription:
-    """Mimics the OpenAI Transcription pydantic model for testing."""
-
-    segments: list = field(default_factory=list)
+def make_transcription(segments: list | None = None) -> dict:
```
lebaudantoine, line 19: Do we really need a utility function for this? What do you think about inlining it instead?

cameledev: I find the original proposal neater as it lays out a schema for a transcription output. In each test, we can focus on the variable parts rather than checking if the transcription is well formed.

lebaudantoine (approved):

PR #1528, opened 2026-07-22
♻️(backend) factorize s3 client creation in utils

cameledev:
## Purpose

Remove code duplication


lebaudantoine (approved):
LGTM, historically imported from LaSuite Drive, we could probably backport those utils in the Django LaSuite organization. You can ask in the Internal dev chan.

PR #1533, opened 2026-07-24
✨(frontend) add connection test page 

arnaud-robin:
## Purpose

Users currently have no reliable way to check browser, media devices, and LiveKit connectivity before joining a room. Support also lacks a simple diagnostic report when people hit network issues (e.g. `could not establish pc connection`).

## Proposal

Add a connection test page that runs local device checks plus LiveKit `ConnectionCheck` (WebSocket, WebRTC, TURN, publish audio/video, etc.), backed by a throttled `GET /api/v1.0/connection-test/` endpoint that issues a short-lived token for an ephemeral room (including for anonymous users). When the run finishes, users can download a JSON report of the results.

### DIFF
--- a/src/backend/core/services/livekit_events.py
+++ b/src/backend/core/services/livekit_events.py
@@ -137,6 +137,13 @@ class LiveKitEventsService:
 
         room_name = data.room.name or data.egress_info.room_name
 
+        if self._is_connection_test_room(room_name):
+            logger.info(
+                "Ignoring webhook event for connection test room '%s'.",
+                room_name,
+            )
+            return
+
         if self._filter_regex and not self._filter_regex.search(room_name):
             logger.info("Filtered webhook event for room '%s'", room_name)
             return
@@ -227,6 +234,11 @@ class LiveKitEventsService:
                 )
 
         # Silently ignoring EGRESS_ABORTED, EGRESS_FAILED
+
+    @staticmethod
+    def _is_connection_test_room(room_name: str) -> bool:
+        """Return True for ephemeral rooms created by the connection test endpoint."""
+        return room_name.startswith(settings.CONNECTION_TEST_ROOM_PREFIX)
 
     def _handle_room_started(self, data):
         """Handle 'room_started' event."""
--- a/src/backend/core/tasks/connection_test.py
+++ b/src/backend/core/tasks/connection_test.py
@@ -1,0 +1,39 @@
+"""Tasks related to connection test rooms."""
+
+import logging
+
+from django.conf import settings
+
+from core.services.room_management import (
+    RoomManagement,
+    RoomManagementException,
+    RoomNotFoundException,
+)
+from core.tasks._task import task
+
+logger = logging.getLogger(__name__)
+
+
+@task
+def delete_connection_test_room(room_name: str):
+    """Force-delete an ephemeral connection-test room.
+
+    Used as a hard cap so a participant cannot keep an auto-refreshed
+    LiveKit session open indefinitely after requesting a test token.
+    """
+    prefix = settings.CONNECTION_TEST_ROOM_PREFIX
+    if not room_name.startswith(prefix):
+        logger.error(
+            "Refusing to delete room '%s': expected prefix '%s'.",
+            room_name,
+            prefix,
+        )
+        return
+
+    try:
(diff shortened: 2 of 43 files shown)


src/frontend/src/layout/Footer.tsx
```diff
@@ -266,6 +266,16 @@ export const Footer = () => {
...
+          <StyledLi divider>
+            <Link
+              underline={false}
+              footer="minor"
+              to="/test-connection"
+              aria-label={t('links.connectionTest')}
+            >
+              {t('links.connectionTest')}
+            </Link>
+          </StyledLi>
```
lebaudantoine: Note: Unfortunately, the footer is hidden on self-hosted instances. This isn't blocking for the merge, but it would be great to give self-hosters the option to display it as well.

lebaudantoine:
@greptileai

src/backend/core/api/connection_test.py
```diff
@@ -0,0 +1,51 @@
...
+    ConnectionTestAnonRateThrottle,
+    ConnectionTestUserRateThrottle,
+)
+from core.tasks.connection_test import delete_connection_test_room
+from core.utils import generate_token
+
+CONNECTION_TEST_USERNAME = "Test connexion"
+
+
+@api_view(["GET"])
```
lebaudantoine: It would be better to use a POST instead of a GET. This is an issue we currently have: only POST requests are protected by CSRF protection. GET should not be used for state-changing operations (e.g. creating a Celery worker) or for issuing a token, as it is neither protected by CSRF nor aligned with HTTP semantics.

src/backend/core/api/connection_test.py
```diff
@@ -0,0 +1,51 @@
...
+from rest_framework.response import Response
+
+from core.api.throttling import (
+    ConnectionTestAnonRateThrottle,
+    ConnectionTestUserRateThrottle,
+)
+from core.tasks.connection_test import delete_connection_test_room
+from core.utils import generate_token
+
+CONNECTION_TEST_USERNAME = "Test connexion"
```
lebaudantoine: It's a French default value

src/backend/core/urls.py
```diff
@@ -46,6 +47,11 @@
                 *router.urls,
                 *oidc_urls,
                 path("config/", get_frontend_configuration, name="config"),
+                path(
+                    "connection-test/",
+                    get_connection_test_config,
+                    name="connection_test",
+                ),
```
lebaudantoine: The current flat naming is fine. That said, I wonder if it would make sense to introduce a dedicated viewset for diagnostics, as we'll likely add more of them in the future.

Something like /api/v1.0/diagnostics/connection/ would give us a cleaner and more extensible structure.

src/backend/core/tasks/connection_test.py
```diff
@@ -0,0 +1,51 @@
...
+        logger.info("Deleted connection test room '%s'.", room_name)
+    except TwirpError as exc:
+        # Room may already be gone after empty/departure timeout.
+        logger.info(
+            "Could not delete connection test room '%s': %s",
+            room_name,
+            exc,
+        )
+    finally:
+        await lkapi.aclose()
```
lebaudantoine: Could we add this to the RoomManagement service?

src/backend/core/services/livekit_events.py
```diff
@@ -256,6 +268,13 @@ def _handle_room_started(self, data):
     def _handle_room_finished(self, data):
         """Handle 'room_finished' event."""
 
+        if self._is_connection_test_room(data.room.name):
+            logger.info(
+                "Ignoring room_finished event for connection test room '%s'.",
+                data.room.name,
+            )
+            return
```
lebaudantoine: why not add this check in the `receive_method` instead?

lebaudantoine (commented):
Backend review

src/backend/core/urls.py
arnaud-robin: Agreed that it could make sense if we ever wish to study rtcstats in more depth in the future. I think this should be done in a further PR

lebaudantoine:
@greptileai

lebaudantoine (approved):

PR #1541, opened 2026-07-31
Exposes media state in the DOM

lebaudantoine:
Exposes the local participant's media state in the DOM so external tools
(e.g. bots automating the frontend) can reliably read the microphone and
camera state, and watch for changes with a MutationObserver:

```
const el = document.getElementById('media-state')
new MutationObserver(...).observe(el, { attributes: true })
```


lebaudantoine:
@greptileai

Issue #615, opened 2025-07-02
Add PWA icon


lebaudantoine:
## Bug Report

When installing the web app as a PWA on IOS homescreen, no icon is shown.

davd-gzl:
No iOS device here, so this is a question rather than a verdict.

b34a1c86 added `apple-touch-icon.png`, and [`index.html`](https://github.com/suitenumerique/meet/blob/8cbcad76/src/frontend/index.html#L5) links it at 180x180, which is what [iOS reads for the homescreen](https://developer.apple.com/library/archive/documentation/AppleApplications/Reference/SafariWebContent/ConfiguringWebApplications/ConfiguringWebApplications.html#//apple_ref/doc/uid/TP40002051-CH3-SW4). Still blank on a real device, or can this close?

Separate from the icon: [`site.webmanifest`](https://github.com/suitenumerique/meet/blob/8cbcad76/src/frontend/public/site.webmanifest#L1) declares no `start_url`. [Without one, the page that linked the manifest becomes it](https://developer.mozilla.org/en-US/docs/Web/Progressive_web_apps/Manifest/Reference/start_url), and rooms are top-level paths in [`routes.ts`](https://github.com/suitenumerique/meet/blob/8cbcad76/src/frontend/src/routes.ts#L52), so an install begun inside a room reopens that room on every launch. #1549 declares it, and fills the empty `name` and `short_name` while it is there.

PR #1549, opened 2026-08-02
🐛(frontend) stop the installed app reopening the room it came from

davd-gzl:
[`site.webmanifest`](https://github.com/suitenumerique/meet/blob/8cbcad76/src/frontend/public/site.webmanifest#L1) declares no `start_url`, so [the page that linked the manifest becomes one](https://developer.mozilla.org/en-US/docs/Web/Progressive_web_apps/Manifest/Reference/start_url). Rooms are top-level paths in [`routes.ts`](https://github.com/suitenumerique/meet/blob/8cbcad76/src/frontend/src/routes.ts#L52), so installing from inside a room gives an app that reopens that room every time it is launched.

The same file ships `name` and `short_name` as empty strings. That is survivable today: Chromium falls back to the document title, which [`index.html`](https://github.com/suitenumerique/meet/blob/8cbcad76/src/frontend/index.html#L10) fills from `VITE_APP_TITLE`, so an install on Android is labelled correctly. But [the manifest specification](https://www.w3.org/TR/appmanifest/#name-member) defines no such fallback, so the label rests on behaviour no browser owes us.

The title cannot be hardcoded: the DINUM build sets `VITE_APP_TITLE` to `Visio` in [`docker/dinum-frontend/Dockerfile`](https://github.com/suitenumerique/meet/blob/8cbcad76/docker/dinum-frontend/Dockerfile#L19). Vite substitutes `%VITE_APP_TITLE%` in `index.html` only, and copies `public/` verbatim.

So the manifest moves out of `public/`, declares `start_url`, and [`vite.config.ts`](https://github.com/davd-gzl/meet/blob/0303ec2f/src/frontend/vite.config.ts#L29-L44) writes the title in through the [`viteStaticCopy`](https://github.com/sapphi-red/vite-plugin-static-copy) target already used for the MediaPipe assets. The frontend [`Dockerfile`](https://github.com/suitenumerique/meet/blob/8cbcad76/src/frontend/Dockerfile#L36-L38) declares the `VITE_APP_TITLE` build argument as well, which it never did, so the value [`compose.yml`](https://github.com/suitenumerique/meet/blob/8cbcad76/compose.yml#L165) passes stops being dropped.

Built with the variable set, unset, and on the dev server: `dist/site.webmanifest` carries the right name each time.

Refs #615

### DIFF
--- a/src/frontend/vite.config.ts
+++ b/src/frontend/vite.config.ts
@@ -26,6 +26,20 @@ export default defineConfig(({ mode }) => {
             dest: 'assets/mediapipe/wasm',
             rename: { stripBase: 4 },
           },
+          {
+            // Kept out of public/ so the instance title reaches the manifest,
+            // the way index.html gets it through %VITE_APP_TITLE%.
+            src: 'site.webmanifest',
+            dest: '.',
+            transform: (content) => {
+              const title = env.VITE_APP_TITLE
+              return JSON.stringify({
+                ...JSON.parse(content),
+                name: title,
+                short_name: title,
+              })
+            },
+          },
         ],
       }),
       env.VITE_ANALYZE === 'true' &&
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -41,6 +41,7 @@ and this project adheres to
 - 🐛(frontend) fall back to user.full_name on request-entry
 - 🚸(frontend) show two initials in the Avatar when possible
 - 🩹(all) clear the SonarCloud reliability finding and the lint debt
+- 🐛(frontend) stop the installed app reopening the room it came from
 
 ## [1.24.0] - 2026-07-21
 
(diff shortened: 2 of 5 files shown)


src/frontend/vite.config.ts
```diff
@@ -26,6 +26,22 @@ export default defineConfig(({ mode }) => {
...
           },
+          {
+            // Kept out of public/ so the instance title reaches the manifest,
+            // the way index.html gets it through %VITE_APP_TITLE%.
+            src: 'site.webmanifest',
+            dest: '.',
+            transform: (content) => {
+              // An unset variable and one set to the empty string both have to
+              // land on the fallback, so this tests truthiness, not nullishness.
+              const title = env.VITE_APP_TITLE || 'LaSuite Meet'
```
lebaudantoine: I'm always a bit confused by how Vite handles environment variables, but I think this fallback isn't necessary. Defining the values in env.development and env.production should be enough.

Please feel free to correct me if i'm wrong.

lebaudantoine (approved):
Minor comment

src/frontend/vite.config.ts
davd-gzl: That's right, and if not set it will fallback like before on the document title or the URL, which is not incoherent since env production variable are pushed on the repo, if not it was removed intentionally by author (unset == default behavior).
Thank you for pushing the fix!

Issue #1547, opened 2026-08-01
Share images in the chat


davd-gzl:
## Feature Request

A participant sends one image in the chat, with an optional caption. Everyone present sees it inline. It is gone when the meeting ends, and no filename travels with the picture.

This is the image half of #948, which asks for documents. It is worth splitting because images need less: they render inline, so nothing is ever downloaded, and they can be restricted to bytes that actually decode as an image. Arbitrary files need a download path, which raises questions images do not, such as what a browser does with an unknown type and whether anything needs scanning. #948 stays open for those.

#965 tried this with presigned S3 URLs and was closed in favour of LiveKit byte streams and ephemeral attachments. This should follow that direction.

Two tasks:

- [ ] Send, receive, enlarge and save an image
- [ ] Strip metadata before an image leaves the sender

The second is separate because doing it in a browser means either re-encoding every image, which ruins the lossless screenshots this exists for, or hand-writing a parser per format. Until it lands, an image carries whatever its camera wrote.

Happy to build both.

PR #965, opened 2026-02-10
drag and drop, upload files in chat feature

darthvader58:
## Proposal

Add file sharing in chat:

  - Allow users to drag and drop a file into chat
  - Allow users to click a + button to choose a file
  - Show the file in chat so other participants can download it
  - [x] item 1: Add + button file upload in chat
  - [x] item 2: Add drag-and-drop file upload in chat
  - [x] item 3: Supports multiple files extensions (backend api endpoints built too)

Added this feature according to the fix #948 
@Ovgodd @lebaudantoine

Issue #948, opened 2026-02-03
send/share a file in the chat


Vincentparisgithub:
**Is your feature request related to a problem or unsupported use case? Please describe.**
Some users like to share some documents directly in the chat, so the participants are able to download the file.

**Describe the solution you'd like**
We could drag and drop a file into the chat or click on a button "+"

<img width="1080" height="545" alt="Image" src="https://github.com/user-attachments/assets/5185ac43-eca8-4ca8-8368-e1f325b9a4b5" />

**Describe alternatives you've considered**

**Discovery, Documentation, Adoption, Migration Strategy

Do you want to work on it through a Pull Request?**

Vincentparisgithub:
somebody created the same feautre few days ago : 
drag and drop, upload files in chat feature#965
https://github.com/suitenumerique/meet/pull/965

and sorry i juste saw that the feature is in Projects : 

<img width="1260" height="620" alt="Image" src="https://github.com/user-attachments/assets/02bc7b0b-4f71-4191-bee8-d2e5bf96ec29" />

PR #1551, opened 2026-08-03
✨(all) share images in the meeting chat

davd-gzl:
Participants can drop a screenshot into the chat instead of leaving the call to send it. [#948](https://github.com/suitenumerique/meet/issues/948) asked for this in February, and [#1547](https://github.com/suitenumerique/meet/issues/1547) splits the image half out of it. This is its first task. Six of the calls sit in a [decision record](https://claude.ai/code/artifact/334a19c4-468b-42c8-868f-c9c4ebb33a13) with the settled ones, each showing what it rejected.

Images move as [LiveKit byte streams](https://docs.livekit.io/transport/data/byte-streams/) on a `chat-media` topic and never reach object storage, following your note on [#965](https://github.com/suitenumerique/meet/pull/965). Nothing is provisioned, nothing is retained, nothing is purged, and an image is gone when the meeting ends.

Written with `streamBytes` rather than `sendFile`. The convenience wrapper accepts only a topic, a MIME type and destinations, and this needs three fields it drops. The caption rides in the stream's attributes, so an image and its text are one stream rather than two messages to reconcile, and an attachment cannot be forged by typing a marker into the message box, which is how #965's were. The name is `image.png`, never the sender's. The total size is what lets a receiver show a percentage instead of an indeterminate bar.

```
sender                                    receiver
  sniff type from the bytes                 handler fires as the stream opens
  downscale if over 5 MB                    row appears, empty, with progress
  streamBytes(topic, name,      ──────▶     re-check length once assembled
    mimeType, totalSize,                    sniff type again, ignore the claim
    attributes: {caption, w, h})            decode, then render
  write 15 kB chunks, close                 ▲
  insert own row locally                    └── one stream is one chat row
```

The receiver trusts nothing the sender declares, because a room admits unauthenticated participants. The declared size is checked before the stream is read and the assembled length again after it, the declared MIME type is ignored in favour of the payload's own leading bytes, and the result must decode as an image before it is shown. `image/svg+xml` is absent from the allowlist because SVG executes script, and a blob URL is only ever an `img` source or an anchor's download target, never navigated, since `blob:` is same-origin.

An image over 5 MB is reduced rather than refused. An animation over 5 MB is refused instead: the browser has no GIF encoder, and flattening it to a single frame is a silent surprise.

Metadata is deliberately not stripped, which is the second task on #1547. Doing it in a browser means either re-encoding every image, which ruins the lossless screenshots this feature exists for, or hand-writing a parser for each format, and neither belongs in the change that introduces the feature. Only the filename is withheld, which needs no code to withhold. Late joiners see no images, exactly as they see no earlier messages.

One thing rides along, because the feature does not work without it: the unread count moves from LiveKit's `chatMessages` to the row list. Byte streams never appear in `chatMessages`, and rows are the only place that sees both kinds.

## Visual evidence

Left is the sender, right the receiver, in one room. A screenshot dragged in and captioned, a photo through the picker, the receiver enlarging one, and a 6.8 MB PNG reduced on the way out.

receiver

animation-refused

[chat-image-sharing.webm](https://github.com/user-attachments/assets/a0ecc9eb-ca20-4d12-92d3-c95b0ef7eb48)

## Verification

Every commit typechecks alone. CI builds only the tip of the branch, and this repository rebase-merges, so a commit that breaks in the middle would reach `main` with nothing to catch it.

The frontend has no test runner, so the rest was run by hand: two participants driven through a real meeting against a local stack, 34 checks, no console errors on either side. Drag and drop, the file picker and paste each stage an image. The caption travels with it and becomes the alt text. A PNG arrives byte-identical rather than re-encoded, an animated GIF arrives still animated rather than flattened, and a 6.8 MB PNG arrives at 1.71 MB. An 8.5 MB animation is refused with the message that explains why. SVG and text files do not even highlight the drop target. The lightbox opens, downloads under the sniffed extension and closes on Escape. A late joiner sees nothing earlier, and a sender vanishing mid-transfer leaves no row spinning.

---

Assisted by AI. The design calls are mine.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ and this project adheres to
 
 ### Added
 
+- ✨(all) share images in the meeting chat #1547
 - ✨(summary) report exception type in failure analytics
 - ✨(frontend) add configurable documentation menu item
 - ✨(frontend) allow promoting authenticated participants
--- a/src/backend/core/api/__init__.py
+++ b/src/backend/core/api/__init__.py
@@ -60,6 +60,12 @@ def get_frontend_configuration(request):
                 "allowed_mimetypes"
             ],
         },
+        "chat_media": {
+            "enabled": settings.CHAT_MEDIA_ENABLED,
+            "max_size": settings.CHAT_MEDIA_MAX_SIZE,
+            "allowed_mimetypes": settings.CHAT_MEDIA_ALLOWED_MIMETYPES,
+            "allowed_extensions": settings.CHAT_MEDIA_ALLOWED_EXTENSIONS,
+        },
         "telephony": build_telephony_config(),
         "resource": {
             "default_access_level": settings.RESOURCE_DEFAULT_ACCESS_LEVEL,
--- a/src/backend/core/tests/test_api_config.py
+++ b/src/backend/core/tests/test_api_config.py
@@ -1,0 +1,65 @@
+"""
+Test the frontend configuration endpoint.
+"""
+
+from rest_framework.test import APIClient
+
+
+def test_api_config_is_public():
+    """The configuration is readable without authentication."""
+    response = APIClient().get("/api/v1.0/config/")
+
+    assert response.status_code == 200
+
+
+def test_api_config_chat_media_defaults():
+    """Chat media is advertised with its default limits and enabled by default."""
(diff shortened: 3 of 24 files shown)


lebaudantoine:
let's connect on https://matrix.to/#/#meet-official:matrix.org

PR #1557, opened 2026-08-04
✨ (backend) POC: Store recordings in LaSuite Drive

cameledev:
## Purpose

Store video recordings made by Meet in [LaSuite Drive](https://github.com/suitenumerique/drive). This will allow for longer access to files by end users, reduce dependency on email links and help create usage for the LaSuite ecosystem.

## Proposal

- Create a shared local dev environment for meet and drive:
  -  Most importantly, both will share the **Keycloack** OIDC server (match v. 26.3.2 of drive)
  - Additionaly, to save resources, we also share **postgres** and **mailcatcher**
  - Create local *lasuite* docker network
- Use the [existing implementation of resource server in drive](https://github.com/suitenumerique/drive/blob/main/docs/resource_server.md#)
- Make the upload call from a worker (which may be queued)
- Store the owner access token encrypted in the recording object when starting the recording (as session user is not accessible from worker)
- When recording is complete, stream S3 file through to drive (push is done in meet)

### How to use

- Use this current branch of meet
- Use this branch of drive: https://github.com/suitenumerique/drive/pull/796
- make sure both projects are using the right env variables (see changes to defaults in both PRs)
- `make migrate` drive before run
- Sign in with same user in meet and drive
- Start a visio, select *record a video*. After stopping, the video should appear in the user's drive
- NB: if not working, it may be due to access token timeout, in this case, logout, login and run the experiment again emidiately.

### Limitiation of this POC for dev

- Access token has limited validity. So this code only works for short recordings launched shortly after login

### Limitiations of this POC for prod

**Problem:** This POC is not meant for prod with Visio/Fichiers with the current OIDC provider as access tokens and session durations do not match. ProConnect has: (12h session, 2h refresh token, 1h access token) whereas Visio has: (7d session). 

**Solution:** Ultimately, we expect to Visio/Drive to use [Menshen](https://github.com/suitenumerique/menshen) as a Token Exchange server, which would issue long-lived tokens allowing the narrow task of "pushing to Fichiers for User X at Path Y", potentially with extra restrictions.

## Other discussed options

- Server to server through API key (may be implemented as a temporary solution depending on temporality of this being put into production)
-  (orthogonal) "Pull from drive" instead of "push from meet"
- (orthogonal) Delete files from meet S3 once transfered

image

## Todo 

- [ ] Check drive file size limitation and document it
- [ ] Propose upload as a configurable class so that other services may be used for storage


src/backend/core/tests/services/test_livekit_events.py
```diff
@@ -159,12 +159,13 @@ def test_handle_egress_updated_non_handled(
 )
 @mock.patch("core.utils.notify_participants")
 @mock.patch("core.services.room_management.RoomManagement.update_metadata")
-def test_handle_egress_ended_metadata_update_fails(  # noqa: PLR0913 # pylint: disable=too-many-arguments, too-many-positional-arguments
-    mock_update_metadata, mock_notify, mode, notification_type, service, settings
+# Without storage events, completion falls back to the egress event itself.
+@override_settings(RECORDING_STORAGE_EVENT_ENABLE=False)
+def test_handle_egress_ended_metadata_update_fails(
```
cameledev: unrelated

lebaudantoine: you can rebase on main !

src/backend/core/tasks/push_recording.py
```diff
@@ -0,0 +1,110 @@
...
+
+
+def _build_filename(recording: models.Recording) -> str:
+    """Return a filename for the Drive item."""
+
+    return (
+        f"{recording.room.slug}-"
+        f"{recording.created_at:%Y-%m-%d-%H-%M}."
+        f"{recording.extension}"
+    )
```
cameledev, line 27: Change to configurable template

PR #1559, opened 2026-08-04
🐛(frontend) keep Unicode initials intact in avatar

arnaud-robin:
## Purpose
Participant avatar initials break when a display name starts with a multi-unit Unicode character (for example an  emoji such as🦎). When the camera is off, the placeholder shows a broken glyph (?) instead of the intended character.

## Proposal
Extract avatar initials using the first user-perceived character (grapheme), with a code-point-safe fallback, so Unicode characters that span multiple UTF-16 units stay intact.

### DIFF
--- a/src/frontend/src/components/Avatar.tsx
+++ b/src/frontend/src/components/Avatar.tsx
@@ -1,5 +1,5 @@
 import { css, cva, RecipeVariantProps } from '@/styled-system/css'
-import React from 'react'
+import React, { useLayoutEffect, useMemo } from 'react'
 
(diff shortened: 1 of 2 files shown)


src/frontend/src/components/Avatar.tsx
lebaudantoine: When it's an emoji, this manually adjusted offset doesn't seems relevant.
I guess my initial strategy to adjust centering manually wasn't the optimal one.
Please feel free to change it.

lebaudantoine:
@greptileai

PR #1564, opened 2026-08-05
🐛(backend) serialize lazy title in summary payload

lebaudantoine:
`_generate_title` returned a lazy gettext_lazy proxy in the recording_datetime is None branch, which `json.dumps` cannot serialize.

This crashed `requests.post(json=payload)` with "Object of type __proxy__ is not JSON serializable" whenever the LiveKit egress lookup failed (started_at=None).

_The lookup fails each time you try to replay a recording past the 24h of the egress' start, the CLI eemptiesits cache._

Force evaluation with `str(_("Transcription"))` inside the active locale.

Add a regression test asserting the v2 payload is a real str and is JSON-serializable when timestamps are unavailable. The existing `without_metadata` test missed this: mocked requests.post never serialized, and a lazy proxy compares equal to its string.

### DIFF
--- a/src/backend/core/recording/event/notification.py
+++ b/src/backend/core/recording/event/notification.py
@@ -9,7 +9,7 @@ from zoneinfo import ZoneInfo, ZoneInfoNotFoundError
 from django.conf import settings
 from django.core.mail import send_mail
 from django.template.loader import render_to_string
-from django.utils.translation import get_language, override
+from django.utils.translation import get_language, gettext, override
 from django.utils.translation import gettext_lazy as _
 
 import aiohttp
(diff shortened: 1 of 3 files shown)


FloChehab (approved):

src/backend/core/recording/event/notification.py
```diff
@@ -192,7 +192,7 @@ def _generate_title(
         """Generate title from context or return default."""
         if recording_datetime is None:
             with override(locale):
-                return _("Transcription")
+                return str(_("Transcription"))
```
FloChehab: Or maybe a non lazy version ?

PR #1566, opened 2026-08-05
🔒️(backend) derive connection-test room max age from token TTL

lebaudantoine:
Refactor CONNECTION_TEST_ROOM_MAX_AGE_SECONDS so it is no longer an independent setting but a quantity derived from (or added on top of) the token TTL.

This prevents a misconfiguration where the token would outlive the delete-room callback. In that case, an attacker holding a valid token could recreate the room after the callback fired and escape the intended cleanup.


lebaudantoine:
@greptileai

PR #1607, opened 2026-08-15
🔒(ci) harden every workflow install and pin every action to a hash

davd-gzl:
## Problem

Every install step in these workflows fetches packages and runs code from them before the job's own commands start, with `GITHUB_TOKEN` in the environment.

Every `uses:` names a tag or a branch, and either one its owner can repoint at different code without the reference changing.

SonarCloud counts 60 security findings across the four files.

### Installs

`pip install` refuses source distributions, so nothing runs a `setup.py` on the runner. `uv run` stops re-resolving what the `uv sync --locked` above it already pinned. `npm ci` and `yarn install` skip dependency lifecycle scripts. The dockerize download stays on https across redirects.

### Pins

All 65 `uses:` name a 40-character commit, each one the commit its old ref already pointed at, with the version in a trailing comment. Nothing about what runs changes.

Renovate is configured to write and update this form, though its GitHub Actions branch has been sitting on the dashboard rather than opening. Either way the refs are as current as they were: a stale tag is not fresher than a stale hash, it just changes underneath you.

### The rename

`meet.yml` says nothing the repository name does not, so it becomes `ci.yml`.

## What this leaves

Five of the 60 findings, all MAJOR. The gate reads the worst finding rather than the count, so it stays red.

Three are `uv sync --locked`, which has nowhere to put `--no-build` while `brevo-python` ships as source only. Two are `pip install .[dev]`, which wants a lock file for `src/summary` that is a maintainer's call rather than this branch's.

wheel: a package that is a zip and gets unpacked.

source distribution: a package that is a tarball whose `setup.py` the installer runs, on the runner, with the job's environment around it.

lifecycle script: `preinstall`, `install`, `postinstall` or `prepare`, which npm and yarn run from every package in the resolved tree, not only the ones `package.json` names.

pin: a `uses:` naming a commit rather than a tag or a branch, so the owner cannot change what runs.

</details>

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,8 @@
+version: 2
+updates:
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "weekly"
+    commit-message:
+      prefix: "⬆️(ci)"
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,8 +1,0 @@
-version: 2
-updates:
-  - package-ecosystem: "github-actions"
-    directory: "/"
-    schedule:
-      interval: "weekly"
-    commit-message:
-      prefix: "⬆️(ci)"
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -31,12 +31,12 @@ jobs:
         if: always()
         run: |
           ! git log | grep 'fixup!'
-      - name: Install gitlint
+      - name: Install uv
         if: always()
-        run: "pip install --user --only-binary=:all: requests==2.34.2 gitlint-core==0.19.1"
+        uses: astral-sh/setup-uv@37802adc94f370d6bfd71619e3f0bf239e1f3b78 # v7.6.0
       - name: Lint commit messages added to main
         if: always()
-        run: ~/.local/bin/gitlint --commits origin/${{ github.event.pull_request.base.ref }}..HEAD
+        run: uvx --no-build --from gitlint-core==0.19.1 gitlint --commits origin/${{ github.event.pull_request.base.ref }}..HEAD
 
   check-changelog:
     runs-on: ubuntu-latest
@@ -153,7 +153,7 @@ jobs:
       - name: Install uv
         uses: astral-sh/setup-uv@37802adc94f370d6bfd71619e3f0bf239e1f3b78 # v7.6.0
       - name: Install the project
-        run: uv sync --locked --all-extras
+        run: uv sync --locked --all-extras --no-build
       - name: Check code formatting with ruff
         run: uv run --no-sync --no-build ruff format . --diff
       - name: Lint code with ruff
@@ -173,13 +173,14 @@ jobs:
         uses: actions/setup-python@ece7cb06caefa5fff74198d8649806c4678c61a1 # v6.3.0
         with:
           python-version: "3.13"
-          cache: "pip"
-      - name: Install development dependencies
-        run: "pip install --user --only-binary=:all: .[dev]"
+      - name: Install uv
+        uses: astral-sh/setup-uv@37802adc94f370d6bfd71619e3f0bf239e1f3b78 # v7.6.0
+      - name: Install the project
+        run: uv sync --locked --all-extras
       - name: Check code formatting with ruff
-        run: ~/.local/bin/ruff format . --diff
+        run: uv run --no-sync --no-build ruff format . --diff
       - name: Lint code with ruff
-        run: ~/.local/bin/ruff check .
+        run: uv run --no-sync --no-build ruff check .
 
   test-back:
     runs-on: ubuntu-latest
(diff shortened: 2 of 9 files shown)


.github/workflows/ci.yml
```diff
@@ -386,10 +388,10 @@ jobs:
         working-directory: src/sdk/library
     steps:
       - name: Checkout repository
-        uses: actions/checkout@v6
+        uses: actions/checkout@d23441a48e516b6c34aea4fa41551a30e30af803 # v6.1.0
 
       - name: Install dependencies
-        run: npm ci
+        run: npm ci --ignore-scripts
```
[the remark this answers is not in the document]
davd-gzl: `@swc/core` ships the native binary in a [platform package listed under `optionalDependencies`](https://github.com/davd-gzl/meet/blob/fix/pin-workflow-installs/src/sdk/package-lock.json#L2134-L2144), which npm installs whatever `--ignore-scripts` says. The postinstall covers the case that list misses: it [installs `@swc/wasm` when the platform package fails to load](https://github.com/swc-project/swc/blob/v1.12.9/packages/core/src/postinstall.ts#L4-L5). [`esbuild`](https://github.com/davd-gzl/meet/blob/fix/pin-workflow-installs/src/sdk/package-lock.json#L3752) resolves the same way, and its install script only [hard-links the binary so the command line starts faster](https://github.com/evanw/esbuild/blob/v0.27.7/lib/npm/node-install.ts#L191-L192).

<details><summary>the two SDK jobs, from a clean tree</summary>

```bash
cd src/sdk/library
rm -rf ../node_modules node_modules
npm ci --ignore-scripts
npm run lint && npm run check && npm run build
```

`@swc/core` finds its binding, and the build emits both bundles:

```console
added 364 packages, and audited 366 packages in 33s

> @gouvfr-lasuite/visio-sdk@0.1.22 build
> tsc -b && vite build

vite v7.3.2 building client environment for production...
✓ 13 modules transformed.
dist/visio-sdk.js  21.75 kB │ gzip: 6.55 kB
dist/visio-sdk.umd.cjs  13.97 kB │ gzip: 5.59 kB
✓ built in 14.53s
```

`node_modules/@swc/` after the install, with the scripts skipped:

```console
core  core-linux-x64-gnu  core-linux-x64-musl  counter  types
```

node 22.23.2, npm 10.9.8, linux x64 gnu.

</details>

.github/dependabot.yml
```diff
@@ -0,0 +1,8 @@
+version: 2
+updates:
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "weekly"
```
lebaudantoine: I'll push for frequent updates when security issues arise, and monthly upgrades on a regular basis. Regarding the repo configuration, we already allowlist specific actions that are permitted to run. Updating every action hash every week would be a pain, without actually making the team more secure or more productive.

.github/workflows/crowdin-download.yml
```diff
@@ -13,10 +13,10 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Checkout
-        uses: actions/checkout@v6
+        uses: actions/checkout@d23441a48e516b6c34aea4fa41551a30e30af803 # v6.1.0
```
lebaudantoine, line 16: We already allowlist specific action hashes in the repository settings, so @main actions cannot run regardless of what the owner pushes. However, making this explicit and versioned directly in the code remains a solid best practice.

lebaudantoine (commented):
I would probably squash your last commit.
Thank you for your contribution

.github/workflows/ci.yml
```diff
@@ -18,22 +18,22 @@ jobs:
       contents: read
     steps:
       - name: Checkout repository
-        uses: actions/checkout@v6
+        uses: actions/checkout@d23441a48e516b6c34aea4fa41551a30e30af803 # v6.1.0
```
Villaquiranm: this instead of having the hash and the comment wdyt ?
```suggestion
        uses: actions/checkout@v6.1.0
```
Same for all other actions

.github/workflows/ci.yml
```diff
@@ -18,22 +18,22 @@ jobs:
...
-          ! git diff origin/${{ github.event.pull_request.base.ref }}..HEAD -- . ':(exclude)**/meet.yml' | grep "print("
+          ! git diff origin/${{ github.event.pull_request.base.ref }}..HEAD -- . ':(exclude).github/workflows/**' | grep "print("
       - name: Check absence of fixup commits
         if: always()
         run: |
           ! git log | grep 'fixup!'
       - name: Install gitlint
         if: always()
-        run: pip install --user requests gitlint
+        run: "pip install --user --only-binary=:all: requests==2.34.2 gitlint-core==0.19.1"
```
jmaupetit: I made a proposal to remove the `requests` dependency and also use `uv` in this job for the Menshen's project: https://github.com/suitenumerique/menshen/pull/73

WDYT?

.github/workflows/ci.yml
```diff
@@ -168,14 +168,14 @@ jobs:
...
+        uses: actions/checkout@d23441a48e516b6c34aea4fa41551a30e30af803 # v6.1.0
       - name: Install Python
-        uses: actions/setup-python@v6
+        uses: actions/setup-python@ece7cb06caefa5fff74198d8649806c4678c61a1 # v6.3.0
         with:
           python-version: "3.13"
           cache: "pip"
       - name: Install development dependencies
-        run: pip install --user .[dev]
+        run: "pip install --user --only-binary=:all: .[dev]"
```
jmaupetit: Why not using`uv` in this job too?

.github/workflows/ci.yml
```diff
@@ -320,21 +322,21 @@ jobs:
...
       - name: Install Python
-        uses: actions/setup-python@v6
+        uses: actions/setup-python@ece7cb06caefa5fff74198d8649806c4678c61a1 # v6.3.0
         with:
           python-version: "3.13"
           cache: "pip"
 
       - name: Install development dependencies
-        run: pip install --user .[dev]
+        run: "pip install --user --only-binary=:all: .[dev]"
```
jmaupetit: ditto

PR #1614, opened 2026-08-19
✨(fullstack) show who is already in the meeting on the join screen

### LINKED ISSUE #1611: Show who is already in the meeting on the join screen
## Feature Request

The join screen asks for a name and previews the camera, and says nothing about the meeting behind it. Whoever arrives cannot tell whether anyone is there yet, or whether the people they came for have joined. Finding out means joining, which everyone already in the meeting sees.

the join screen of a meeting three people are already in: the media server lists Quill, Neel and Zora, none of the three is named on the screen, and all of them appear one click after joining

The join screen should say how many people are in the meeting. Two shapes work: a count alone, which says whether the meeting has started and names nobody, or a count with the names, which answers who came and hands the roster to whoever holds the link on a meeting anyone may enter.

davd-gzl:
Fixes #1611

## Problem

The join screen asks for a name and previews the camera, and says nothing about the meeting behind it.

Whoever arrives cannot tell whether the people they came for are there. Finding out means joining, which everyone inside sees and hears.

## Design

**[The design in one page](https://davd-gzl.github.io/agent-artifact/meet/who-is-in-the-meeting/)**, with the arithmetic and the clips.

### What the screen says

How many people are in the meeting, and who they are: five names at most, then a count of the rest. It follows the meeting while the screen is open.

Filmed against this branch on a local instance. The panel under the page is the media server's own roster, beside what the screen says.

the join screen of a meeting three people are already in: it reads "3 people are already here" over "Neel, Quill, and Zora", a fourth person joins, both lines follow with nothing reloaded, and the four are on screen one click later

### The endpoint

`GET /rooms/<id>/participants/` is a new action rather than a field on `retrieve`, which mints a token on every call and would mint one every five seconds behind this screen.

It answers with every name, and `?names=<n>` asks for at most n. `count` is the whole meeting either way, and the cut happens after the cache, so a number is not a second copy of the same room. The join screen asks for five.

### Who is told

Only the people the room would let in without approval, which is the rule `LobbyService.can_bypass_lobby` already applies. It moved to `Room.is_joinable_by` and both call it, so the lobby and this screen cannot drift on who may enter.

### What it costs the media server

About one call per meeting every six seconds, whatever the number of people waiting on the join screen. Forty of them asking for themselves would be 480 calls a minute; behind the hold the media server sees about 13.

A media server that stops answering is given three seconds, not the minute its own client would wait.

## What to review closely

- [`unregistered_room_name`](https://github.com/davd-gzl/meet/blob/feat/participants-on-join-screen/src/backend/core/api/viewsets.py#L117): a meeting nobody created is addressed by the name in the URL, and a name reading like a room id would otherwise reach a room that does exist.
- [`get_object`](https://github.com/davd-gzl/meet/blob/feat/participants-on-join-screen/src/backend/core/api/viewsets.py#L277) and that helper ask the same question about what an id is: answer it twice and a name can miss the lookup here and be taken for an id there.
- [`Room.is_joinable_by`](https://github.com/davd-gzl/meet/blob/feat/participants-on-join-screen/src/backend/core/models.py#L482): the lobby, the token mint and this endpoint all ask it now, so a mistake in it is a mistake in all three.
- [The answer is held six seconds](https://github.com/davd-gzl/meet/blob/feat/participants-on-join-screen/src/backend/meet/settings.py#L714), failures included: shorten it and a busy join screen multiplies calls to the media server.
- [The media server gets three seconds](https://github.com/davd-gzl/meet/blob/feat/participants-on-join-screen/src/backend/meet/settings.py#L718): drop the clock and one slow call holds a worker, of which the deployment runs three.

the media server: LiveKit, which carries the audio and video and knows who is connected.

the join screen: the screen where you type your name before entering a meeting.

the roster: who is in the meeting right now, as `ListParticipants` answers it.

waiting for approval: held in the lobby until someone inside lets you in.

an agent, a recorder: the transcription and the recording worker, which join a meeting and are not people.

</details>


src/backend/core/services/room_management.py
```diff
@@ -90,6 +111,59 @@ async def update_metadata(
...
+        lkapi = utils.create_livekit_client()
+
+        try:
+            # The join screen reaches this without anyone clicking anything, and
+            # the deployment runs three synchronous workers, so a LiveKit that
+            # accepts the connection and then says nothing must not hold one.
+            # The clock has to be here: the SDK passes timeout=None to aiohttp
+            # on every call, which overrides whatever the session was given and
+            # leaves the request with no timeout at all.
+            async with asyncio.timeout(settings.ROOM_PARTICIPANTS_TIMEOUT_SECONDS):
```
davd-gzl: The client cannot be given a timeout: it passes `timeout=None` to aiohttp on every request, which overrides whatever the session was built with. Configured for three seconds it still hung past ninety against an address that drops packets.

src/backend/core/services/room_management.py
```diff
@@ -2,23 +2,44 @@
...
+def _is_machine(participant: ParticipantInfo) -> bool:
+    """Check whether a participant is a recorder or an agent rather than a person.
+
+    The three clauses mirror LiveKit's own IsDependent, which is what decides
+    whether a participant counts toward the total it reports for a room. The
+    grants are checked as well as the kind because an egress worker can carry
+    the grant without the kind.
+    """
+    return (
+        participant.kind in (ParticipantInfo.Kind.AGENT, ParticipantInfo.Kind.EGRESS)
```
davd-gzl: The kind and the grants both, because an egress worker can carry the grant without the kind.

src/backend/core/api/viewsets.py
```diff
@@ -391,6 +394,65 @@ def perform_update(self, serializer):
...
+        else:
+            if not room.is_joinable_by(request.user, room.get_role(request.user)):
+                raise Http404
+            livekit_room = str(room.id)
+
+        # The join screen polls this, so everyone waiting on one meeting asks the
+        # same question at once. Holding the answer for less than one poll turns
+        # them into a single call to LiveKit however many they are.
+        cache_key = f"room_participants_{livekit_room:s}"
+        answer = cache.get(cache_key)
```
davd-gzl: Held for less than one poll, so a meeting costs one call whatever the number of people waiting on the join screen.

src/backend/core/api/viewsets.py
```diff
@@ -391,6 +394,65 @@ def perform_update(self, serializer):
...
+
+        if answer is None:
+            try:
+                answer = (
+                    RoomManagement().get_participants(livekit_room),
+                    drf_status.HTTP_200_OK,
+                )
+            except RoomManagementException:
+                # The failure is held as well as the answer. A LiveKit that
```
davd-gzl: The failure is held too. An unreachable media server is when it can least afford one call per waiting browser.

src/backend/core/api/throttling.py
```diff
@@ -87,6 +87,37 @@ class RoomKitJoinRateThrottle(MonitoredUserRateThrottle):
...
+        """Use the authenticated user ID as the throttle cache key.
+
+        Without this, an anonymous request is counted twice: UserRateThrottle
+        falls back to the IP address, which is the key ParticipantsAnonRate
+        Throttle already uses under the same scope, so each request spends two
+        of the allowance and the limit is silently halved.
+        """
+
+        if request.user and not request.user.is_authenticated:
+            return None  # Defer to ParticipantsAnonRateThrottle.
```
davd-gzl: Both throttles share a scope, and `UserRateThrottle` falls back to the IP for an anonymous request, which is the key `AnonRateThrottle` builds. Without this, every anonymous poll spends two of its allowance.

src/backend/core/api/viewsets.py
```diff
@@ -391,6 +409,78 @@ def perform_update(self, serializer):
...
+                raise Http404
+            livekit_room = str(room.id)
+
+        # The join screen polls this, so everyone waiting on one meeting asks the
+        # same question at once. Holding the answer for less than one poll turns
+        # them into a single call to LiveKit however many they are.
+        cache_key = f"room_participants_{livekit_room:s}"
+        answer = cache.get(cache_key)
+
+        if answer is None:
```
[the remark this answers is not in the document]
davd-gzl: Good point, I will change it

davd-gzl:
Answering [Qodo's review](https://github.com/suitenumerique/meet/pull/1614#issuecomment-5361136861).

- Not a bug. `=0` turns the cache off: Django expires [a `timeout` of 0](https://docs.djangoproject.com/en/5.2/topics/cache/#basic-usage) at once, and django-redis [deletes the key](https://github.com/jazzband/django-redis/blob/7.0.0/django_redis/client/default.py#L264-L274).

<details><summary>the endpoint on redis, the hold at zero and at six</summary>

`src/backend/core/tests/rooms/test_cache_off_probe.py`, which swaps the suite's local cache for the redis the deployment runs:

```python
"""The real redis backend, cache off and cache on."""

from unittest import mock

from django.core.cache import cache, caches
from django.test.utils import override_settings
from django.urls import reverse

import pytest
from livekit.protocol.models import ParticipantInfo
from livekit.protocol.room import ListParticipantsResponse
from rest_framework.test import APIClient

from core.factories import RoomFactory
from core.models import RoomAccessLevel

pytestmark = pytest.mark.django_db

REDIS = {
    "default": {
        "BACKEND": "django_redis.cache.RedisCache",
        "LOCATION": "redis://redis:6379/9",
        "OPTIONS": {"CLIENT_CLASS": "django_redis.client.DefaultClient"},
    }
}


@pytest.fixture
def lk():
    with mock.patch("core.utils.create_livekit_client") as create:
        client = mock.AsyncMock()
        client.room.list_participants.return_value = ListParticipantsResponse(
            participants=[ParticipantInfo(name="Zora"), ParticipantInfo(name="Neel")]
        )
        create.return_value = client
        yield client


@override_settings(CACHES=REDIS, ROOM_PARTICIPANTS_CACHE_SECONDS=0)
def test_probe_zero_on_real_redis(lk):
    caches["default"].clear()
    room = RoomFactory(access_level=RoomAccessLevel.PUBLIC)
    url = reverse("rooms-participants", kwargs={"pk": room.id})
    for _ in range(3):
        assert APIClient().get(url).status_code == 200
    print("\nZERO calls to the media server:", lk.room.list_participants.call_count)
    print("ZERO key held:", cache.get(f"room_participants_{room.id!s}"))
    assert lk.room.list_participants.call_count == 3


@override_settings(CACHES=REDIS, ROOM_PARTICIPANTS_CACHE_SECONDS=6)
def test_probe_six_on_real_redis(lk):
    caches["default"].clear()
    room = RoomFactory(access_level=RoomAccessLevel.PUBLIC)
    url = reverse("rooms-participants", kwargs={"pk": room.id})
    for _ in range(3):
        assert APIClient().get(url).status_code == 200
    print("\nSIX calls to the media server:", lk.room.list_participants.call_count)
    print("SIX ttl:", cache.ttl(f"room_participants_{room.id!s}"))
    assert lk.room.list_participants.call_count == 1
```

```bash
docker compose up -d postgresql redis
docker compose run --rm --no-deps -T app-dev pytest -q -s -p no:randomly core/tests/rooms/test_cache_off_probe.py
```

Three requests at a hold of zero are three calls and no key left in redis; three at six are one call and a key that expires in six seconds.

```
ZERO calls to the media server: 3
ZERO key held: None
SIX calls to the media server: 1
SIX ttl: 6
2 passed, 2 warnings in 19.62s
```

django-redis 7.0.0, [the version the lock file pins](https://github.com/davd-gzl/meet/blob/feat/participants-on-join-screen/src/backend/pyproject.toml#L39), never writes a key it has to expire in the past: it deletes instead and returns.
</details>

- Fixed: a98860eb

src/backend/core/api/viewsets.py
```diff
@@ -391,6 +409,78 @@ def perform_update(self, serializer):
...
+            except RoomManagementException:
+                # The failure is held as well as the answer. A LiveKit that
+                # cannot be reached is exactly when it can least afford one call
+                # per browser waiting on it.
+                answer = (
+                    {"error": "Could not reach the meeting."},
+                    drf_status.HTTP_503_SERVICE_UNAVAILABLE,
+                )
+
+            cache.set(cache_key, answer, settings.ROOM_PARTICIPANTS_CACHE_SECONDS)
```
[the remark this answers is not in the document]
davd-gzl: Not a bug. `=0` turns the cache off: Django expires [a `timeout` of 0](https://docs.djangoproject.com/en/5.2/topics/cache/#basic-usage) at once, and django-redis [deletes the key](https://github.com/jazzband/django-redis/blob/7.0.0/django_redis/client/default.py#L264-L274).

<details><summary>the endpoint on redis, the hold at zero and at six</summary>

`src/backend/core/tests/rooms/test_cache_off_probe.py`, which swaps the suite's local cache for the redis the deployment runs:

```python
"""The real redis backend, cache off and cache on."""

from unittest import mock

from django.core.cache import cache, caches
from django.test.utils import override_settings
from django.urls import reverse

import pytest
from livekit.protocol.models import ParticipantInfo
from livekit.protocol.room import ListParticipantsResponse
from rest_framework.test import APIClient

from core.factories import RoomFactory
from core.models import RoomAccessLevel

pytestmark = pytest.mark.django_db

REDIS = {
    "default": {
        "BACKEND": "django_redis.cache.RedisCache",
        "LOCATION": "redis://redis:6379/9",
        "OPTIONS": {"CLIENT_CLASS": "django_redis.client.DefaultClient"},
    }
}


@pytest.fixture
def lk():
    with mock.patch("core.utils.create_livekit_client") as create:
        client = mock.AsyncMock()
        client.room.list_participants.return_value = ListParticipantsResponse(
            participants=[ParticipantInfo(name="Zora"), ParticipantInfo(name="Neel")]
        )
        create.return_value = client
        yield client


@override_settings(CACHES=REDIS, ROOM_PARTICIPANTS_CACHE_SECONDS=0)
def test_probe_zero_on_real_redis(lk):
    caches["default"].clear()
    room = RoomFactory(access_level=RoomAccessLevel.PUBLIC)
    url = reverse("rooms-participants", kwargs={"pk": room.id})
    for _ in range(3):
        assert APIClient().get(url).status_code == 200
    print("\nZERO calls to the media server:", lk.room.list_participants.call_count)
    print("ZERO key held:", cache.get(f"room_participants_{room.id!s}"))
    assert lk.room.list_participants.call_count == 3


@override_settings(CACHES=REDIS, ROOM_PARTICIPANTS_CACHE_SECONDS=6)
def test_probe_six_on_real_redis(lk):
    caches["default"].clear()
    room = RoomFactory(access_level=RoomAccessLevel.PUBLIC)
    url = reverse("rooms-participants", kwargs={"pk": room.id})
    for _ in range(3):
        assert APIClient().get(url).status_code == 200
    print("\nSIX calls to the media server:", lk.room.list_participants.call_count)
    print("SIX ttl:", cache.ttl(f"room_participants_{room.id!s}"))
    assert lk.room.list_participants.call_count == 1
```

```bash
docker compose up -d postgresql redis
docker compose run --rm --no-deps -T app-dev pytest -q -s -p no:randomly core/tests/rooms/test_cache_off_probe.py
```

Three requests at a hold of zero are three calls and no key left in redis; three at six are one call and a key that expires in six seconds.

```
ZERO calls to the media server: 3
ZERO key held: None
SIX calls to the media server: 1
SIX ttl: 6
2 passed, 2 warnings in 19.62s
```

django-redis 7.0.0, [the version the lock file pins](https://github.com/davd-gzl/meet/blob/feat/participants-on-join-screen/src/backend/pyproject.toml#L39), never writes a key it has to expire in the past: it deletes instead and returns.
</details>

src/backend/core/api/viewsets.py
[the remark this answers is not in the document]
davd-gzl: Fixed: a98860eb

src/backend/core/api/viewsets.py
```diff
@@ -391,6 +444,57 @@ def perform_update(self, serializer):
...
             )
 
+    @decorators.action(
+        detail=True,
+        methods=["get"],
+        url_path="participants",
+        url_name="participants",
+        throttle_classes=[MonitoredScopedRateThrottle],
+        # ScopedRateThrottle reads this off the view. Drop it and every request
+        # is allowed, with nothing to say the endpoint went unthrottled.
```
lebaudantoine, line 454: Nit: LLMs tend to add inline comments that describe the *what* of the code rather than the *why*. I’d avoid adding too many comments unless they provide meaningful context, and keep the code as concise and self-explanatory as possible.

src/backend/core/api/viewsets.py
```diff
@@ -391,6 +444,57 @@ def perform_update(self, serializer):
...
+        except Http404:
+            if not settings.ALLOW_UNREGISTERED_ROOMS:
+                raise
+            # An unregistered room is public and named after its slug in
+            # LiveKit, exactly as the token the retrieve endpoint mints for it.
+            livekit_room = unregistered_room_name(self.kwargs["pk"])
+        else:
+            if not room.is_joinable_by(request.user, room.get_role(request.user)):
+                raise Http404
+            livekit_room = str(room.id)
```
lebaudantoine, line 485: As a matter of simplification, we could also restrict this feature to registered rooms only. What do you think? Are you actually using unregistered rooms?

src/backend/core/api/viewsets.py
```diff
@@ -113,6 +114,56 @@
...
+        # least afford one call per browser.
+        answer = (
+            {"error": "Could not reach the meeting."},
+            drf_status.HTTP_503_SERVICE_UNAVAILABLE,
+        )
+
+    # Outlives the window, so a refresh always has a copy to hand out.
+    cache.set(answer_key, answer, hold * 3)
+
+    return answer
```
lebaudantoine, line 149: I’d avoid creating a function that’s only used once, especially since this logic is quite scoped to the viewset and could live directly in the viewset file.

IMO, `get_participants` should be responsible for managing the caching of the response. If we need this at some point, we could simply allow passing a parameter to force a fresh call to LiveKit / bypass the cache.

That would let us simplify the logic and backport it directly into the viewset.

src/backend/core/api/viewsets.py
```diff
@@ -391,6 +444,57 @@ def perform_update(self, serializer):
...
+
+        Only available to users with access to the room, anonymous included.
+        Anyone else gets the same answer as a room that does not exist.
+
+        `?names=<n>` asks for at most n names. Without it the answer names
+        everyone, and `count` is the whole meeting either way.
+        """
+
+        try:
+            limit = int(request.query_params.get("names", 0))
```
lebaudantoine, line 468: Doesn't this fail to handle the positive-integer-only constraint properly? What happens if names=-5? I guess it would return the whole list.

I’d remove the possibility for the client to specify the number of names it wants. The server should be responsible for being conservative about the maximum number of names it is willing to expose.

That would also simplify the v0 implementation.

src/backend/core/api/viewsets.py
```diff
@@ -391,6 +444,57 @@ def perform_update(self, serializer):
...
+        # The join screen polls this, so everyone waiting on one meeting asks the
+        # same question at once, and one call answers all of them.
+        body, status_code = held_participants(livekit_room)
+
+        # Cut after the cache, never before, or one meeting is held once per
+        # number a caller asks for.
+        if limit > 0 and status_code == drf_status.HTTP_200_OK:
+            body = {**body, "names": body["names"][:limit]}
+
+        return drf_response.Response(body, status=status_code)
```
lebaudantoine, line 496: From a security perspective, I'd suggest twisting the disclosure model a bit, into tiers enforced server-side:

- **Anonymous users** (even on a public room): never return names + count.
- **Authenticated users** (public and trusted rooms): return names + count while the room has up to 5 participants, then switch to "meeting has started".
- **Restricted rooms**: only invited users / members of the room get names, capped at 5 as well.

Capping the number of names the API can ever return, rather than letting the client decide via `?names=`, closes an enumeration vector where the full roster of a room could be harvested with a single call.

The threshold (5) could be made configurable via settings.

WDYT?

src/backend/core/api/viewsets.py
```diff
@@ -248,18 +299,20 @@ class RoomViewSet(
     API endpoints to access and perform actions on rooms.
     """
 
+    # pylint: disable=too-many-public-methods
+
+    # DRF refuses an @action keyword the class does not declare, and the
+    # participants action sets its own scope through the decorator.
+    throttle_scope = None
```
lebaudantoine, line 306: You can follow the approach we have for `RoomKitJoinRateThrottle` and other throttling classes to be consistent with the existing codebase

src/frontend/src/features/rooms/hooks/useJoinParticipants.ts
```diff
@@ -0,0 +1,39 @@
+import { useQuery } from '@tanstack/react-query'
+import { keys } from '@/api/queryKeys'
+import { ApiError } from '@/api/ApiError'
+import { fetchParticipants } from '../api/fetchParticipants'
+
+export const POLL_INTERVAL_MS = 5000
```
lebaudantoine, line 6: I guess that, since this isn't a vital or critical feature, I'd be a bit more conservative and probably set the polling interval somewhere between 10 and 15 seconds.

We could also consider increasing the interval progressively if the response hasn't changed between polls. I'm not sure an exponential backoff is worth the added complexity here, though, a simple capped increase might be easier to implement and reason about.

What do you think?

src/backend/core/services/room_management.py
```diff
@@ -90,6 +110,56 @@ async def update_metadata(
...
+            RoomManagementException: the room could not be read.
+        """
+
+        lkapi = utils.create_livekit_client()
+
+        try:
+            # The timeout has to wrap the call: the SDK passes timeout=None to
+            # aiohttp, so the client's own is ignored and a LiveKit that goes
+            # quiet holds one of the three workers until it answers.
+            async with asyncio.timeout(settings.ROOM_PARTICIPANTS_TIMEOUT_SECONDS):
```
lebaudantoine, line 130: That’s probably something we should handle globally and address in a dedicated PR. We should probably make sure that every request we make to the LiveKit API has a timeout, so an unresponsive server can’t leave us hanging indefinitely.

There’s also a related issue around the way we create the LiveKit API session: https://github.com/suitenumerique/meet/issues/1648

PR #1619, opened 2026-08-20
✨(backend) let an operator forbid a room access level

### LINKED ISSUE #1076: Ability to restrict access levels

davd-gzl:
Takes over #1444, its author handed it on.

Fixes #1076

## Problem

A room owner can set any meeting to open access, whatever default the operator picked.

The operator has no way to forbid a level outright.

### The setting

`RESOURCE_ALLOWED_ACCESS_LEVELS` lists the levels a room may be set to. It defaults to all three, so an upgrade changes nothing.

### At boot

A deployment whose settings contradict the list does not start, rather than starting and ignoring one of them.

### Unregistered meetings

`ALLOW_UNREGISTERED_ROOMS` hands a meeting to anyone typing a URL, with no record behind it and no lobby in front of it. There is nowhere to store a level for one, so while the list forbids open meetings those URLs answer 404 instead.

### Existing meetings

The room settings panel offers only what the list holds. A meeting created before the operator narrowed the list is entered at the next stricter level the list does allow. Its host sees an empty picker and a line naming what the meeting runs as, so the choice is theirs to make. The row keeps its own value, so widening the list again restores it.

The first two commits are phanky1's, cherry-picked unchanged.

The host settings panel of an open meeting, filmed on an instance that allows all three levels, then on the same instance allowing trusted and restricted, with the room untouched in the database between the two:

the host settings panel offers three access levels with Open selected, then the same panel offers two with none selected, under a line reading that this access type is no longer allowed and the meeting runs as Open to trusted people until you choose a new one

operator: whoever runs the instance and sets its environment variables.

room owner, administrator: whoever can open a meeting's settings and change who may join.

open, trusted, restricted: the three access levels, in that order anyone with the link, anyone signed in, invited people only. Everyone else asks to join.

the allow-list: `RESOURCE_ALLOWED_ACCESS_LEVELS`, the setting this pull request adds.

</details>


src/backend/meet/settings.py
```diff
@@ -33,6 +33,34 @@
 MB = 1024 * KB
 GB = 1024 * MB
 
+# Mirrors core.models.RoomAccessLevel, which cannot be imported here because
+# settings are read before the app registry is loaded. A test asserts the two
+# stay equal.
+ROOM_ACCESS_LEVELS = ["public", "trusted", "restricted"]
```
davd-gzl: Settings load before the models, so the real list of levels cannot be imported here. A test keeps this copy equal to the model's.

src/backend/core/api/viewsets.py
```diff
@@ -324,9 +324,12 @@ def perform_create(self, serializer):
         user = self.request.user
         save_kwargs = {}
 
+        # The default is read from the database, so an allow-list narrowed after
+        # it was saved has to drop it here rather than at the serializer.
         if (
             "access_level" not in serializer.validated_data
-            and user.default_room_access_level not in (None, "")
+            and user.default_room_access_level
+            in settings.RESOURCE_ALLOWED_ACCESS_LEVELS
```
davd-gzl, line 339: This value goes straight into `save()`, where no serializer check sees it. Without the line, a default saved before the list narrowed still creates rooms at a forbidden level.

src/backend/core/api/serializers.py
```diff
@@ -161,6 +175,14 @@ def validate_configuration(self, value):
             raise serializers.ValidationError(e.errors()) from e
         return value
 
+    def validate_access_level(self, access_level):
```
davd-gzl: Checked here rather than on the model: the model validates on every save, so an old open meeting would stop saving at all. A write that keeps a room's own level is allowed for the same reason.

src/backend/core/models.py
```diff
@@ -478,10 +486,29 @@ def clean_fields(self, exclude=None):
 
         super().clean_fields(exclude=exclude)
 
+    @property
+    def effective_access_level(self):
```
davd-gzl: Read here rather than written to the row, so widening the list again restores the meeting. Next stricter rather than strictest keeps the change to the smallest one that satisfies the operator.

src/backend/core/api/serializers.py
```diff
@@ -148,8 +162,17 @@ class RoomSerializer(serializers.ModelSerializer):
...
+            "id",
+            "name",
+            "slug",
+            "configuration",
+            "access_level",
+            "effective_access_level",
+            "pin_code",
+        ]
+        read_only_fields = ["id", "slug", "effective_access_level", "pin_code"]
+        extra_kwargs = {"access_level": {"validators": [check_access_level_allowed]}}
```
davd-gzl: `validate_configuration` below uses the hook form, and this does not, so the same check reaches the two room serializers and the user preference from one place.

src/backend/core/api/viewsets.py
```diff
@@ -274,7 +274,12 @@ def retrieve(self, request, *args, **kwargs):
         try:
             instance = self.get_object()
         except Http404:
-            if not settings.ALLOW_UNREGISTERED_ROOMS:
+            # An unregistered room has no row to hold a level and no lobby in
+            # front of it, so a forbidden level cannot be enforced on one.
+            if (
+                not settings.ALLOW_UNREGISTERED_ROOMS
+                or RoomAccessLevel.PUBLIC not in settings.RESOURCE_ALLOWED_ACCESS_LEVELS
```
davd-gzl, line 281: `ALLOW_UNREGISTERED_ROOMS` mints a meeting for any slug, with no row behind it. There is nowhere to put a level, and no lobby stands in front of it, so it is an open meeting or it is nothing. Without this, an operator who forbids open meetings still hands one to anyone who types a URL.

src/backend/meet/settings.py
```diff
@@ -34,6 +34,17 @@
...
+def validate_access_level_settings(*, allowed_levels, default_level, external_default):
+    """Check that both defaults creating rooms sit inside the allowed levels."""
+    defaults = {
+        "RESOURCE_DEFAULT_ACCESS_LEVEL": default_level,
+        "EXTERNAL_API_DEFAULT_ACCESS_LEVEL": external_default,
+    }
+    for name, level in defaults.items():
+        if level not in allowed_levels:
+            raise ValueError(f"{name} must be one of RESOURCE_ALLOWED_ACCESS_LEVELS")
+
```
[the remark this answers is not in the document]
davd-gzl: Fixed: 552ee9f4

src/backend/core/api/viewsets.py
```diff
@@ -373,6 +383,7 @@ def perform_update(self, serializer):
         metadata = {
             "configuration": room.configuration,
             "access_level": room.access_level,
+            "effective_access_level": room.effective_access_level,
```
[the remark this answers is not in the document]
davd-gzl: Fixed, though not by always syncing: `effective_access_level` is derived from the settings, so no room update can carry a change to them. The metadata holds the saved level alone again and the frontend voids the computed one when the saved level arrives changed. db38afae

PR #1622, opened 2026-08-20
✨(backend) add OpenShift-compatible recording download endpoint

mbenhammouda:
Implement a Django-based download endpoint that streams recording files directly from S3 and removes the dependency on NGINX Ingress authentication annotations required by the existing download flow.

## Purpose

The current recording download flow relies on NGINX Ingress authentication annotations (`auth-url` and `auth-response-headers`) to authorize access to recordings stored in S3.

This approach works with NGINX Ingress but is not supported by OpenShift, making the recording download feature unavailable on OpenShift deployments.

## Proposal

Introduce a dedicated Django endpoint to handle recording downloads without relying on ingress-level authentication annotations.

The endpoint:

- Validates the recording identifier and file extension
- Enforces the same authorization rules as the existing `media_auth` endpoint
- Ensures the recording is in a saved state before allowing downloads
- Retrieves the file directly from S3 using the application's credentials
- Streams the content to the client using `StreamingHttpResponse`
- Returns download-related HTTP headers (`Content-Disposition`, `Content-Type`, `Content-Length`) to trigger a native browser download.

### Benefits

- Enable recording downloads on OpenShift deployments.
- Remove dependency on NGINX-specific ingress annotations.
- Preserve existing security and permission checks.
- Stream files efficiently without loading them entirely into memory.
- Provide a platform-independent implementation.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,10 @@ and this project adheres to
 
 ## [Unreleased]
 
+### Added
+
+- ✨(backend) add OpenShift-compatible recording download endpoint
+
 ### Changed
 
 - ✨(backend) accept form-urlencoded on the user token endpoint
--- a/src/backend/core/api/viewsets.py
+++ b/src/backend/core/api/viewsets.py
@@ -12,12 +12,13 @@ from django.core.exceptions import ValidationError as DjangoValidationError
 from django.core.files.storage import default_storage
 from django.db import IntegrityError, transaction
 from django.db.models import Q
-from django.http import Http404
+from django.http import Http404, StreamingHttpResponse
 from django.shortcuts import get_object_or_404
 from django.utils import timezone
 from django.utils.text import slugify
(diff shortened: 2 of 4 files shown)


mbenhammouda:
@lebaudantoine 
This MR has been reworked following the architectural feedback received during the review.

The previous approach based on dedicated media Ingresses and Services has been replaced with a Caddy-based frontend proxy responsible for handling media authorization and proxying requests to MinIO.

The Helm chart configuration has been simplified and cleaned up to rely directly on frontend.mediaProxy for both recording and file downloads.

As part of this change:

- Dedicated media Ingress and Service resources have been removed.
- Media access is now handled through the frontend Caddy proxy.
- The solution no longer depends on ingress-controller specific authentication features.
- Chart documentation has been updated to reflect the new deployment model.
- Local development configurations have been updated accordingly.

I have successfully validated the implementation on OpenShift and verified that recording and file downloads continue to work as expected.

PR #1641, opened 2026-08-26
🔥(backend) remove the S3 storage-event webhook for recordings (+95; -1258)

Leobouloc:
Recordings used to be finalized by an inbound notifications sent by S3. This tied the recording lifecycle to bucket notifications, adding complexity and dependency to limited S3 services.

The LiveKit egress_ended webhook, added as a fallback in aee1847, does the same job without any object-storage dependency. Make it the only mechanism: RecordingEventsService.handle_complete is now called on EGRESS_COMPLETE / EGRESS_LIMIT_REACHED unconditionally, instead of only when RECORDING_STORAGE_EVENT_ENABLE is False. Remove the storage-event path entirely.


src/backend/core/tests/recording/event/test_authentication.py
```diff
@@ -11,135 +11,97 @@
 
 from core.recording.event.authentication import (
```
Leobouloc, line 12: this file is kept but changing `StorageEventAuthentication` for `RecordingProcessWebhookAuthentication` as this is the only place where `HeaderBasedAuthentication` is being tested. This should ideally be refactored but seems out of scope for this PR.

PR #1642, opened 2026-08-26
🔥(backend) remove the S3 storage-event webhook for recordings (+95; -1258)

cameledev:
Recordings used to be finalized by an inbound notifications sent by S3. This tied the recording lifecycle to bucket notifications, adding complexity and dependency to limited S3 services.

The LiveKit egress_ended webhook, added as a fallback in aee1847, does the same job without any object-storage dependency. Make it the only mechanism: RecordingEventsService.handle_complete is now called on EGRESS_COMPLETE / EGRESS_LIMIT_REACHED unconditionally, instead of only when RECORDING_STORAGE_EVENT_ENABLE is False. Remove the storage-event path entirely.


src/backend/core/tests/recording/event/test_authentication.py
```diff
@@ -11,135 +11,97 @@
```
cameledev, line 11: this file is kept but changing StorageEventAuthentication for RecordingProcessWebhookAuthentication as this is the only place where HeaderBasedAuthentication is being tested. This should ideally be refactored but seems out of scope for this PR.

src/backend/meet/settings.py
```diff
@@ -729,20 +729,6 @@ class Base(Configuration):
...
-        "core.recording.event.parsers.MinioParser",
-        environ_name="RECORDING_EVENT_PARSER_CLASS",
-        environ_prefix=None,
-    )
-    RECORDING_ENABLE_STORAGE_EVENT_AUTH = values.BooleanValue(
-        True, environ_name="RECORDING_ENABLE_STORAGE_EVENT_AUTH", environ_prefix=None
-    )
-    RECORDING_STORAGE_EVENT_ENABLE = values.BooleanValue(
-        False, environ_name="RECORDING_STORAGE_EVENT_ENABLE", environ_prefix=None
-    )
```
lebaudantoine, line 742: make sure to cleanup the `env.d/development/common.dist`

CHANGELOG.md
```diff
@@ -8,6 +8,11 @@ and this project adheres to
 
 ## [Unreleased]
 
+### Changed
+
+- 🔥(backend) remove the S3 storage-event webhook for recordings
```
lebaudantoine, line 13: let's add an entry in the UPGRADE.md to explain why we delete this part, what's is the alternative, and why we don't consider it as a breaking change

src/backend/meet/settings.py
cameledev: oups forgot that. done

compose.yml
```diff
@@ -46,21 +46,6 @@ services:
...
-        condition: service_healthy
-        restart: true
-    entrypoint: >
-      sh -c "
-      /usr/bin/mc alias set meet http://minio:9000 meet password &&
-      /usr/bin/mc admin config set meet notify_webhook:meet-webhook endpoint='http://app-dev:8000/api/v1.0/recordings/storage-hook/' auth_token='Bearer password' &&
-      /usr/bin/mc admin service restart meet --wait --json &&
-      sleep 15 &&
-      /usr/bin/mc event add meet/meet-media-storage arn:minio:sqs::meet-webhook:webhook --event put --prefix "recordings" &&
-      exit 0;"
```
lebaudantoine, line 62: we've the same on the Tilt stack, https://github.com/suitenumerique/meet/commit/46fdbc0430eb4fd964b4999a946993cc3795bea9

src/backend/core/tests/recording/event/test_authentication.py
```diff
@@ -98,10 +98,11 @@ def test_authenticate_header():
 
 
 def test_multiple_spaces_in_auth_header(settings):
-    """Test success when Authorization header contains multiple spaces."""
-    settings.SUMMARY_SERVICE_WEBHOOK_API_TOKEN = "valid-test-token"
+    """Extra spaces between the scheme and the token should be tolerated."""
+    settings.SUMMARY_SERVICE_WEBHOOK_API_TOKEN = "extra-spaces-token"
```
cameledev, line 102: fixed test

CHANGELOG.md
cameledev: done

compose.yml
cameledev: was removed also from: src/helm/extra/templates/minio.yaml

lebaudantoine (approved):

PR #1664, opened 2026-09-01
 External API: update a room (config, access level)

cameledev:
## Context
RoomViewSet lacks UpdateModelMixin (viewsets.py:137), but the scope (rooms:update)
and object permissions (admin/owner) are already wired in permissions.py.

## Scope

Add UpdateModelMixin to RoomViewSet (PATCH). Make configuration, access_level, writable on update; keep id/slug/
pin_code read-only. Reuse existing validate_configuration and validate_access_level. Add analytics/audit log for external updates, mirroring the create path (viewsets.py:200-219). 

## Acceptance
Admin/owner token with `rooms:update` can PATCH name/access_level/configuration. Token without the scope -> 403
slug stays immutable.

## NB
- name will be handled in future PR, after decoupling from slug
- We allow switching to public if authorized to create public rooms
- CLAUDE: 403-vs-404 existence oracle (pre-existing, flagging for awareness): the queryset is Room.objects.all(), so a caller with rooms:update and no role on a room gets 403, confirming the UUID exists. Consistent with the existing retrieve behavior, so not a regression — but PATCH makes probing marginally more attractive.


src/backend/core/services/room_management.py
```diff
@@ -116,3 +116,32 @@ async def delete_room(self, room_name: str):
...
+    except RoomNotFoundException:
+        logger.info(
+            "LiveKit room %s does not exist yet, skipping metadata sync",
+            room.id,
+        )
+    except RoomManagementException:
+        logger.warning(
+            "Failed to sync metadata to LiveKit for room %s",
+            room.id,
+        )
```
lebaudantoine: To stay consistent with our service pattern, could we move sync_room_metadata into the RoomManagement class as a method? Keeping all helper utilities encapsulated behind the service interface makes discoverability and imports much cleaner. Wdyt?

```suggestion
    @classmethod
    def sync_room_metadata(cls, room):
        """Push a room's configuration and access level to its LiveKit room metadata.

        Failures are swallowed: a room that is not live yet, or a LiveKit hiccup,
        should never fail the request that triggered the update.
        """
        metadata = {
            "configuration": room.configuration,
            "access_level": room.access_level,
        }

        try:
            cls().update_metadata(
                room_name=str(room.id),
                metadata=metadata,
            )
        except RoomNotFoundException:
            logger.info(
                "LiveKit room %s does not exist yet, skipping metadata sync",
                room.id,
            )
        except RoomManagementException:
            logger.warning(
                "Failed to sync metadata to LiveKit for room %s",
                room.id,
            )
```

src/backend/core/external_api/viewsets.py
```diff
@@ -189,36 +195,72 @@ def list(self, request, *args, **kwargs):
...
+
+        if updated_fields:
+            sync_room_metadata(room)
+
+        self._track_room_event(
+            room,
+            analytics.AnalyticsEvent.ROOM_UPDATED,
+            updated_fields=updated_fields,
+            previous_access_level=previous_values["access_level"],
+        )
```
lebaudantoine, line 267: This will be triggered if no update occurs, is it voluntary?

src/backend/core/factories.py
```diff
@@ -48,7 +48,7 @@ def users(self, create, extracted, **kwargs):
                 else:
                     UserResourceAccessFactory(resource=self, user=item[0], role=item[1])
 
-        self.save()
+            self.save()
```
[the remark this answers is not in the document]
lebaudantoine: is it expected?

src/backend/core/external_api/viewsets.py
```diff
@@ -189,36 +195,72 @@ def list(self, request, *args, **kwargs):
...
+
+        self._track_room_event(room, analytics.AnalyticsEvent.ROOM_CREATED)
+
+    def perform_update(self, serializer):
+        """Persist the room update, sync it to LiveKit, then log and track it."""
+
+        previous_values = {
+            "access_level": serializer.instance.access_level,
+            "configuration": serializer.instance.configuration,
```
lebaudantoine: Since the configuration is a dict, is there any risk that something could mutate the same dict we're aliasing here?

src/backend/core/services/room_management.py
cameledev: agreed. changed

src/backend/core/external_api/viewsets.py
cameledev: was put here to track the call, in case of suspicious requests.

but we can remove it. wdyt ?

src/backend/core/external_api/viewsets.py
cameledev: fixed by changing to  `copy.deepcopy(serializer.instance.configuration)`

src/backend/core/factories.py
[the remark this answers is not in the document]
cameledev: actually, it seems like we can remove these save calls, because `self` is not being modified in the function so no need to save.

src/backend/core/external_api/viewsets.py
lebaudantoine: keep it

src/backend/core/services/room_management.py
```diff
@@ -116,3 +116,32 @@ async def delete_room(self, room_name: str):
...
+        should never fail the request that triggered the update.
+        """
+
+        metadata = {
+            "configuration": room.configuration,
+            "access_level": room.access_level,
+        }
+
+        try:
+            RoomManagement().update_metadata(
```
lebaudantoine:
```suggestion
            cls().update_metadata(
```
Would it work?

cameledev: change `update_metadata` and `delete_metadata` to classmethods. 

therefore this is now `cls.update_metadata`

lebaudantoine (approved):

PR #1673, opened 2026-09-02
🐛(backend) allow all ASCII characters in the user sub field

### LINKED ISSUE #1609: OIDC UUID-Subs with special characters don't work
## Bug Report

**Problematic behavior**
Several OIDC providers use the format for the sub-claim "provider|userid" (with the | separator). This causes an error. Meet's Django user model has a RegexValidator on the sub-field that allows only [A-Za-z0-9@.+-_] but no |.

**Expected behavior/code**
No error when "|" is used in sub claim.

**Steps to Reproduce**
Use Auth0 free account to test behaviour, Auth0 uses provider|userid.

**Environment**
- Meet version: 1.26.0
- Platform: Debian

**Possible Solution**
Please allow also "|" in the sub claim
OR
allow to specify another claim (e.g. email) as the user identifier.

lebaudantoine:
The user `sub` field was rejecting some ASCII characters that are
actually valid according to the OIDC spec.

Loosen the validation to accept the full ASCII range except control
characters, so the field is compliant with the RFC and works with
any spec-compliant identity provider.

Based on the Stack Overflow discussion in question 279832.

Closes #1609.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@ and this project adheres to
 
 ### Fixed
 
+- 🐛(backend) allow any printable ASCII characters in user sub field #1673
 - 🐛(frontend) keep the sending resolution picked while the camera is off #1667
 
 ## [1.30.0] - 2026-09-01
--- a/src/backend/core/authentication/backends.py
+++ b/src/backend/core/authentication/backends.py
@@ -3,7 +3,11 @@
 import contextlib
 
 from django.conf import settings
-from django.core.exceptions import ImproperlyConfigured, SuspiciousOperation
+from django.core.exceptions import (
+    ImproperlyConfigured,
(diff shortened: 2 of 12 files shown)


src/backend/core/migrations/0023_alter_user_sub_ascii_validator.py
```diff
@@ -0,0 +1,30 @@
...
+                    "255 characters or fewer. ASCII characters only."
+                ),
+                max_length=255,
+                null=True,
+                unique=True,
+                validators=[core.validators.sub_validator],
+                verbose_name="sub",
+            ),
+        ),
+    ]
```
qbey: update a previous migration to prevent Django from creating this one, because I guess this is a noop (no corresponding SQL)

qbey (approved):
👍

PR #1674, opened 2026-09-03
✨(fullstack) let an operator forbid public rooms

### LINKED ISSUE #1076: Ability to restrict access levels

davd-gzl:
Fixes #1076 · [Overview](https://github.com/davd-gzl/agent-artifact/blob/main/meet/forbid-public-rooms/overview.md)

Takes over #1444, its author handed it on. Written from main rather than carried on, after your comment on #1619: this is the single toggle.

## Problem

A room owner can open any meeting to anyone holding the link, whatever default the operator picked, and the operator has no way to forbid it.

## Design

`ALLOW_PUBLIC_ROOMS` defaults to true, so an upgrade changes nothing. A change to it lands on the next restart and never under a running server: from that boot the level is refused wherever it is set, the picker stops offering it, and the backend moves every meeting already open to anyone onto trusted.

Filmed against this branch on a local instance: the same meeting, opened by its owner, either side of the operator turning the setting off and restarting the backend. The row reads `public` on the left and `trusted` on the right.

the meeting settings of one room: with ALLOW_PUBLIC_ROOMS true the picker offers open, open to trusted people and restricted with open selected, and after the backend restarts with it false the same meeting offers two levels with open to trusted people selected

### A meeting already open to anyone

The backend moves it to trusted the first time it starts with the setting off, and a person whose default for new meetings was open moves with it. The move is one way. Turning the setting back on restores nothing, so an operator who allows the level again does not silently reopen meetings people have relied on being closed since.

The move runs from the WSGI entry point, so gunicorn and runserver run it, and management commands, celery and the test runner never do. Each worker runs it, the statements are idempotent, and the worker that finds rows logs how many it moved. A database error there fails the boot rather than serving with the meetings still open.

### Turning it off takes two settings

`RESOURCE_DEFAULT_ACCESS_LEVEL` ships public, so an instance forbidding public rooms while that default stands refuses to start rather than starting and ignoring one of the two. The documentation row says so.

### A meeting code nobody registered

Such a meeting has no row to hold a level and no waiting room in front of it, so it is open to anyone or it is nothing, and those URLs answer 404 while the setting is off. It is the one behaviour the toggle removes outright rather than narrowing.

### What is left out

Nothing tells a meeting's owner that their level moved: the picker shows trusted selected, as if they had chosen it. And a meeting live across the restart keeps `public` in its media-server metadata until its next change, so the browser reads that value as trusted while the setting is off and the host keeps the lobby controls.

open, trusted, restricted: who gets in. Anyone holding the link, anyone signed in, people invited to that meeting.

the waiting room: where someone is held until a person already inside lets them in. A meeting open to anyone has none, so forbidding that level puts one in front of meetings that never had one.

the move: what the backend does to meetings stored open, and to a person's open default, when it starts with the setting off. Both become trusted, and stay so.

the external room API: a narrower API that lets another product create meetings without a person signing in.

</details>


src/backend/core/services/room_management.py
```diff
@@ -129,7 +129,7 @@ def sync_room_metadata(cls, room):
 
         metadata = {
             "configuration": room.configuration,
-            "access_level": room.access_level,
+            "access_level": room.effective_access_level,
```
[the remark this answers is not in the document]
davd-gzl: Fixed: 6414575bad7a6eb68706bada9024968c07b6c614

The metadata a room already carries is applied over the API's answer on join, so that is where the stale level came from. It now sets the level only from a change witnessed during the meeting, which is newer than the API's answer by construction; the snapshot taken on join no longer overrides it.

src/backend/core/api/viewsets.py
```diff
@@ -256,7 +256,12 @@ def retrieve(self, request, *args, **kwargs):
...
             instance = self.get_object()
         except Http404:
-            if not settings.ALLOW_UNREGISTERED_ROOMS:
+            # An unregistered room has no row to hold a level and no lobby in
+            # front of it, so it is a public room or it is nothing.
+            if (
+                not settings.ALLOW_UNREGISTERED_ROOMS
+                or models.is_access_level_forbidden(RoomAccessLevel.PUBLIC)
+            ):
                 raise
```
WaDadidou, line 265: Silent here, loud at boot.

[meet/settings.py:37](https://github.com/suitenumerique/meet/blob/81cd3405cfb7f174f3edc2a0b31ea32f265b2a42/src/backend/meet/settings.py#L37) refuses to start on a public default. The same conflict answers 404 here, and the operator finds out from the `kubernetes.md` row or not at all. Deliberate? A boot guard on `ALLOW_UNREGISTERED_ROOMS`, or a warning, would make both halves behave alike.

src/backend/core/api/serializers.py
```diff
@@ -24,6 +24,27 @@
 logger = logging.getLogger(__name__)
 
 
+class AccessLevelField(serializers.ChoiceField):
```
WaDadidou, line 27: PUT rewrites the stored level.

This field reads the effective level and writes the stored one. `RoomViewSet` sets no `http_method_names`, so the router maps PUT as well as PATCH; the external viewset [pins its methods](https://github.com/suitenumerique/meet/blob/81cd3405cfb7f174f3edc2a0b31ea32f265b2a42/src/backend/core/external_api/viewsets.py#L164). GET then PUT writes `trusted` over a stored `public`, against "It never changes on its own" in your description. The frontend only PATCHes single fields, so nothing reaches it today. Bound the claim, or drop PUT.

src/frontend/src/features/rooms/livekit/hooks/useSyncLiveKitMetadata.ts
```diff
@@ -48,14 +48,19 @@ export const useSyncLiveKitMetadata = () => {
...
       queryClient.setQueryData<ApiRoom>([keys.room, roomSlug], (prev) => {
         if (!prev) return prev
         const nextConfiguration = parsed.configuration ?? prev.configuration
-        const nextAccessLevel = parsed.access_level ?? prev.access_level
+        // Only a change we witnessed is newer than the answer the API just
+        // gave us. The metadata a room already carries can name a level the
+        // server has since stopped enforcing.
+        const nextAccessLevel = live
+          ? (parsed.access_level ?? prev.access_level)
+          : prev.access_level
```
WaDadidou, line 63: This moves the stale window rather than closing it.

6414575 stops the join snapshot from setting `access_level`, but [the comment on the snapshot call](https://github.com/suitenumerique/meet/blob/81cd3405cfb7f174f3edc2a0b31ea32f265b2a42/src/frontend/src/features/rooms/livekit/hooks/useSyncLiveKitMetadata.ts#L79-L80) says it is there to cover joining after the last metadata change. Change the level between the API GET and the LiveKit connect and no event fires, so the client keeps the old value until the next one. Only a label is wrong, the backend still decides.

Keeping the snapshot and mapping `public` to `trusted` when `config.resource.allow_public_rooms` is false would restore the intent. That needs `useConfig` here, so no patch attached.

src/frontend/src/locales/fr/rooms.json
```diff
@@ -592,6 +592,7 @@
       "title": "Accès à la réunion",
       "description": "Ces paramètres s'appliqueront également aux futures occurrences de cette réunion.",
       "type": "Type d'accès à la réunion",
+      "enforced": "L'accès ouvert n'est plus autorisé ici, cette réunion fonctionne donc en {{level}}.",
```
WaDadidou: The French level labels are infinitive verbs, so this sentence breaks.

`trusted` is `Ouvrir aux personnes de confiance`, so the host of a room stored public reads:

> L'accès ouvert n'est plus autorisé ici, cette réunion fonctionne donc en Ouvrir aux personnes de confiance.

The other four locales use noun phrases and read fine. Quoting the label makes room for a verb:

```suggestion
      "enforced": "L'accès ouvert n'est plus autorisé ici, cette réunion est donc réglée sur « {{level}} ».",
```

WaDadidou (commented):
Second pair of eyes. I checked the scope against @lebaudantoine's ask on #1619,
ran the enforcement paths down, and read the diff in full. It holds. Four notes,
three of them inline.

### ci.yml has never run here

```
$ gh run list --branch feat/forbid-public-rooms
CI    action_required    0s    (every run since 2026-09-04)
```

The green checks are GitHub apps, which run on fork pull requests without
approval. Lint and the backend suite need a maintainer to approve the
workflows, so none of these 20 commits has been tested by CI.

I ran the backend suite locally on the branch: 1210 passed, 9 failed. Five pass
on a serial re-run, so they are xdist flakes. The other four fail identically on
the merge base ac4be27, and CI is green on `main`, so they are my environment.
The test files this PR touches: 255 of 255.

### Checked

- **Token issuance.** Three paths: [core/api/serializers.py:229](https://github.com/suitenumerique/meet/blob/81cd3405cfb7f174f3edc2a0b31ea32f265b2a42/src/backend/core/api/serializers.py#L229),
  [core/services/lobby.py:153](https://github.com/suitenumerique/meet/blob/81cd3405cfb7f174f3edc2a0b31ea32f265b2a42/src/backend/core/services/lobby.py#L153) and
  [:174](https://github.com/suitenumerique/meet/blob/81cd3405cfb7f174f3edc2a0b31ea32f265b2a42/src/backend/core/services/lobby.py#L174). All three now ask
  `is_joinable_by`. None left on the old check.
- **`is_joinable_by` dropping `user.is_authenticated` on RESTRICTED.** Safe:
  [get_role](https://github.com/suitenumerique/meet/blob/81cd3405cfb7f174f3edc2a0b31ea32f265b2a42/src/backend/core/models.py#L322) returns `None` for anonymous
  users, so `role is not None` already implies it.
- **"No frontend change needed".** `InviteDialog`, `LaterMeetingDialog`,
  `Lobby` and `useCanManageLobby` all read the effective level. Holds.
- **The external API default.** [external_api/serializers.py:105](https://github.com/suitenumerique/meet/blob/81cd3405cfb7f174f3edc2a0b31ea32f265b2a42/src/backend/core/external_api/serializers.py#L105)
  puts `EXTERNAL_API_DEFAULT_ACCESS_LEVEL` into `create` without passing
  `AccessLevelField`. Only the boot guard stops it.
- **The security claim.** [test_api_rooms_retrieve_anonymous_public_forbidden_by_the_instance](https://github.com/suitenumerique/meet/blob/81cd3405cfb7f174f3edc2a0b31ea32f265b2a42/src/backend/core/tests/rooms/test_api_rooms_retrieve.py#L179)
  asserts no `livekit` key for an anonymous visitor on a room stored public, and
  [test_can_bypass_lobby_public_room_forbidden_by_the_instance](https://github.com/suitenumerique/meet/blob/81cd3405cfb7f174f3edc2a0b31ea32f265b2a42/src/backend/core/tests/services/test_lobby.py#L932)
  asserts it at the service. Covered.

Scope: #1619 was 644 additions for a list, this is 472 for a boolean, and
[is_access_level_forbidden](https://github.com/suitenumerique/meet/blob/81cd3405cfb7f174f3edc2a0b31ea32f265b2a42/src/backend/core/models.py#L109) keeps a second
forbiddable level down to one function.

### The external API spec misses the new read semantics

[docs/openapi.yaml:503](https://github.com/suitenumerique/meet/blob/ac4be274450a53970bdeceec0f95195edf7f4bcf/docs/openapi.yaml#L503) and
[docs/resource_server.yaml:344](https://github.com/suitenumerique/meet/blob/ac4be274450a53970bdeceec0f95195edf7f4bcf/docs/resource_server.yaml#L344) say
`public` is refused on write, which still holds. Neither says `access_level` now
reads back as the level in force: an integrator fetching a room stored `public`
gets `trusted`, with nothing to account for it. Both files are hand-written, and
neither is in this diff, so I cannot anchor this one.

---

AI-assisted, per the README's *AI contributions* section. I directed it and checked every claim before posting; the permalinks are pinned to `81cd3405` so you can verify any of them.

src/backend/core/api/viewsets.py
davd-gzl: Fixed: 3e510f3c

A warning rather than a refusal: `ALLOW_UNREGISTERED_ROOMS` defaults on, so raising would stop an upgrade over a value nobody set.

[81cd3405..d3986d4a](https://github.com/suitenumerique/meet/pull/1674/changes/81cd3405cfb7f174f3edc2a0b31ea32f265b2a42..d3986d4af692ebfbd951e71096c187f60010e7e2)

src/backend/core/api/serializers.py
davd-gzl: Documented: d3986d4a

A nit in practice: nothing writes the level back today, so nobody is affected. Closing it means splitting `access_level` into a stored key and a read-only enforced one, across both APIs, the user default and sixteen tests.

Dropping PUT covers half of it. A PATCH carrying the answered level writes it too.

<details><summary>measured</summary>

```
GET answered: trusted
PATCH {"access_level": "trusted"} -> 200
stored access_level after PATCH: trusted
```

Room stored `public`, `ALLOW_PUBLIC_ROOMS=False`. The same round trip takes the user default to `null` through `/users/<id>/`, on either verb.
</details>

src/frontend/src/features/rooms/livekit/hooks/useSyncLiveKitMetadata.ts
davd-gzl: Fixed: 9fa4cea0

Your shape: the snapshot stays and a forbidden level reads as the level in force, through one rule both paths share.

[81cd3405..d3986d4a](https://github.com/suitenumerique/meet/pull/1674/changes/81cd3405cfb7f174f3edc2a0b31ea32f265b2a42..d3986d4af692ebfbd951e71096c187f60010e7e2)

src/frontend/src/locales/fr/rooms.json
davd-gzl: Fixed: ea66df94

cameledev:
Hi. Thank you for all this work ! 

I would argue against the "effective_access_level" approach as this would allow rooms to revert from effectively being TRUSTED to PUBLIC, without the user explicitly asking to do so, on an env change, which seems like a security loophole (if a link is reused for a long time and users forget). Even though I agree this will probably not be an issue for most users. 

I would be in favor of a DB sweep to actually change all PUBLIC access levels to TRUSTED on startup (if ALLOW_PUBLIC_ACCESS==False). This would:
- reduce the amount of checking that has to be done in runtime 
- limit extra code in the codebase
- add a potentially lengthy db step on startup

What do you think @lebaudantoine ?

davd-gzl:
Hi @cameledev, thank you for the review!

I initially kept the stored level untouched because I did not want to change a permission the owner never asked to change. But you are right, I am moving to your design, it will simplify the code.

src/backend/core/models.py
```diff
@@ -471,6 +487,18 @@ def clean_fields(self, exclude=None):
 
         super().clean_fields(exclude=exclude)
 
+    def is_joinable_by(self, user, role=None):
+        """Whether this user enters the meeting rather than waiting in its lobby."""
+        access_level = self.access_level
+
+        return (
+            access_level == RoomAccessLevel.PUBLIC
+            or (access_level == RoomAccessLevel.TRUSTED and user.is_authenticated)
```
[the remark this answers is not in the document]
davd-gzl: The window is deliberate. The sweep runs on every worker boot, so a row an old
worker leaves is moved by the next worker to start, and by every restart after
that.

<details>
<summary>What closes it, and what that costs</summary>

A check in `is_joinable_by` closes the window and puts the setting back on the
read path, which is what this branch moved away from on review: the stored
level is the level in force, so turning the setting back on restores nothing.
An operator notification for a moved row was weighed against the same
complexity and left out.

</details>

src/frontend/src/features/rooms/hooks/useAllowPublicRooms.ts
```diff
@@ -0,0 +1,10 @@
+import { useConfig } from '@/api/useConfig'
+
+/**
+ * Whether this instance still lets a room be open to anyone holding the link.
+ * The key is absent on an instance that has never set it, which allows it.
+ */
+export const useAllowPublicRooms = () => {
+  const { data: config } = useConfig()
+  return config?.resource?.allow_public_rooms !== false
```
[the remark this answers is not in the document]
davd-gzl: Fixed: 92bd6d78eeece2f486142f6d4afe80a54bc1cdda

The hook also feeds the LiveKit metadata sync, where a request still in flight
reading as forbidden would relabel a meeting the instance allows. It answers
three states now: the picker offers the level only on true, and the metadata
clamp fires only on false.

src/backend/core/startup.py
```diff
@@ -0,0 +1,33 @@
...
+    """
+    if models.access_level_error(models.RoomAccessLevel.PUBLIC) is None:
+        return
+
+    rooms = models.Room.objects.filter(
+        access_level=models.RoomAccessLevel.PUBLIC
+    ).update(access_level=models.RoomAccessLevel.TRUSTED)
+    users = models.User.objects.filter(
+        default_room_access_level=models.RoomAccessLevel.PUBLIC
+    ).update(default_room_access_level=models.RoomAccessLevel.TRUSTED)
```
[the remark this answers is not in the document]
davd-gzl: Fixed: 92bd6d78eeece2f486142f6d4afe80a54bc1cdda

src/backend/core/models.py
```diff
@@ -471,6 +487,18 @@ def clean_fields(self, exclude=None):
 
         super().clean_fields(exclude=exclude)
 
+    def is_joinable_by(self, user, role=None):
+        """Whether this user enters the meeting rather than waiting in its lobby."""
+        access_level = self.access_level
```
lebaudantoine, line 492: nit: unnecessary temporary variable

src/backend/core/models.py
```diff
@@ -471,6 +487,18 @@ def clean_fields(self, exclude=None):
...
+
+        return (
+            access_level == RoomAccessLevel.PUBLIC
+            or (access_level == RoomAccessLevel.TRUSTED and user.is_authenticated)
+            # get_role answers None for anyone not signed in, so a role is
+            # already proof of that and the last branch needs no second check.
+            or (access_level == RoomAccessLevel.RESTRICTED and role is not None)
```
lebaudantoine, line 499: I guess this is a slight change from the previous logic. The app should still behave as expected, but instead of relying on an inline comment, I'd make the `user.is_authenticated` requirement explicit in the code itself.

src/backend/core/models.py
```diff
@@ -106,6 +106,20 @@ class RoomAccessLevel(models.TextChoices):
...
+        return ValidationError(_("Public rooms are not allowed on this instance."))
+    return None
+
+
+def validate_access_level(access_level):
+    """Refuse a level this instance forbids, whichever writer sets it."""
+    error = access_level_error(access_level)
+    if error is not None:
+        raise error
+
```
lebaudantoine, line 121: The two functions split a single check without adding much value. Unless a caller needs the error returned rather than raised, I’d simplify this to one validator: 

```
def validate_access_level(access_level):
    if not settings.ALLOW_PUBLIC_ROOMS and access_level == RoomAccessLevel.PUBLIC:
        raise ValidationError(_("Public rooms are not allowed on this instance."))
```

I would modify the check for unregistered rooms:

```
if (
   not settings.ALLOW_UNREGISTERED_ROOMS
   or not settings.ALLOW_PUBLIC_ROOMS
):
```

src/backend/meet/wsgi.py
```diff
@@ -15,3 +15,8 @@
 environ.setdefault("DJANGO_CONFIGURATION", "Development")
 
 application = get_wsgi_application()
+
+# The application is set up now, so its own modules can be imported.
+from core import startup  # pylint: disable=wrong-import-position
+
+startup.forbid_public_rooms()
```
lebaudantoine, line 22: I’d prefer keeping this data conversion out of application startup and providing a Django management command for operators changing the configuration of an existing instance.

This permanently changes room access levels and user defaults, so making it an explicit operation, with a dry run and a summary of affected records, would make the transition easier to control.

It also gives operators more control over when the changes are applied and reduces the risk of disrupting application startup.

We should also define how existing public rooms behave when ALLOW_PUBLIC_ROOMS=False before the command has run, so enforcement doesn’t depend solely on the conversion.

Should we consider joinable a public room when ALLOW_PUBLIC_ROOMS=False? To be discussed.

src/frontend/src/features/rooms/livekit/hooks/useSyncLiveKitMetadata.ts
```diff
@@ -30,6 +31,21 @@ const parseMetadata = (raw: string | undefined): RoomLiveKitMetadata | null => {
...
+ * configuration leaves the level alone, since clamping one the instance
+ * allows would mislabel a meeting that is open on purpose.
+ */
+const enforcedAccessLevel = (
+  level: ApiAccessLevel,
+  allowPublic: boolean | undefined
+): ApiAccessLevel =>
+  allowPublic === false && level === ApiAccessLevel.PUBLIC
+    ? ApiAccessLevel.TRUSTED
+    : level
```
lebaudantoine, line 47: Could this display a room as trusted while it is still public in the backend? We’re deriving its access level from the instance configuration rather than the room’s actual state. I’d prefer the frontend to stay sync with the backend.

lebaudantoine (commented):
Thx ! I've not yet reviewed the tests

src/backend/meet/wsgi.py
davd-gzl: The new environment will load at the next restart, which is why I thought of changing the db, but I understand that can be a problem. I think we shouldn't allow existing public rooms to stay joinable in that state. An operator may not execute the command, and be frustrated when seeing them available.

Not doing it means those rooms will become public again if we change that setting later, so we need to make that state explicit to the room's administrator, with a warning in the room settings. I will also create the command, with a dry run and a summary of the affected rooms, so the move can be permanent.

Let me know if this works for you.

davd-gzl:
- Enforcement is at read time. A room stored open runs as trusted while the setting is off, and the picker stops offering the level.
- The stored level is kept. The level comes back if the setting is turned on again.
- The conversion is the operator's. `manage.py forbid_public_rooms` moves the rooms and user defaults to trusted, and --dry-run lists the rows first.
- The host is told. No selection in the picker, and a warning that the meeting runs as trusted.
- The browser follows the backend. It does not derive the level from the configuration, and the media server's copy only triggers a refetch.
- Your comments are in. The two nits, the two validators as one check, the docs row, the changelog and the tests.

PR #1676, opened 2026-09-04
Fix/perf issues

lebaudantoine:


src/backend/core/services/lobby.py
```diff
@@ -79,13 +79,76 @@ class LobbyService:
...
         return f"{settings.LOBBY_KEY_PREFIX}_{room_id!s}_{participant_id}"
 
+    @staticmethod
+    def _get_index_key(room_id: UUID) -> str:
+        """Raw Redis key of the per-room participant index (a native SET).
+
+        Built through django-redis' make_key so it lives under the same
+        KEY_PREFIX/version namespace as the participant entries.
+        """
+        return cache.client.make_key(f"{settings.LOBBY_KEY_PREFIX}-index_{room_id!s}")
```
lebaudantoine, line 92: Align with the participant key

src/backend/core/services/lobby.py
```diff
@@ -79,13 +79,76 @@ class LobbyService:
...
+        members = self._redis(write=False).smembers(self._get_index_key(room_id))
+        return [
+            member.decode() if isinstance(member, bytes) else member
+            for member in members
+        ]
+
+    def _index_touch(self, room_id: UUID) -> None:
+        """Re-arm the room index backstop TTL.
+
+        Called whenever a participant entry is written or refreshed so the
```
lebaudantoine: Streamline docs

src/backend/core/services/lobby.py
lebaudantoine, line 77: General: evaluate having a lock to avoid concurrent updates of the index

src/backend/core/services/lobby.py
```diff
@@ -79,13 +79,76 @@ class LobbyService:
...
+        Refreshes a backstop TTL on the index so an abandoned room cannot
+        leak its set beyond the longest participant timeout.
+        """
+        index_key = self._get_index_key(room_id)
+        pipe = self._redis().pipeline(transaction=False)
+        pipe.sadd(index_key, participant_id)
+        pipe.expire(index_key, settings.LOBBY_ACCEPTED_TIMEOUT)
+        pipe.execute()
+
+    def _index_members(self, room_id: UUID) -> List[str]:
```
FloChehab: Nit: to avoid errors down the road, I personnaly prefer returning unmutable data structures. Here I would go for `frozenset`.

src/backend/core/services/presence.py
```diff
@@ -29,24 +15,65 @@ def _get_cache_key(room_id: UUID | str, identity: str) -> str:
...
+    @staticmethod
+    def _redis(write: bool = True):
+        """Raw redis-py client.
+
+        SADD/SREM/SMEMBERS are not exposed by the Django cache API; this is
+        the documented django-redis escape hatch.
+        """
+        return cache.client.get_client(write=write)
+
+    def _index_members(self, room_id: UUID | str) -> List[str]:
```
FloChehab: nit: same, should return a frozenset[str]

src/backend/core/services/presence.py
```diff
@@ -29,24 +15,65 @@ def _get_cache_key(room_id: UUID | str, identity: str) -> str:
...
+
+        Deletes the indexed entries and the index itself with targeted
+        commands instead of a full-keyspace pattern scan.
+        """
+        identities = self._index_members(room_id)
+        if identities:
+            cache.delete_many(
+                [self._get_cache_key(room_id, identity) for identity in identities]
+            )
+        self._redis().delete(self._get_index_key(room_id))
```
FloChehab, line 79: I would run this delete first before the delete many (it's the most important element to delete I think)

src/backend/core/services/presence.py
```diff
@@ -29,24 +15,65 @@ def _get_cache_key(room_id: UUID | str, identity: str) -> str:
...
+        """Forget presence for every participant of a room (on room_finished).
+
+        Deletes the indexed entries and the index itself with targeted
+        commands instead of a full-keyspace pattern scan.
+        """
+        identities = self._index_members(room_id)
+        if identities:
+            cache.delete_many(
+                [self._get_cache_key(room_id, identity) for identity in identities]
+            )
```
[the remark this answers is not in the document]
FloChehab: Yeah, there might be migration issues uppon deployment

src/backend/core/services/lobby.py
```diff
@@ -267,15 +333,31 @@ def _get_participant(
             return None
 
     def list_waiting_participants(self, room_id: UUID) -> List[dict]:
```
FloChehab: I'd go for tuple and not list here, also no need to import from typing.

FloChehab (commented):
Few comments. Generally looks good to me.
One thing I am not sure is what happens uppon deployment.
And this note may trigger another one : what happens when redis is cleared for some reason. (like this is not really a cache but more a fast key / value store instead of querying the livekit API everytime).

src/backend/core/services/lobby.py
```diff
@@ -79,13 +79,76 @@ class LobbyService:
...
+    def _get_index_key(room_id: UUID) -> str:
+        """Raw Redis key of the per-room participant index (a native SET).
+
+        Built through django-redis' make_key so it lives under the same
+        KEY_PREFIX/version namespace as the participant entries.
+        """
+        return cache.client.make_key(f"{settings.LOBBY_KEY_PREFIX}-index_{room_id!s}")
+
+    @staticmethod
+    def _redis(write: bool = True):
```
qbey, line 95: Note: this might not be a redis cache you get here, it's "only" the client of the "default cache", anyway, the intent is clear to you might keep the naming.

src/backend/core/services/lobby.py
```diff
@@ -267,15 +333,31 @@ def _get_participant(
...
+            participant_id: self._get_cache_key(room_id, participant_id)
+            for participant_id in member_ids
+        }
+        data = cache.get_many(list(keys_by_id.values()))
+
+        dead_ids = [
+            participant_id
+            for participant_id, cache_key in keys_by_id.items()
+            if cache_key not in data
+        ]
```
qbey: I miss some information, is it expected to have dead IDs? or just a fail safe? should we have a warning log?

qbey (approved):
did not finish review, LGTM

src/backend/core/services/presence.py
lebaudantoine: I'd rather have an index pointing at dead keys than keys in the wild with no reference back to them. General rule: delete what's referenced before what references it, dangling references self-heal on the next read, orphaned keys don't.

src/backend/core/services/lobby.py
lebaudantoine: Yes, it's expected.

To clarify how the index works: the Redis set acts as an intermediary data layer to track participant IDs in the waiting room or accepted states. Since modern versions of Redis only support setting an expiry on a global set rather than per entry, the set indexes references to individual participant keys, and each key handles its own TTL. This behavior was already in place prior to this refactoring.

Regarding dead IDs and performance:

Is it expected? Yes, finding and cleaning up dead keys lazily during a list call acts as our safety mechanism. When a user lists waiting participants, the index returns all keys, we check their status, and any dead keys are removed from the intermediary index on the fly so subsequent calls are safe.

Network & Scale Tradeoffs: While a bit suboptimal due to the back-and-forth between the backend and Redis, this drastically narrows our search space down to the room level instead of scanning the entire database space (which previously caused performance issues).

While this request might not fully resolve recent performance troubles if the network stack itself is the bottleneck, the network chatter is now bound by the size of the room rather than the total DB size.

PR #1678, opened 2026-09-06
✨(backend) add default video codec on apiConfig struct

### LINKED ISSUE #1643: Support AV1 for webcam / screensharing
## Feature Request

In addition to VP9 I would like to support AV1 if all participants support it.

Better quality and performance when supported.

Seems to be hardcoded in meet and LiveKit backend supports it. Happy to open a PR if needed.

Since it's a simple change in the frontend, are there any considerations that I may have missed which caused this to be deferred? On the README.md it says that both VP9 and AV1 are supported.

Villaquiranm:
closes #1643 

---
The video codec was hardcoded to vp9 in Conference.tsx. It is now driven by
a new LIVEKIT_DEFAULT_VIDEO_CODEC backend setting (default: vp9), exposed in
the frontend config API and read by the room's publishDefaults, so
self-hosters can switch to av1 — or any codec LiveKit supports — without
patching the frontend.

The value is normalized and validated at startup, so a typo fails loudly
instead of silently downgrading every publisher.


lebaudantoine (approved):
LGTM

davd-gzl (commented):
> AI review, [skills](https://github.com/davd-gzl/skills) · Status: APPROVE

- `LIVEKIT_DEFAULT_VIDEO_CODEC` appears in neither environment table, so an operator looking for it has only the chart line to find it in: [`docs/installation/kubernetes.md`](https://github.com/suitenumerique/meet/blob/04fd79b56bd77bfb03acd35a86cdef1ca405de37/docs/installation/kubernetes.md?plain=1#L398-L403) lists six of the eight `LIVEKIT_` settings and [`docs/features/signaling.md`](https://github.com/suitenumerique/meet/blob/04fd79b56bd77bfb03acd35a86cdef1ca405de37/docs/features/signaling.md?plain=1#L20-L21) carries the two nearest siblings again.

src/backend/meet/settings.py
```diff
@@ -53,6 +53,26 @@ def get_release():
...
 
 
+class VideoCodecValue(values.Value):
+    """
+    A video codec name, normalized to lowercase and validated against the codecs
+    supported by the LiveKit client, so that a typo fails at startup instead of
+    silently downgrading every publisher to another codec.
+    """
+
+    codecs = frozenset(("vp8", "h264", "vp9", "av1", "h265"))
```
davd-gzl, line 63: Nit: `h265` passes this check, and every viewer whose browser cannot encode it publishes VP8 instead: Chrome [needs 136 and a hardware encoder](https://chromestatus.com/feature/5153479456456704), Firefox [has none](https://developer.mozilla.org/en-US/docs/Web/Media/Guides/Formats/WebRTC_codecs#other_video_codecs).

```suggestion
    codecs = frozenset(("vp8", "h264", "vp9", "av1"))
```

<details><summary>what each value publishes</summary>

Set `LIVEKIT_DEFAULT_VIDEO_CODEC`, join a meeting in Chrome, and read `mimeType` on the outbound video stream in `chrome://webrtc-internals`. Chromium 152 and Firefox 155 both answer `RTCRtpSender.getCapabilities('video')` with VP8, H264, AV1 and VP9 and no H.265, so the library [swaps in the first codec the media server advertises](https://github.com/livekit/client-sdk-js/blob/v2.21.0/src/room/participant/LocalParticipant.ts#L1047-L1055). The machine running this exposes a VA-API HEVC encoder and headless chromium still lists none, with `--enable-features=WebRtcAllowH265Send,PlatformHEVCEncoderSupport` making no difference, so the value turns on each viewer's browser, hardware and window rather than on the deployment. Against the repository's [`app-dev` service](https://github.com/suitenumerique/meet/blob/04fd79b56bd77bfb03acd35a86cdef1ca405de37/compose.yml#L64-L74) with a `livekit/livekit-server` reporting 1.13.5 and [no codec list configured](https://github.com/suitenumerique/meet/blob/04fd79b56bd77bfb03acd35a86cdef1ca405de37/docker/livekit/config/livekit-server.yaml):

| Setting value | Published | Streams sent |
| --- | --- | --- |
| `vp8` | VP8 | 3 |
| `h264` | H264 | 3 |
| `vp9` | VP9 | 1 |
| `av1` | AV1 | 1 |
| `h265` | VP8 | 3 |

The configuration endpoint answers `"default_video_codec": "h265"` on that last row, and neither the browser console nor the server log mentions the swap. With the suggestion applied, `h265` stops Django loading and the other four resolve unchanged.
</details>

src/backend/meet/settings.py
```diff
@@ -53,6 +53,26 @@ def get_release():
...
+
+    def to_python(self, value):
+        """Normalize the codec name and ensure it is a supported one."""
+        codec = super().to_python(value).strip().lower()
```
davd-gzl, line 67: Missing test: this line has no coverage, so dropping its `.lower()` leaves `LIVEKIT_DEFAULT_VIDEO_CODEC=AV1` failing at startup with no test red.

<details><summary>test cases</summary>

`VideoCodecValue(environ=False)` is what a test needs: the class resolves to its value inside [`Value.__new__`](https://github.com/jazzband/django-configurations/blob/2.5.1/configurations/values.py#L53-L58) whenever `environ` and `environ_name` are both set, so the instance only survives with `environ` off. Six cases, 6 passed on this head as `core/tests/test_settings_video_codec.py`:

```python
"""Unit tests for the LIVEKIT_DEFAULT_VIDEO_CODEC setting value."""

import pytest

from meet.settings import VideoCodecValue


@pytest.mark.parametrize(
    "raw,expected",
    [("vp9", "vp9"), ("AV1", "av1"), ("  h264  ", "h264")],
)
def test_video_codec_value_normalizes(raw, expected):
    """Whitespace is trimmed and the name is lowercased before it is checked."""
    # environ=False keeps __new__ from resolving the value, so the instance survives.
    assert VideoCodecValue(environ=False).to_python(raw) == expected


@pytest.mark.parametrize("raw", ["vp10", "", "h.264"])
def test_video_codec_value_rejects_unsupported(raw):
    """A name outside the accepted list stops the settings module loading."""
    with pytest.raises(ValueError, match="Unsupported video codec"):
        VideoCodecValue(environ=False).to_python(raw)
```
</details>

src/backend/meet/settings.py
```diff
@@ -53,6 +53,26 @@ def get_release():
...
+        if codec not in self.codecs:
+            raise ValueError(
+                f"Unsupported video codec {value!r}, "
+                f"expected one of: {', '.join(self.codecs)}."
```
davd-gzl, line 71: Nit: this message joins a `frozenset`, whose iteration order follows per-process string hashes, so an operator who hits the error twice reads the accepted names in two different orders.

```suggestion
                f"expected one of: {', '.join(sorted(self.codecs))}."
```

<details><summary>three starts, three orders</summary>

Three container starts with `LIVEKIT_DEFAULT_VIDEO_CODEC=vp10`, nothing else changed between them, out of the 120 orders the five names can take:

```
ValueError: Unsupported video codec 'vp10', expected one of: h264, av1, vp9, h265, vp8.
ValueError: Unsupported video codec 'vp10', expected one of: h265, av1, vp8, h264, vp9.
ValueError: Unsupported video codec 'vp10', expected one of: vp9, h264, h265, av1, vp8.
```

With the suggestion applied every start prints `av1, h264, h265, vp8, vp9`, `LIVEKIT_DEFAULT_VIDEO_CODEC="  AV1 "` still resolves to `av1`, and `pylint meet demo core` exits 0.
</details>

davd-gzl (approved):
Otherwise PR looks solid to me! :)

src/backend/meet/settings.py
Villaquiranm: Added thanks!

PR #1687, opened 2026-09-08
Fix/log enhancement

lebaudantoine:


src/backend/core/logging_filters.py
```diff
@@ -0,0 +1,21 @@
+"""Logging filters for the core application."""
+
+import logging
+
+
+class SilenceExpected401(logging.Filter):
```
lunika, line 8: Maybe allow disabling this filter using a setting? If for any reason we want to have this log.

src/backend/meet/settings.py
```diff
@@ -1136,6 +1142,9 @@ class Base(Configuration):
             ),
         },
         "loggers": {
+            "request.summary": {
+                "level": "WARNING",
```
lunika:
```suggestion
                "level": values.Value(
                    "WARNING",
                    environ_name="LOGGING_LEVEL_REQUEST_SUMMARY",
                    environ_prefix=""
                )
```

Same here, to allow changing the level if needed

lunika (approved):

Issue #1391, opened 2026-06-03
Handle EGRESS_ABORTED, EGRESS_FAILED in _handle_egress_ended


cameledev:
Currently, EGRESS_ABORTED and EGRESS_FAILED are silently ignored in _handle_egress_ended. Handle these cases.

_Originally posted by @lebaudantoine in https://github.com/suitenumerique/meet/pull/1386#discussion_r3344539964_

PR #1688, opened 2026-09-08
Handle failed and aborted egress events

cameledev:
## Purpose

https://github.com/suitenumerique/meet/issues/1391

Currently, EGRESS_ABORTED and EGRESS_FAILED are silently ignored in _handle_egress_ended. These recordings stay indefinitely as ACTIVE, and users are not notified. Handle these cases ! This may help / solve the issue: "Stale recording rows permanently block start-recording with 409, even when no egress is running".

- rename `handle_complete` to `handle_savable` to avoid confusion as `EGRESS_COMPLETE` is an egress status, but previously named `handle_complete` was handling `EGRESS_COMPLETE` *AND* `EGRESS_LIMIT_REACHED`, whereas other functions were matching name and egress status they are handling (e.g. `handle_aborted` handles `EGRESS_ABORTED`)
- in `LiveKitEventsService`, prefix handler functions with `_lkes` to avoid confusion in the codebase (between `LiveKitEventsService._lkes_handle_aborted` and `RecordingEventsService._handle_aborted`). Even though this is not great practice, I believe may be helpful to maintainers working on these parts.
- log FAILED and ABORTED recordings 
- FAILED recordings are recordings that started and were interrupted (for which a truncated audio/video may exist). In this PR, they are simply discarded (not proposed to the user); They may be usable (as are LIMIT_REACHED recordings) but this may induce errors as we don't know the reason for failure (corrupt data sent ?). I suggest leaving this as is, and depending on frequency of failed files, investigate accordingly.
- This PR chooses to leave a conceptual hole. We don't handle properly recordings which are stopped "while" a change of state is happening (stop received before egress_update to failed). In practice, this should not be a problem as: 1. This should be almost transparent for a user that clicked on stop. 2. The timeframe for such an action happening is very short (fraction of a second).


src/backend/core/models.py
```diff
@@ -582,6 +579,7 @@ class Recording(BaseModel):
     4. NOTIFICATION_SUCCEEDED: External service has been notified of this recording
 
     Error States:
+    - FAILED: Livekit egress returned EGRESS_FAILED
```
cameledev, line 582: perhaps rename this EGRESS_FAILED ?

src/backend/core/models.py
```diff
@@ -79,17 +80,13 @@ def is_final(cls, status):
...
             cls.EXTERNAL_PROCESS_FAILED,
             cls.FAILED_TO_START,
             cls.FAILED_TO_STOP,
         }
 
-    @classmethod
-    def is_unsuccessful(cls, status):
-        """Determine if the recording status represents an unsuccessful state."""
-        return status in {cls.ABORTED, cls.FAILED_TO_START, cls.FAILED_TO_STOP}
-
```
cameledev, line 92: unused in project

src/backend/core/services/livekit_events.py
```diff
@@ -178,9 +178,92 @@ def _handle_egress_updated(self, data):
...
+        """Log a participant notification error on an unsuccessful egress."""
+
+        logger.exception(
+            "Failed to notify participants that recording %s %s (room=%s)",
+            recording.id,
+            event,
+            recording.room.id,
+        )
+
+    def _lkes_handle_limit_reached(self, data, recording):
```
cameledev, line 212: prefixed with `lkes` to avoid confusion with `RecordingEventsService.handle_limit_reached`

src/backend/core/tests/recording/service/test_recording_events.py
```diff
@@ -48,28 +48,85 @@ def test_handle_limit_reached_success(mock_notify, mode, notification_type, serv
...
+    ("handler", "expected_status", "event", "notification_suffix"),
+    (
+        ("handle_limit_reached", "stopped", "limit reached", "LimitReached"),
+        ("handle_failed", "failed", "failed", "Failed"),
+        ("handle_aborted", "aborted", "aborted", "Aborted"),
     ),
 )
 @mock.patch("core.utils.notify_participants")
-def test_handle_limit_reached_error(mock_notify, mode, notification_type, service):
-    """Test handle_limit_reached raises RecordingEventsError when notification fails."""
```
cameledev, line 57: Changed behavior to logging the exception

src/backend/core/tests/services/test_livekit_events.py
```diff
@@ -126,29 +126,6 @@ def test_handle_egress_updated_success(
...
-    recording = RecordingFactory(worker_id="worker-1", status="initiated")
-    mock_data = mock.MagicMock()
-    mock_data.egress_info.egress_id = recording.worker_id
-    mock_data.egress_info.status = egress_status
-
-    service._handle_egress_updated(mock_data)
-
-    mock_update_metadata.assert_not_called()
-
-
```
cameledev, line 151: these egress statuses are never sent by livekit through update:

https://deepwiki.com/search/what-statuses-are-sent-via-egr_8edcff01-6ae5-42db-9280-c8294899f716

src/backend/core/recording/worker/services.py
```diff
@@ -71,6 +71,9 @@ def stop(self, worker_id: str) -> str:
         if response.status == livekit_api.EgressStatus.EGRESS_ABORTED:
             return "ABORTED"
 
+        if response.status == livekit_api.EgressStatus.EGRESS_FAILED:
+            return "FAILED"
```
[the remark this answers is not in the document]
cameledev: This comment:

          # Cases below should be very infrequent as status changes should be
          # received and processed by `handle_ended`, thus `stop` would not
          # be called (unless failure and stop are very close in time).
          # We therefore accept not to notify the user in this code branch.
          # This could be fixed in a future refactoring.

src/backend/core/recording/worker/services.py
```diff
@@ -66,14 +86,26 @@ def stop(self, worker_id: str) -> str:
                 "LiveKit response is missing the recording status."
             )
 
-        # To avoid exposing EgressStatus values and coupling with LiveKit outside of this class,
-        # the response status is mapped to simpler "ABORTED", "STOPPED" or "FAILED_TO_STOP" strings.
+        if response.status == livekit_api.EgressStatus.EGRESS_ENDING:
+            return "STOPPED"
+
+        if response.status == livekit_api.EgressStatus.EGRESS_LIMIT_REACHED:
+            return "STOPPED"
```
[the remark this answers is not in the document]
cameledev: If we are in this branch, it means the user clicked stop. The case described can only happen in the very short span where a user clicked stop while EGRESS_LIMIT_REACHED is being handled. In this case, it doesn't seem useful to notify the user about EGRESS_LIMIT_REACHED (adds confusion if anything).

src/backend/core/models.py
lebaudantoine, line 582: The worker and recording system are intended to be as LiveKit-agnostic as possible, so switching SFUs shouldn’t require rewriting this part. That separation isn’t consistently maintained elsewhere in the codebase, but I’d prefer to preserve it here with:

```
- FAILED: Worker failed mid-recording
```

src/backend/core/models.py
```diff
@@ -57,7 +57,8 @@ class RecordingStatusChoices(models.TextChoices):
     ACTIVE = "active", _("Active")
     STOPPED = "stopped", _("Stopped")
     SAVED = "saved", _("Saved")
-    ABORTED = "aborted", _("Aborted")
+    ABORTED = "aborted", _("Aborted")  # from livekit egress
+    FAILED = "failed", _("Failed")  # from livekit egress
```
lebaudantoine, line 61: The worker and recording system are intended to be as LiveKit-agnostic as possible, so switching SFUs shouldn’t require rewriting this part. That separation isn’t consistently maintained elsewhere in the codebase, but I’d prefer to preserve it here with:

```suggestion
    ABORTED = "aborted", _("Aborted")
    FAILED = "failed", _("Failed")
```

src/backend/core/services/livekit_events.py
```diff
@@ -178,9 +184,95 @@ def _handle_egress_updated(self, data):
...
+            except RecordingNotSavableError:
+                logger.warning(
+                    "Recording %s is not savable on egress complete "
+                    "(already saved or in an error state); ignoring.",
+                    recording.id,
+                )
+
     def _handle_egress_ended(self, data):
-        """Handle 'egress_ended' event."""
+        """Handle 'egress_ended' event.
```
lebaudantoine, line 269: To keep the LiveKit event service focused on routing webhooks to the appropriate handlers, I’d move the logic you introduced, along with most of handle_egress_ended, into the recording event service. What do you think?

src/backend/core/services/livekit_events.py
```diff
@@ -178,9 +184,95 @@ def _handle_egress_updated(self, data):
...
 
+        Egress ended is sent with one of these statuses:
+        EGRESS_COMPLETE, EGRESS_FAILED, EGRESS_ABORTED, EGRESS_LIMIT_REACHED
+        """
+
+        # Fetch recording
```
lebaudantoine, line 275: The code should be self-explanatory, so let’s avoid inline comments where possible.

src/backend/core/services/livekit_events.py
```diff
@@ -203,38 +301,18 @@ def _handle_egress_ended(self, data):
...
-                    "(already saved or in an error state); ignoring.",
-                    recording.id,
-                )
-
-        # Silently ignoring EGRESS_ABORTED, EGRESS_FAILED
+        self._lkes_handle_limit_reached(data, recording)
+        self._lkes_handle_aborted(data, recording)
+        self._lkes_handle_failed(data, recording)
+        # Handle EGRESS_COMPLETE & EGRESS_LIMIT_REACHED
+        self._lkes_handle_successful(data, recording)
```
lebaudantoine, line 315: This feels a bit odd. Calling each handler in sequence makes it look like they all apply, even if each checks the status internally. Is there a reason for this pattern? A match statement or an if/elif chain would make it clearer that these states are mutually exclusive.

I'm not sure my comment is relevant.

Also, I don't get what `lkes` means tbh.

src/backend/core/recording/services/recording_events.py
```diff
@@ -26,19 +27,75 @@ class RecordingNotSavableError(Exception):
...
+    },
+    RecordingEvent.FAILED: {
+        models.RecordingModeChoices.SCREEN_RECORDING: "screenRecordingFailed",
+        models.RecordingModeChoices.TRANSCRIPT: "transcriptionFailed",
+    },
+    RecordingEvent.ABORTED: {
+        models.RecordingModeChoices.SCREEN_RECORDING: "screenRecordingAborted",
+        models.RecordingModeChoices.TRANSCRIPT: "transcriptionAborted",
+    },
+}
```
lebaudantoine, line 52:
```suggestion
NOTIFICATION_PREFIXES = {
    models.RecordingModeChoices.SCREEN_RECORDING: "screenRecording",
    models.RecordingModeChoices.TRANSCRIPT: "transcription",
}

NOTIFICATION_SUFFIXES = {
    RecordingEvent.LIMIT_REACHED: "LimitReached",
    RecordingEvent.FAILED: "Failed",
    RecordingEvent.ABORTED: "Aborted",
}


def get_notification_type(event, recording_mode):
    return (
        f"{NOTIFICATION_PREFIXES[recording_mode]}"
        f"{NOTIFICATION_SUFFIXES[event]}"
    )
```

lebaudantoine (commented):
LGTM, minor comments, I haven't tested it yet.

src/backend/core/recording/services/recording_events.py
cameledev: done

src/backend/core/services/livekit_events.py
cameledev: Why should we avoid inlines when the code is not obvious at a glance ?

**With inlines**

image

**Without inlines**

image

The code doesn't change but I find option 1 more pleasant to work on and wouldn't argue for 1 time use functions when we could avoid them.

src/backend/core/services/livekit_events.py
cameledev: I agree. Originally moved it out of of `_handle_egress_ended` for cognitive overload.

Now moved all if/elif to a new `def dispatch_recording_to_handles(self, data, recording):`

```

    def dispatch_recording_to_handles(self, data, recording):
        """Process recording with the appropriate handle according to egress status."""

        # Limit reached
        if (
            data.egress_info.status == api.EgressStatus.EGRESS_LIMIT_REACHED
            and recording.status == models.RecordingStatusChoices.ACTIVE
        ):
            try:
                self.recording_events.handle_limit_reached(recording)
            except RecordingEventsError:
                self._log_notification_failure(recording, "limit reached")

        # Aborted
        elif (
            data.egress_info.status == api.EgressStatus.EGRESS_ABORTED
            and recording.status == models.RecordingStatusChoices.ACTIVE
        ):
            try:
                self.recording_events.handle_aborted(recording)
            except RecordingEventsError:
                self._log_notification_failure(recording, "aborted")

        # Failed
        elif (
            data.egress_info.status == api.EgressStatus.EGRESS_FAILED
            and recording.is_savable()
        ):
            try:
                self.recording_events.handle_failed(recording)
            except RecordingEventsError:
                self._log_notification_failure(recording, "failed")

        # Successful (must be placed after limit reached)
        if data.egress_info.status in [
            api.EgressStatus.EGRESS_COMPLETE,
            api.EgressStatus.EGRESS_LIMIT_REACHED,
        ]:
            try:
                self.recording_events.handle_successful(recording)
            except RecordingNotSavableError:
                logger.warning(
                    "Recording %s is not savable on egress complete "
                    "(already saved or in an error state); ignoring.",
                    recording.id,
                )

```

src/backend/core/services/livekit_events.py
cameledev: Not sure which of the new code we could actually move since it is intended to handle livekit-specific statuses. Do you mean the notifications ?

cameledev: handled by latest commits

src/backend/core/recording/services/recording_events.py
```diff
@@ -26,22 +26,106 @@ class RecordingNotSavableError(Exception):
...
+
+        logger.error(
+            "Recording worker reported %s for recording %s (room=%s, mode=%s): %s (error_code=%s)",
+            event.value,
+            recording.id,
+            recording.room.id,
+            recording.mode,
+            error or "no error reported",
+            error_code or "no error_code reported",
+        )
```
lebaudantoine: I'm concerned this could generate false-positive error logs. When a recording never actually starts because no user publishes an audio or video track, the LiveKit webhook sends an `aborted` event, which would trigger this error log. In that case, though, the abort is expected behavior and shouldn't be treated as an error.

PR #1700, opened 2026-09-10
⚡️(backend) hash application secrets with SHA-256

lebaudantoine:
Application token authentication currently uses Django's default password hasher. On our production hardware, token requests take at least ~500 ms. Authentication happens per user of each application, so this cost accumulates across frequently used integrations.

Password hashers deliberately make guessing expensive to protect human-chosen passwords after a database leak. Our application secrets are generated server-side using a cryptographically secure random generator, with a default length of 128 alphanumeric characters. Guessing these secrets is already computationally infeasible, making password stretching an unnecessary CPU cost.

Use salted SHA-256 for application secrets while retaining constant-time comparison, and keep user password hashing unchanged. Existing secrets migrate after successful verification without requiring key rotation. Conditional updates prevent migration from overwriting a concurrent rotation.

Store the new hash in `client_secret_sha256` while preserving the original Django hash in `client_secret`. New applications populate both fields, allowing the previous release to authenticate both existing and newly created applications if we need to roll back.

Once every application has migrated and the rollback window has closed, complete the migration by removing the legacy verification code and `client_secret` field.

This assumes securely generated, high-entropy secrets. Deployments that reduce  APPLICATION_CLIENT_SECRET_LENGTH` or supply predictable secrets lose the offline guessing protection that the
previous slow hasher provided.


src/backend/core/models.py
```diff
@@ -828,6 +832,43 @@ class Meta:
     def __str__(self):
         return f"{self.name!s}"
 
+    def save(self, *args, **kwargs):
+        """Populate the fast hash when creating an application."""
+        if self._state.adding:
+            self.client_secret_sha256 = hashers.hash_client_secret(
+                self.client_secret
+            )
```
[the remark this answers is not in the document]
lebaudantoine: yet there is no way to rotate a secret from the code or the Django admin, but yes it could be an issue if some modifications are introduced without being mindful of this constraint. The code is probably not bullet proof enough.

src/backend/core/hashers.py
```diff
@@ -0,0 +1,42 @@
...
+# Accept only the salted format: sha256$<salt>$<digest>.
+CLIENT_SECRET_HASH_PATTERN = re.compile(
+    rf"{re.escape(CLIENT_SECRET_HASH_PREFIX)}"
+    r"(?P<salt>[a-zA-Z0-9]{22})\$(?P<digest>[0-9a-f]{64})"
+)
+
+
+def hash_client_secret(raw_secret):
+    """Hash a machine-generated application secret without key stretching."""
+    salt = get_random_string(22)
```
briquet, line 24: Since `raw_secret` is generated server-side with high entropy via a CSPRNG, a salt adds little defense against precomputation attacks.

Instead of salting, I'd recommend dropping the salt and switching from plain `hashlib.sha256` to an [HMAC-SHA256](https://docs.python.org/3/library/hmac.html#hmac.digest) using a server-side pepper (which we can rotate using the mechanism in this PR). If `SecretField` can accept external values, we should also add a validator in `pre_save` to enforce minimum entropy.

src/backend/core/hashers.py
```diff
@@ -0,0 +1,42 @@
...
+    digest = hashlib.sha256(salt.encode() + force_bytes(raw_secret)).hexdigest()
```
briquet, line 25: With the default settings `APPLICATION_CLIENT_SECRET_LENGTH`, the **sha256 **method is fed a 128 alphanumeric characters input to produce an output of 256 bits. That is wasting 506 bits for little extra security. 

There are a few drawbacks for clients : 
- They need to send a longer HTTP header on the initial authentication request
- It will double the number of round of sha256 because it operates on 512 bits blocks

You could reduce the secret length to align to a sha256/sha512 block length depending on the security requirements.

src/backend/core/models.py
```diff
@@ -828,6 +832,47 @@ class Meta:
     def __str__(self):
         return f"{self.name!s}"
 
+    def save(self, *args, **kwargs):
+        """Populate the fast hash on creation when the raw secret is available."""
+        if self._state.adding:
```
briquet, line 837: Won't it be problematic if one needs to revoke a compromised client secret?

If we regenerate the `client_secret` the `client_secret_sha256` field will still hold a hash of the compromised `client secret` and won't be updated unless we do it manually: it means an attacker can still authenticate using the compromised secret.

lebaudantoine: yes, https://github.com/suitenumerique/meet/pull/1700#discussion_r3997864377, agree! I'll adress it before we merge the PR

lebaudantoine: rethinking it, 

I accept this as a documented transitional risk.

This risk will disappear once the legacy field is removed and secret rotation writes directly to the fast field. In the meantime, I have added a `rotate_client_secret()` model method, which should be the single supported way to rotate a client secret.

There is currently no way to rotate a key anywhere in the codebase or through the Django admin. If such a feature is introduced later, it should rely on this official method. Bypassing it would be an implementation error rather than a limitation of this approach.

src/backend/core/hashers.py
lebaudantoine: I agree we can reduce the secret length.

The additional entropy is not harmful, but it does not provide meaningful security benefits once we are already around 256 bits of entropy. For randomly generated API secrets, 128 bits is already considered strong, and 256 bits provides a very comfortable margin.

I would not make performance the main argument, though. Hashing a ~150-byte input instead of a ~60-byte one may require an additional SHA-256 compression round, but the difference is negligible compared with the ~500 ms PBKDF2 verification we are replacing. The performance improvement comes almost entirely from removing key stretching, not from reducing the SHA-256 input size.

Likewise, the HTTP overhead is fairly small.

So the main takeaway for me is simply that a 128-character secret is unnecessarily long. We can reduce APPLICATION_CLIENT_SECRET_LENGTH to a size that still gives us ~256 bits of entropy, making the credentials shorter and easier for clients to handle without materially reducing security.

src/backend/core/hashers.py
lebaudantoine: Agreed on dropping the salt as it adds no meaningful protection. I’ll remove it and use a versioned format (`sha256$v0$<digest>`) so we can evolve the scheme later.

I’d handle the pepper in a follow-up. With the now default 50-character alphanumeric secret (~298 bits of entropy), a read-only DB leak does not make recovering secrets practical. A pepper is mainly useful for weaker secrets, isn't it ?

I’ve added a startup warning below 43 characters (~256 bits). The versioned format also gives us a clean path to HMAC later.

Rather than estimating entropy in `pre_save`, I’d make secrets generated-only and read-only in admin, with a regenerate action, so high entropy is guaranteed by construction.

briquet (approved):

PR #1712, opened 2026-09-12
🐛(frontend) play the waiting room notification sound on every arrival

### LINKED ISSUE #1705: Receive an audible and visual notification when someone enters the waiting room
## Feature Request

When someone enters the waiting room and I am doing something else while waiting, I do not notice when the person arrives.

I would like to receive an audible and visual notification when someone enters the waiting room.

N/A

N/A

N/A

Thanks.

tanguychenier:
## Purpose

Closes #1705.

The waiting room already shows a toast and already plays a sound, so the feature @raphaeloger asks for is mostly there. Two things stop it from doing its job.

**The sound meant for the waiting room cannot be played.** `notifications.mp3` carries a sprite for it, 710 ms at offset 2039, distinct from the others. `triggerNotificationSound` plays `{ id: type }` where `type` is a `NotificationType`, so the sprite has to be named after the enum value. The sprite is named `waiting`, the enum value is `participantWaiting`, and howler returns without playing when the sprite id is unknown:

```js
} else if (typeof sprite === 'string' && self._state === 'loaded' && !self._sprite[sprite]) {
  // If the passed sprite doesn't exist, do nothing.
  return null;
```

`ParticipantWaiting` was also missing from `soundNotifications` in the store, so `triggerNotificationSound` would have refused it before reaching howler. `WaitingParticipantNotification` calls `NotificationType.ParticipantJoined` instead, which plays the arrival sound of a participant who is already in the call.

**A second person arriving is silent.** The sound is tied to the list going from empty to non empty:

```js
!isParticipantListEmpty(waitingParticipants) &&
isParticipantListEmpty(prevWaitingParticipant) &&
```

So once somebody waits, nobody else is announced. That is the case the issue describes: you are doing something else, someone is already in the lobby, and the next arrival passes unnoticed.

## Proposal

- [x] Name the sprite `participantWaiting`, after the notification type that plays it
- [x] Register `ParticipantWaiting` in `soundNotifications`, which also gives it a switch in the sound settings
- [x] Label it in `locales/fr`, the Crowdin source, and leave the other locales to Crowdin
- [x] Play on every arrival rather than on the first one, comparing the participant ids between two refreshes so that an admission and an arrival do not cancel each other out
- [x] Changelog entry

The quick actions panel keeps its current rule, first arrival only, so the visible behaviour does not change. The sound is no longer tied to the participants panel being closed: someone who is doing something else in another window does not see that panel either, which is the situation the issue describes, and the switch in the settings is there for anyone who would rather not hear it.

`success`, the other sprite, has no notification type either and stays unreachable. I left it alone.

## Verification

`npm run lint`, `npx prettier --check` and `npm run build` are clean.

The sprite mismatch is read off the code rather than measured on a call: `NotificationType.ParticipantWaiting` is `'participantWaiting'`, `_sprite` holds `participantJoined`, `handRaised`, `messageReceived`, `waiting` and `success`, and the howler branch above returns `null`. The volume of each sprite in the mp3 confirms `waiting` is a real sound and not padding: mean -21.3 dB, peak -3.8 dB, in the same range as `participantJoined` at -20.8 and -4.1.

What I could not check: I have no LiveKit and no backend here, so I did not hear the result in a real lobby. If you want the arrival sound rather than the dedicated one, the change is one line.


src/frontend/src/features/notifications/components/WaitingParticipantNotification.tsx
```diff
@@ -34,6 +34,13 @@ export const WaitingParticipantNotification = () => {
   const isParticipantListEmpty = (p?: WaitingParticipant[]) => p?.length == 0
 
   useEffect(() => {
+    const hasNewWaitingParticipant =
+      waitingParticipants.length > (prevWaitingParticipant?.length ?? 0)
```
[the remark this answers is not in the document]
tanguychenier: Bien vu, la longueur ne suffit pas. La détection compare maintenant les identifiants des participants entre deux rafraîchissements, donc une admission et une arrivée entre deux relevés ne s annulent plus.

src/frontend/src/features/notifications/components/WaitingParticipantNotification.tsx
```diff
@@ -34,6 +34,13 @@ export const WaitingParticipantNotification = () => {
...
+
+    if (hasNewWaitingParticipant && !isParticipantsOpen) {
+      triggerNotificationSound(NotificationType.ParticipantWaiting)
+    }
```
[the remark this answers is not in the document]
tanguychenier: Retenu aussi. Une personne occupée dans une autre fenêtre ne voit pas plus le panneau des participants que le bandeau, donc le son ne dépend plus de son ouverture. Le réglage sonore ajouté dans cette PR permet de le couper.

CHANGELOG.md
```diff
@@ -16,6 +16,7 @@ and this project adheres to
 
 ### Fixed
 
+- 🐛(frontend) sound the waiting room notification on every arrival
```
Villaquiranm: is a nit but could you improve the entry on the changelog ?
feels weird for me to read 👍 
maybe something like `🐛(frontend) play the waiting room notification sound on every arrival`

src/frontend/src/locales/fr/settings.json
```diff
@@ -102,6 +102,9 @@
       },
       "messageReceived": {
         "label": "Un message reçu"
+      },
+      "participantWaiting": {
```
Villaquiranm, line 106: It would be nice to have all the other locales also, no need to translate them manually I think we could also review the translations

CHANGELOG.md
tanguychenier: Thank you for the review. I have used your wording for the changelog entry.

The changelog entry, the commit subject and the pull request title were all the same sentence, so I updated the three of them together. If you would rather the commit subject and the title stayed as they were, I will change them back.

src/frontend/src/locales/fr/settings.json
tanguychenier: Thank you. The key is now in de, en, es and nl as well as fr.

For each one I reused the waiting room wording the project already uses (Warteraum, waiting room, sala de espera, wachtkamer) and followed the phrasing of the neighbouring entries in the same block:

- en: Participant waiting
- de: Person im Warteraum
- es: Una persona en la sala de espera
- nl: Deelnemer in de wachtkamer

Please correct any of them if they read wrong to you.

Villaquiranm (approved):
Looks good to me, I verified the behavior in main vs this branch
**In main** we do have a notification but only for the first requester.
from the second on the sound notification is missing.

On this branch the notification is triggered for every new participant

lebaudantoine (approved):
LGTM!

Your fix actually highlighted an issue we have with sound notifications: since notifications are mounted in both the main window and the PiP window, when PiP is open and someone re-enters the lobby, the sound notification is triggered twice.

I’m adding this comment as a reminder. We should probably open a separate issue for it, as I don’t think it should be addressed in this PR. @Villaquiranm would you be interested to work on it?

I'll handle the rebase and merge the pr

Villaquiranm:
> LGTM!
> 
> Your fix actually highlighted an issue we have with sound notifications: since notifications are mounted in both the main window and the PiP window, when PiP is open and someone re-enters the lobby, the sound notification is triggered twice.
> 
> I’m adding this comment as a reminder. We should probably open a separate issue for it, as I don’t think it should be addressed in this PR. @Villaquiranm would you be interested to work on it?
> 
> I'll handle the rebase and merge the pr

Sure I'll like to work on the follow-up 👍

PR #1714, opened 2026-09-14
🐛(frontend) apply the saved reception resolution when joining a meeting

### LINKED ISSUE #1606: The saved reception resolution is ignored when you join
## Bug Report

In Settings › Video, set **Select your reception resolution (max.)** to **Low definition**. Join a meeting already in progress. The cameras already on arrive at **High definition**.

In a meeting of three, you get High definition from whoever was there before you and Low definition from whoever joins after.

Here is an example run in Chromium 151. Settings shows Low definition, while the video arriving on that same device is 1280x720, the largest of the three sizes a camera offers.

Settings shows Low definition while the video arriving on that same device is 1280x720

[`VideoResolutionSubscription`](https://github.com/suitenumerique/meet/blob/e0ff28ed48c9ad2b3532cbe697b0a13bb6c0253f/src/frontend/src/features/rooms/livekit/components/VideoResolutionSubscription.tsx#L43) applies the setting only at the moment a camera starts sending, on `RoomEvent.TrackPublished`. livekit-client [does not raise that event for cameras already sending when you joined](https://github.com/livekit/client-sdk-js/blob/v2.20.0/src/room/events.ts#L87-L96), so nothing ever applies the setting to them. Turning that camera off and on does not help either: the camera is paused rather than stopped, so it never starts sending again.

Since c330ec6f, 2025-08-13.

kaelvar:
## Purpose

Fixes #1606, which was closed as fixed by #1667. #1667 fixed the **sending** resolution; #1606 is about the **reception** resolution, and the code path it describes is untouched — I checked `main` at b2abf81, and `videoSubscribeQuality` is read in exactly two places, neither of which runs on join.

Set Settings › Video › reception resolution to Low definition, then join a meeting already in progress: the cameras that were already on keep arriving in High definition. Only the participants who join after you honour the setting.

## Proposal

`VideoResolutionSubscription` listened on `RoomEvent.TrackPublished` only, which livekit-client does not raise for tracks that were already published when the local participant joined.

- apply the preference to the publications we already know about when the effect runs;
- keep `TrackPublished`, which stays the earliest point to cap a camera that starts after us — under autoSubscribe the publication is already desired, so `setVideoQuality` takes effect there;
- add `TrackSubscribed`, the first event raised for the cameras that were already sending when we joined.

That initial pass is also what `VideoTab.updateExistingRemoteVideoQuality` was doing on its own, so it is removed: joining and changing the setting are now the same code path.

The three entry points overlap on purpose; the `publication.videoQuality` guard makes the repeats free. It reads as High definition when nothing was ever requested, so the default case costs no signal round trip either.

Sent on behalf of [ALTERNATECH](https://alternatech.io), a French non-profit running a Meet instance for its members.

- [x] `tsc -b --force` clean
- [x] `eslint --max-warnings 0` clean
- [x] `prettier --check` clean
- [x] CHANGELOG entry under `[Unreleased]`


src/frontend/src/features/rooms/livekit/components/VideoResolutionSubscription.tsx
```diff
@@ -1,48 +1,57 @@
...
+    const handleTrackSubscribed = (
+      _track: RemoteTrack,
+      publication: RemoteTrackPublication
+    ) => applyQuality(publication)
+
     room.on(RoomEvent.TrackPublished, handleTrackPublished)
+    room.on(RoomEvent.TrackSubscribed, handleTrackSubscribed)
     return () => {
       room.off(RoomEvent.TrackPublished, handleTrackPublished)
+      room.off(RoomEvent.TrackSubscribed, handleTrackSubscribed)
```
lebaudantoine, line 54: Shouldn't we replace the `TrackPublished` listener with `TrackSubscribed` here?

Reading the LiveKit docs, `TrackPublished` only indicates that the track has been published to the room, while `TrackSubscribed` fires once the track is actually available to the local participant.

It also seems redundant to call `applyQuality` on both events, since subscribing to a newly published track would cause us to apply the quality twice.

lebaudantoine (commented):
Tested locally. LGTM except my comment

lebaudantoine (approved):

PR #1717, opened 2026-09-14
✨(frontend) let signed-out visitors start a meeting

kaelvar:
### Problem

`ALLOW_UNREGISTERED_ROOMS` is enabled by default, and the backend honours it in
`RoomViewSet.retrieve()`: an anonymous `GET /rooms/<slug>/` on an unknown slug
returns an ephemeral room and a LiveKit token with `canPublish: true`. This is
what happens today whenever someone opens a meeting link they were sent.

The home page, however, gates meeting creation on authentication:

```tsx
{isLoggedIn ? <CreateMeetingMenu /> : <LoginButton proConnectHint={false} />}
```

So on an instance running the default, a visitor without an account sees
"Join a meeting" and "Login", and has no way to start a meeting — even though
the server would serve them one. The only workaround is to hand them a
hand-crafted URL, which is what we ended up doing on our instance.

### Change

- backend: mirror `ALLOW_UNREGISTERED_ROOMS` into the frontend configuration;
- frontend: when the flag is on and the visitor is signed out, show the existing
  `createMeeting` button next to the login button.

The new button navigates to a freshly generated room id instead of calling
`POST /rooms/`, which stays reserved for registered (persistent) rooms and for
authenticated users. No new translation key is needed: `home:createMeeting`
already exists and is translated in every locale.

Instances that want creation reserved to accounts keep that behaviour by setting
`ALLOW_UNREGISTERED_ROOMS=False`, which already disables anonymous joining of
unknown slugs too — the two now agree instead of contradicting each other.

### Invite dialog

`Room` derived the creator mode as `isLoggedIn && history.state?.create`, so an
anonymous visitor arriving on a room they had just created was treated as a
joiner and never got the automatic prompt to copy the meeting link — which is
the whole point of having just created it.

`mode` feeds exactly one consumer, `<InviteDialog mode={mode} />`, so the session
check moved off it. `skipJoinScreen` keeps it: anonymous participants have no
display name yet and must go through the join screen. Existing flows are
unaffected, since nothing set `create` for signed-out visitors before this PR.

### Layout

Signed-out visitors can now have three controls in that row. On an instance using
the ProConnect button (fixed at `214px`) the `xsm` breakpoint (500px) leaves about
410px, which is not enough: the flex items do not overflow, they shrink, and the
two remaining labels wrap onto three lines. The row is therefore allowed to wrap.
Measured at 500px — create/join go from 84×94 and 96×94 back to 149×46 and 177×46.

### Tests

- `test_api_config_exposes_allow_unregistered_rooms`, parameterised on both
  values of the setting.
- Frontend: `tsc -b`, ESLint and Prettier pass on the changed files.

### Open question for the reviewers

For signed-out visitors the row now reads **Create a meeting | Login | Join a
meeting**. Moving "Login" last reads better, but demoting it is a product call,
so this PR leaves the order untouched.


src/frontend/src/features/rooms/routes/Room.tsx
```diff
@@ -36,7 +36,11 @@ const Room = () => {
   const { roomId } = useParams()
   const [location, setLocation] = useLocation()
   const initialRoomData = history.state?.initialRoomData
-  const mode = isLoggedIn && history.state?.create ? 'create' : 'join'
+  // `mode` only drives the invite dialog, so it does not require a session:
+  // an anonymous creator needs the share prompt just as much as a member.
+  const mode = history.state?.create ? 'create' : 'join'
+  // Skipping the join screen still does: anonymous participants have no display
+  // name yet, and that screen is where they pick one.
   const skipJoinScreen = isLoggedIn && mode === 'create'
```
lebaudantoine, line 44: I’d prefer not to change the behavior of the whole application to address this specific use case, as this modification has broader implications.

I would revert this change for now. Users creating unregistered meeting links can already access this information from the info side panel.

If this is a blocker for you, let’s discuss it. A more targeted solution could be to update the invite dialog instead: for unregistered rooms, we could display this information to any user when `mode=create`.

In that context, the distinction between admin/owner and creator is not particularly restrictive anyway, so this would solve the UX issue without changing the permission behavior globally.

lebaudantoine (commented):
Nice, thanks for the fix!

Could you also clean up the inline/block comments and keep only the ones that are really necessary? Ideally, the code should be self-explanatory, and comments should focus on the *why* rather than describing what the code already makes clear.

If there are important trade-offs or implementation choices worth documenting, I’d prefer adding a short explanation in the PR/commit body instead.

kaelvar:
Thanks for the review. Reverted the `Room.tsx` change: the invite dialog now opens on its own for unregistered rooms (null id) when the navigation carries `create`, so permissions and the join screen are unchanged. Inline comments trimmed; the rationale moved to the commit body. Rebased on `main` and squashed into a single commit.

PR #1719, opened 2026-09-15
Feat/screen share separate window

### LINKED ISSUE #1695: Allow screen sharing to be opened in a separate window
### ## Feature Request

When someone is sharing their screen, the shared content is displayed inside the main meeting window.

For users with multiple monitors, it would be useful to move the shared content to another screen while keeping the main meeting window visible to follow participants, chat, reactions, and meeting controls.

### **Describe the solution you'd like**

Add an "Open in separate window" button to the screen share control bar, next to the existing zoom and fullscreen controls.

Clicking this button would open the current screen share in a separate floating window.

The user could then move and resize this window, including moving it to another monitor, while keeping the main meeting interface fully usable.

Closing the separate window should not interrupt the screen share and should return the shared content to its normal state in the meeting.

The feature should also be fully accessible:
- The new control should be a native button with a clear accessible name.
- It should be reachable and operable using the keyboard.
- A visible focus indicator should be provided.
- Opening and closing the separate window should have predictable focus behavior, without causing keyboard users to lose their position in the meeting interface.
- The action and current state should be properly announced to screen reader users when necessary.

### **Discovery, Documentation, Adoption, Migration Strategy**

The feature could be exposed through a new button in the existing screen share control bar, alongside the zoom and fullscreen controls.

The interaction should follow the same keyboard, focus, tooltip, and accessible naming patterns as the existing controls.

### **Do you want to work on it through a Pull Request?**

Yes.

Ovgodd:
## Purpose

Screen share stays stuck in the meeting window. With two monitors you
have to pick: watch the share, or follow people, chat and controls.

We want to pop the share out, drag it to the other screen, and keep
the meeting next to it.

Stacks on `feat/screen-share-zoom`, merge that one first.

## Proposal

An “Open in separate window” button on the zoom / fullscreen bar.
Only if you’re **watching** someone else’s share, and desktop only
(mobile browsers usually block popups).

It’s a real `window.open`, not the meeting PiP. PiP is already taken,
and you only get one of those. Closing the popup (the X, the button,
or “bring it back”) puts the video back in the tile. The share keeps
going.

- [x] Open the share in another window, move it to a second screen,
      meeting still usable
- [x] Close the popup (X / button / placeholder):


Moustique82:
Hi @Ovgodd ,
Reading the purpose: the two-monitor problem also exists on the other side. When I'm the one sharing, the content is already on my second screen, so having the share take the main stage in the meeting window costs me the view of the participants instead of helping.

The description scopes the control to watching someone else's share. Is that exclusion technical (moving a local track into an auxiliary window, zoom DOM resync), or a product decision? And is the presenter's side covered elsewhere in the screen-share work currently in flight?

Thanks

lebaudantoine:
Before reviewing the code, I wanted to share some feedback on the product behavior.

I think that as soon as the screen share is opened in a separate window, the main meeting window should switch back to the grid layout, so participants can use as much space as possible.

The screen-share track displayed in the separate window should also be excluded from the tracks rendered by the focus/grid layout in the main window. It should only be re-integrated into the main layout once the floating window is closed.

Let's discuss it IRL.

lebaudantoine:
@Moustique82 We’ll address that later. For now, the zoom controls are only available when viewing someone else’s screen share.

Supporting them on your own screen share would effectively allow you to zoom into the content you’re sharing, which requires interacting with additional browser APIs. That’s why we’ve intentionally left this part out for now.

PR #1720, opened 2026-09-15
✅(agents) fix subtitle test

cameledev:
Result of test for display of subtitles was depending on local env config, potentially failing. Fix this by using override_settings.


src/backend/core/tests/rooms/test_api_rooms_subtitle.py
```diff
@@ -117,6 +118,7 @@ def test_start_subtitle_invalid_token():
     assert response.json() == {"detail": "Invalid LiveKit token: Not enough segments"}
 
 
+@override_settings(ROOM_SUBTITLE_ENABLED=False)
 def test_start_subtitle_disabled_by_default(mock_livekit_token):
```
lebaudantoine: you can use the fixture settings instead 
```suggestion
def test_start_subtitle_disabled_by_default(mock_livekit_token, settings):
   """wip."""
   settings.ROOM_SUBTITLE_ENABLED = False
```

lebaudantoine (approved):

PR #1722, opened 2026-09-15
🐛(sentry) read app release from pyproject.toml instead of version.json

briquet:
## Purpose

In settings.py, `get_release()` expects a version.json file generated during the CI Docker image build. However, this file is never created, causing the release version in Sentry to always report as "NA".

## Proposal

Read the meet semantic version directly from `pyproject.toml`


cameledev (commented):
Looks good to me.

Also matches: https://github.com/suitenumerique/meet/blob/main/src/agents/observability.py , which is good for consistency. Perhaps tests could be added there too, as it is currently untested (out of scope for this PR).

cameledev (approved):
Approved

lebaudantoine (approved):

